PR #8, opened 2024-02-29
feat: ollama support

olimorris:
First step is to extract the current OpenAI configuration into its own adapter. Then we can create Ollama and possibly other adapters for the plugin.

Adapters will be an interface which map to the `plenary.curl` library and then some. They'll allow for:
- A url
- Some headers
- Some parameters
- Callbacks which can be leveraged from within the plugin to manipulate and output the api response
- A defined schema for the API settings

The latter is important as it will be displayed to the user within the chat buffer, allowing them to tweak settings on the fly. Using the current cmp implementation it will allow users to see where and by how much they can tweak the models. It will also set out how the adapter options map up into the parameters during the request.


lua/codecompanion/adapters/openai.lua
```diff
@@ -0,0 +1,134 @@
...
+  name = "OpenAI",
+  url = "https://api.openai.com/v1/chat/completions",
+  headers = {
+    content_type = "application/json",
+    Authorization = "Bearer ", -- ignore the API key for now
+  },
+  parameters = {
+    stream = true,
+  },
+  schema = {
```
mrjones2014, line 20: is the openapi schema? you may be able to autogenerate this in CI if so :eyes:

mrjones2014:
This is looking dope I'm excited to try it out with ollama!!

olimorris:
We're slowly getting there.

The OpenAI functionality has almost been all moved over to an adapter. Just the inline functionality that needs to follow. That shouldn't be too painful.

Looking at the [Ollami API](https://github.com/ollama/ollama/blob/main/docs/api.md#generate-a-chat-completion), that should be a very quick adapter to write.

lua/codecompanion/adapters/openai.lua
olimorris: It's from their [chat completion api](https://platform.openai.com/docs/api-reference/chat/create).

Stevearc had brilliantly created a schema in his dotfiles that we leverage for this:

2024-03-03 22_45_21 - CleanShot

Love the idea of being able to do this for every adapter.

lua/codecompanion/client.lua
```diff
@@ -155,31 +154,40 @@ function Client:stream(adapter, payload, bufnr, cb)
...
 end
 
----Call the OpenAI API but block the main loop until the response is received
----@param url string
+---Call the API and block until the response is received
+---@param adapter CodeCompanion.Adapter
 ---@param payload table
 ---@param cb fun(err: nil|string, response: nil|table)
-function Client:block_request(url, payload, cb)
+function Client:call(adapter, payload, cb)
```
olimorris, line 161: Ignore this for now

olimorris:
@mrjones2014 - Ollama support now added.

A couple of tests you can try (with `ollama serve` up and running and assuming you're using `llama2`):

1. Open up the action palette and open a chat. Start conversing with the API
2. In a regular buffer run the command `:CodeCompanion print a table of 5 fruits`

A few bugs to iron out but the majority of it is done.

mrjones2014:
Wow, amazing!!!! 🚀  I will test this out this evening!

mrjones2014:
Sorry, didn't have a chance to try it out last night. Will get to it as soon as I can!

olimorris:
> Sorry, didn't have a chance to try it out last night. Will get to it as soon as I can!

No probs! I'm going to merge this to `main` now but readme is fully updated.

PR #13, opened 2024-03-08
feat(config): Allow specifying adapters by string

mrjones2014:
Allows using lazy.nvim `opts` (declarative) instead of having to use a `config` function (imperative)


olimorris:
Lovely addition. Thank you!

PR #24, opened 2024-03-17
Add support for a shell command in place of environment variable name

abayomi185:
Having the API key in environment variables is not always safe, as any application or program can read it.

> I also had a situation where a program didn't load the project specific `.env` file but used the main key from the user profile - although, this case can be rectified with the recent pattern using a different variable name.

This PR allows for this:
```lua
chat = require("codecompanion.adapters").use("openai", {
  env = {
    api_key = "cmd:gpg --decrypt ~/.openai-api-key.gpg 2>/dev/null",
  },
}),
```

I'm happy to make further changes; this change breaks assumptions about what the config should be.

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,7 +185,7 @@ require("codecompanion").setup({
     chat = require("codecompanion.adapters").use("openai", {
       env = {
         api_key = "DIFFERENT_OPENAI_KEY",
-        -- To execute a shell command, prefix it with `cmd:`.
+        -- To execute a shell command, prefix it with "cmd:"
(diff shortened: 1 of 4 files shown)


olimorris:
Thank you and great suggestion.

More than happy to add this to the plugin. Could you update the readme to reflect this too? I think this would be a popular addition.

abayomi185:
Yes. README has been updated.

olimorris:
Fantastic and thanks again. I will make sure I use this for my own API keys

PR #28, opened 2024-03-20
CodeCompanionActions - Only use telescope if exists

nuvic:
If the user does not have telescope installed, then `CodeCompanionActions` errors out. This PR checks if telescope is installed, and adds it to the `vim.ui.select` opts only if it exists. Otherwise, use the default opts.


olimorris:
Thank you and great spot!

PR #35, opened 2024-03-31
feat: add agentic workflows to the plugin

olimorris:
As per Andrew Ng's beautifully explained [tweet](https://twitter.com/AndrewYNg/status/1773393357022298617).

Initial plan is to support reflection in an automated fashion...that is, prompt the LLM to self reflect on its own work (without the user having to interact with the chat buffer).

In the future we'll figure out a way to add support for tooling and I will endeavour to make sure all of this can be done in the chat buffer.

### DIFF
--- a/lua/codecompanion/actions.lua
+++ b/lua/codecompanion/actions.lua
@@ -290,7 +290,7 @@ M.static.actions = {
     },
   },
   {
-    name = "Workflows...",
+    name = "Agentic Workflows...",
     strategy = "chat",
     description = "Workflows to improve the performance of your LLM",
     picker = {
--- a/lua/codecompanion/actions.lua
+++ b/lua/codecompanion/actions.lua
@@ -290,6 +290,62 @@ M.static.actions = {
     },
   },
   {
+    name = "Agentic Workflows...",
+    strategy = "chat",
+    description = "Workflows to improve the performance of your LLM",
+    picker = {
+      prompt = "Select a workflow",
+      items = {
+        {
+          name = "Code a feature - Outline, draft, consider and then revise",
+          callback = function(context)
+            local agent = require("codecompanion.agent")
+            return agent:workflow({
+              {
+                role = "system",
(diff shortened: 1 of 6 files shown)


olimorris:
@nuvic could you give this branch a go? You should see a new option appear in the action palette. It's still a little rough around the edges but should be working.

lua/codecompanion/actions.lua
```diff
@@ -289,6 +289,124 @@ M.static.actions = {
...
+            local agent = require("codecompanion.agent")
+            return agent
+              .new({
+                context = context,
+                strategy = "chat",
+              })
+              :workflow({
+                {
+                  role = "system",
+                  content = "You are an expert coder and helpful assistant who can help outline, draft, consider and revise code for the "
```
nuvic, line 311: There are some prompting techniques that could be helpful for getting better answers. This is something I've gotten from Jeremy Howard at fast.ai: 

> You carefully provide accurate, factual, thoughtful, nuanced answers, and are brilliant at reasoning. If you think there might not be a correct answer, you say so. always spend a few sentences explaining background context, assumptions, and step-by-step thinking BEFORE you try to answer a question. always spend a few sentences explaining background context, assumptions, and step-by-step thinking BEFORE you try to answer a question.

lua/codecompanion/actions.lua
```diff
@@ -289,6 +289,124 @@ M.static.actions = {
...
+                {
+                  role = "user",
+                  content = "Thanks. Now let's revise the code based on the feedback.",
+                  auto_submit = true,
+                },
+                {
+                  role = "user",
+                  content = "For clarity, can you show the final code without any explanations?",
+                  auto_submit = true,
+                },
```
nuvic, line 349: Since this is where we generate the final output, what do you think about just using one prompt? `Thanks. Now let's revise the code based on the feedback, without additional explanations.`

lua/codecompanion/actions.lua
olimorris: These are great suggestions. Would you PR these?

PR #40, opened 2024-04-02
Improve system and final user prompt for agent workflow

nuvic:
Change list:
- Improve the system prompt for the agent workflows 
  - use chain-of-thought, and step-back prompting.
- Condenses the final two user prompts into one prompt that just outputs the expected code

### DIFF
--- a/lua/codecompanion/actions.lua
+++ b/lua/codecompanion/actions.lua
@@ -308,7 +308,7 @@ M.static.actions = {
               :workflow({
                 {
                   role = "system",
-                  content = "You are an expert coder and helpful assistant who can help outline, draft, consider and revise code for the "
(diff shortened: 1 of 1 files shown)


lua/codecompanion/actions.lua
```diff
@@ -308,7 +308,7 @@ M.static.actions = {
               :workflow({
                 {
                   role = "system",
-                  content = "You are an expert coder and helpful assistant who can help outline, draft, consider and revise code for the "
+                  content = "You carefully provide accurate, factual, thoughtful, nuanced answers, and are brilliant at reasoning. If you think there might not be a correct answer, you say so. Always spend a few sentences explaining background context, assumptions, and step-by-step thinking BEFORE you try to answer a question. Don't be verbose in your answers, but do provide details and examples where it might help the explanation. You are an expert software engineer for the "
```
nuvic, line 311: It's the same as the system prompt below. Should we pull this out into a variable?

PR #45, opened 2024-04-13
Populate ollama models autocompletion with locally available models

lazymaniac:


olimorris:
I love this as an idea however it doesn't work as I expected it to:

https://github.com/olimorris/codecompanion.nvim/assets/9512444/1f06973c-fbda-44eb-9c78-9ef04628b1b8

Clearing the model name, I would expect to see the dropdown appear as I can't recall what Ollama models I have installed. I suspect this might be the cmp implementation.

lazymaniac:
Do you expect autocompletion to trigger autmatically when field is empty? For me it works when I trigger autocompletion with keymap (C-SPACE in my  case). I can tinker about it more
Screenshot from 2024-04-14 17-51-45

olimorris:
Ahhhh. I didn't manually trigger it. Great suggestion and thanks for the PR.

PR #51, opened 2024-05-25
fix: Handled cases with empty json.choices in OpenAI adapter. 

SDGLBL:
Some third-party OpenAI forwarding services may have a return data with an empty json.choices.


olimorris:
Thanks for the PR. How did you encounter this btw? What service were you using?

SDGLBL:
> Thanks for the PR. How did you encounter this btw? What service were you using?

https://www.ohmygpt.com

PR #58, opened 2024-06-29
feat: add inline diff view

olimorris:
Progress so far:

2024-06-29 10_39_30 - WezTerm@2x


olimorris:
2024-07-01 17_19_05 - WezTerm@2x

PR #59, opened 2024-07-06
fix(legendary): update require statement for keymaps

abayomi185:
Fixes codecompanion actions in legendary


olimorris:
That's a great spot! Thank you so much

PR #61, opened 2024-07-10
Add all Ollama options

JPyke3:
Hi There!

Thanks for the great extension!

I was running into some issues with Chat models have too short of a context length when running them through Ollama. With some of the messages I was sending with big code snippets the AI wasn't remembering most of it.

I dived in to change the `num_ctx` setting and noticed that alot of the options you can tweak for Ollama was missing, so I thought it would be handy to add them all in.

The options are all defined here https://github.com/ollama/ollama/blob/main/docs/modelfile.md#valid-parameters-and-values, I just used their defaults and descriptions.

Adjusting the context I managed to greatly improve the output of `llama3` as seen here: https://gist.github.com/JPyke3/3d9a87f138613cb00bd82a828f418fba vs the shorter context window: https://gist.github.com/JPyke3/24f995a900b968381ab6f21f3727e0dd

This is my config for the record with this change:
```lua
	adapters = {
		llama3 = require("codecompanion.adapters").use("ollama", {
			schema = {
				model = {
					default = "llama3",
				},
				num_ctx = {
					default = 16384
				},
				num_predict = {
					default = -1
				},
			},
		}),
		codegemma = require("codecompanion.adapters").use("ollama", {
			schema = {
				model = {
					default = "codegemma"
				},
			},
		}),
	},
	strategies = {
		chat = "llama3",
		inline = "codegemma",
		agent = "llama3",
	},
})

```

Hopefully the formatting/syntax is okay, I am still very much a lua rookie so I am still trying to catch my bearings.

Thanks so much for the extension! It has greatly improved my productivity 😄


olimorris:
Thank you so much! This is awesome.

I'm hoping to add cmp completion for model settings at some point this week to make it more intuitive to know what to change when in the chat buffer.

Btw, do you find all of the plugin functionality works with Ollama?

JPyke3:
No probs at all!

That's a great idea, if you need me to flesh out descriptions for cmp or whatever you need feel free to give me a ping.

I have found all the functionality has worked really well! The default prompts have especially lent themselves well to `llama3:latest` (I wouldn't be able to write them better myself), I found `gemma2:latest` to waffle a bit too much and doesn't keep it relevant to the code. For inline code I am using `codegemma:latest` - Haven't fully been satisfied with that one yet but so far the best of the 2 i tested haha. Still playing around alot though.

I have found it effectively essential to set `num_predict` to `-1` in order to ensure the AI can generate until it stops, if you're using your own PC so it shouldn't matter if the token count is high, it'll just churn until it's done.

If you have a wiki page or adapter best practice page i'd be happy to flesh out my testing and configs so that people can have a starting point for ollama. Happy to help out as I am **heavily** using the extension.

olimorris:
Good shout on `llama3:latest`. I've just tried it out. The cold starts are brutal but then she picks up speed. I've not been able to look past `claude-3-5-sonnet`. The speed and quality of response is amazing. In a chat buffer if you type `@` you'll now get a completion menu for some additional helpers. Being able to send the current buffer to the LLM as context is a huge time saver.

I've just pushed a commit which outputs virtual text next to each of the settings in the chat buffer. Nice way of displaying all of your hard work in this PR.

I wondered if you'd be willing in the next week or two to PR some sensible Ollama defaults? I'd like to have a dedicated section in the README which informs users which models to download.

davidgumberg:
Shouldn't the default be to let ollama pick those values?

JPyke3:
> Shouldn't the defaults be nil so that the ollama modelfile parameters are used?
> 
> Also, the ollama modelfile readme is inaccurate! The default value for `num_predict` is actually -1:
> 
> https://github.com/ollama/ollama/blob/d0634b15961558fd5a06119ea0a08d3bd8511e43/api/types.go#L583

That's a really good point and you're absolutely right. We should nil them out for the defaults. Do you know if there is any way to pull the values from the modelfile via the CLI? would be neat to grab them to transparently how the user what the model is using.

Thanks for the heads up on `num_predict` I will fix that in a PR

JPyke3:
Ah, seems like `ollama show llama3 --parameters` will show this:
```shell
num_keep                       24
stop                           "<|start_header_id|>"
stop                           "<|end_header_id|>"
stop                           "<|eot_id|>"
```

That's probably what I am after. Alright I will snap together a PR that reads from this first, and if there is no value present it will read from the defaults. That way the user knows exactly what the model is using/doing.

PR #62, opened 2024-07-11
feat: update agent interface documentation and implementation

SDGLBL:
- Updated `AGENTS.md` to include new `execute` function for Buffer Editor agent.
- Added detailed explanation for Command-based and Function-based agent implementation types.
- Introduced Buffer Editor agent in the interface and its function execution method.
- Modified `init.lua` to support new agent types and added agent execution logic.
- Extended `config.lua` with Buffer Editor agent configuration.
- Enhanced `helpers/buffers.lua` to provide buffer information and format buffer content.
- Updated `strategies/chat.lua` to correctly require `agents.init` module.
- Improved `strategy.lua` to include context in the agent strategy.
- Added function to get active buffers in `utils/util.lua`.

CleanShot 2024-07-11 at 19 51 23

### DIFF
--- a/lua/codecompanion/adapters/anthropic.lua
+++ b/lua/codecompanion/adapters/anthropic.lua
@@ -176,7 +176,7 @@ return {
       mapping = "parameters",
       type = "enum",
       desc = "The model that will complete your prompt. See https://docs.anthropic.com/claude/docs/models-overview for additional details and options.",
-      default = "claude-3-haiku-20240307",
+      default = "claude-3-5-sonnet-20240620",
       choices = {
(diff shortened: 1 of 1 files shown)


SDGLBL:
During development, it is often found that purely chat models make complex code modifications very troublesome, requiring constant copying and pasting between the buffer and chat. Therefore, I attempted to introduce a type of agent that can directly interact with the buffer to allow the model to modify the buffer directly using a Search/Replace method. This part of the inspiration mainly comes from [aider](https://github.com/paul-gauthier/aider).

PR #66, opened 2024-07-14
Ollama Validation Functions

JPyke3:
Hello again!

Loving the new total token feature by the way!

This is a PR to add Validation to the Ollama Schema, unfortunately due to the nature of Ollama it has to be a little open-ended as you can theoretically set the defaults to whatever you want in alot of cases, but at least this will prevent users from doing something you absolutely *cannot* do.


olimorris:
Awesome!!!! Thank you so much

Issue #67, opened 2024-07-14
[Bug]: Ollama schema incorrrectly invalidating valid models


JPyke3:
### Your `minimal.lua` config

```lua
require("codecompanion").setup({
	adapters = {
		llama3 = require("codecompanion.adapters").use("ollama", {
			schema = {
				model = {
					default = "llama3",
				},
				num_ctx = {
					default = 16384
				},
				num_predict = {
					default = -1
				},
			},
		}),
		codegemma = require("codecompanion.adapters").use("ollama", {
			schema = {
				model = {
					default = "codegemma"
				},
			},
		}),
	},
	strategies = {
		chat = {
			adapter = "llama3",
		},
		inline = {
			adapter = "codegemma",
		},
		agent = {
			adapter = "llama3",
		},
	},
})
```

### Error messages

<img width="652" alt="image" src="https://github.com/user-attachments/assets/7274cdc6-f388-4b71-bd67-247e49038674">

### Health check output

```

==============================================================================
codecompanion: require("codecompanion.health").check()

codecompanion.nvim report ~
- Log file: /Users/jacobpyke/.local/state/nvim/codecompanion.log
- OK plenary.nvim installed
- OK nvim-treesitter installed
- OK telescope.nvim installed
- OK dressing.nvim installed
- OK curl installed
- OK base64 installed

```

### Log output

n/a

### Describe the bug

I noticed the `get_ollama_choices()` https://github.com/olimorris/codecompanion.nvim/blob/ab040343ba0bb4dce08b072a7eab5f5ef4529824/lua/codecompanion/adapters/ollama.lua#L4 function doesn't correctly return some valid models. When using `ollama` you can just stipulate `llama3` and it will resolve to `llama3:latest`. Happy to fix it but ran out of time this morning to get it done, so I am leaving this bug report as a slight reminder. I should be able to have it fixed by this afternoon.

### Final checks

- [X] I have made sure this issue exists in the latest version of the plugin
- [X] I have tested with the `minimal.lua` config file above and still get the issue
- [X] I have shared the contents of the log file

PR #68, opened 2024-07-14
Fix: Include base model name for entries in ollama

JPyke3:
This is a fix for https://github.com/olimorris/codecompanion.nvim/issues/67.

If a user has a `:latest` model installed on their system it should also add the model without a tag. Generally I just use `llama3` in my config file and this was throwing a validation as I *technically* have `llama3:latest` on my system.


olimorris:
Once again, thank you! Great to add so much functionality for Ollama

PR #74, opened 2024-07-23
Fix last instance of `get_code` being called from `codecompanion.helpers.code`

davidgumberg:
https://github.com/olimorris/codecompanion.nvim/commit/e4466f8242821da4b784df88a604bad13b8e9b52 moved the `get_code` function to `codecompanion.helpers.actions`, but the send_code action still tries to call it from helpers.code, resulting in a crash.


olimorris:
Great spot thank you!

PR #75, opened 2024-07-24
feat(copilot) impl copilot (preview agent impl)

SDGLBL:
This is a brand new agent implementation (preview version). Considering its current capabilities, I'm calling it copilot. It's implemented by inheriting from strategy chat and overriding the submit function. The effect is as shown in the gif.

1. add comment then write test then run test

CleanShot 2024-07-24 at 23 39 03

2. read web page then write comment 
CleanShot 2024-07-24 at 23 47 18

3. fix lsp error
CleanShot 2024-07-24 at 23 49 06

Current known drawbacks and todos:

1. It's extremely token-consuming. This implementation can only run for extended periods on high-performance-to-price ratio models like gpt-4o-mini, claude-3-haiku, and Dee-seek. For models with lower IQ like gpt-4o-mini and claude-3-haiku, it's actually unable to achieve the best results.
2. The current prompts still need more fine-tuning. I haven't had time to optimize the previous prompts based on the current implementation yet.
3. The command_runner tool hasn't been rewritten for asynchronous execution.
4. The rag tool in search mode returns a large amount of string data, which easily causes Neovim to get stuck.
5. Write test code.

### DIFF
--- a/lua/codecompanion/actions.lua
+++ b/lua/codecompanion/actions.lua
@@ -65,6 +65,29 @@ M.static.actions = {
     },
   },
   {
+    name = "Copilot",
+    strategy = "copilot",
+    description = "Open a chat buffer with powerful tools to converse with an LLM",
+    type = nil,
+    opts = {
+      stop_context_insertion = true,
+    },
+    prompts = {
+      n = function()
+        return require("codecompanion").copilot()
+      end,
+      v = {
+        {
+          role = "user",
+          contains_code = true,
+          content = function(context)
+            return send_code(context)
+          end,
+        },
+      },
+    },
+  },
+  {
     name = "Prompts ...",
     strategy = " ",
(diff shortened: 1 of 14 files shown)


olimorris:
Firstly, thanks for your time and effort in putting together this PR.

Right now, I don't want to put any PRs through that change the structure of the plugin or alter the config. This PR simply proposes too much change for me to be willing to maintain.

I'm also largely questioning whether agents/tools should be getting as much focus as we've both given them in the last month. As you say, they're super expensive in terms of tokens and I've struggled to get them working with any model that I have configured through Ollama.

SDGLBL:
Well, from a practical standpoint, perhaps this mode can only be considered valuable when the prices of today's most advanced large models drop to the level of models like gpt4o-mini and haiku.

Issue #9, opened 2024-03-07
Notice of breaking changes


olimorris:
Whilst I'm adopting a fairly rigid approach to maintaining this plugin (as per the README), I will aim to notify users of any breaking changes (and possibly other critical info) via this issue. Users are strongly encouraged to subscribe to this issue to get notified of such changes.

Every breaking change is added as a new comment.

olimorris:
## Breaking changes to the config

> [!NOTE] 
> If you've not altered the default configuration, you can ignore this change.

To enable more customisation of adapters, https://github.com/olimorris/codecompanion.nvim/pull/39 will shortly be merged which introduces strategies and a change to how adapters are configured:

### Before:

```lua
-- default config
require("codecompanion").setup({
  adapters = {
    chat = "openai",
    inline = "openai",
  },
})
```

### After:

```lua
-- default config
require("codecompanion").setup({
  adapters = {
    anthropic = require("codecompanion.adapters").use("anthropic"),
    ollama = require("codecompanion.adapters").use("ollama"),
    openai = require("codecompanion.adapters").use("openai"),
  },
  strategies = {
    chat = "openai",
    inline = "openai",
  },
})
```

### Benefits

This feature will  allow you to define/customise adapters much more easily and apply them to strategies:

```lua
require("codecompanion").setup({
   adapters = {
     anthropic = require("codecompanion.adapters").use("anthropic", {
       env = {
         api_key = "ANTHROPIC_API_KEY_1"
       },
     }),
     anthropic_2 = require("codecompanion.adapters").use("anthropic", {
       env = {
         api_key = "ANTHROPIC_API_KEY_2"
       },
     }),
   },
   strategies = {
     chat = "anthropic",
     inline = "anthropic_2",
   },
})
```

It also allows you to do `:CodeCompanionChat ollama` to open a chat with the Ollama adapter.

olimorris:
## Breaking changes to the config

Display options will change significantly in the next commit.

### Before:

```lua
display = {
  chat = {
    type = "float",
    buf_options = {
      buflisted = false,
    },
    float_options = {
      border = "single",
      max_height = 0,
      max_width = 0,
    },
    win_options = {
      cursorcolumn = false,
      cursorline = false,
      foldcolumn = "0",
      linebreak = true,
      list = false,
      signcolumn = "no",
      spell = false,
      wrap = true,
    },
  },
},
```

### After:

```lua
display = {
  chat = {
    buffer = {
[...]

PR #88, opened 2024-08-04
feat: better workflow

olimorris:
Users can now move between open chat buffers with `}` and `{`. There is prevention built into the plugin to stop multiple chat buffers being open at the same time. Global variables have also been reduced.

### DIFF
--- a/README.md
+++ b/README.md
@@ -355,6 +355,8 @@ When in the chat buffer, there are number of keymaps available to you:
 - `gc` - Clear the buffer's contents
 - `ga` - Add a codeblock
 - `gs` - Save the chat to disk
+- `}` - Move to the next chat
+- `{` - Move to the previous chat
 - `[` - Move to the next header
 - `]` - Move to the previous header
 
--- a/lua/codecompanion/actions.lua
+++ b/lua/codecompanion/actions.lua
@@ -76,28 +76,27 @@ M.static.actions = {
       stop_context_insertion = true,
     },
     condition = function()
-      return _G.codecompanion_chats and utils.count(_G.codecompanion_chats) > 0
+      return #require("codecompanion").buf_get_chat() > 0
     end,
     picker = {
       prompt = "Select a chat",
       items = function()
-        local ui = require("codecompanion.utils.ui")
-        local chats = {}
+        local loaded_chats = require("codecompanion").buf_get_chat()
+        local open_chats = {}
 
-        for bufnr, chat in pairs(_G.codecompanion_chats) do
-          table.insert(chats, {
-            name = chat.name,
+        for _, data in ipairs(loaded_chats) do
+          table.insert(open_chats, {
+            name = data.name,
             strategy = "chat",
-            description = chat.description,
+            description = data.description,
             callback = function()
-              _G.codecompanion_chats[bufnr] = nil
-              chat.chat:open()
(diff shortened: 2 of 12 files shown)


olimorris:
@niba please try this PR out and let me know your thoughts

niba:
Wow, that was fast! Thanks! I'll test it tomorrow and update this comment with feedback

olimorris:
Looks like the keymaps in your config don't conform to the new API: https://github.com/olimorris/codecompanion.nvim/issues/9#issuecomment-2265075381

niba:
I updated my config to the latest version. The version from the main branch works without problems. I even created pr #89 to fix some issues with the new config.

I also tried calling the command directly without using keymaps, and the effect is the same. I will double-check everything.

olimorris:
That's odd. Could you share your complete config with me?

niba:
I resolved the issue. You changed the callback names for header navigation, and I still had the old ones in my config. Sorry about that.

olimorris:
Great spot! Yeah I need to do better in terms of breaking the config and making it easier for users to identify what's changed

niba:
This is incredibly good. I tested all combinations, and everything works perfectly. Superb job!

The only nitpick I have is this scenario:

1. Open two chats: Chat 1 and Chat 2.
2. Destroy Chat 2 - **Question**: Should destroying a chat close the chat buffer or shift focus to the previous/next chat?
3. Destroying a chat closes it, leaving you in the code buffer. Use the toggle command to bring up the chat buffer again.
4. The toggle creates a new chat instead of focusing on existing one. This only happens when your last action in the chat buffer was to destroy a chat.

niba:
Thanks!

Functionally, everything works, but there is one small visual detail. I have Treesitter context on my code buffer that renders a white line and function header. When I jump between chats, this line and header appear across the entire screen for a few milliseconds. I didn't notice this thing in the previous version.

Gif:
GifCode

olimorris:
That seems really odd as it shouldn't technically draw the lines. I've just remove some debug statements that were there purely for testing...that's technically an I/O request if you have the log level at `debug` (I need to make that async!). Let me know if it continues.

niba:
@olimorris  so I tested it this morning and this is still happenning

I manage to resolve it by changing the order of operations

```lua
    chats[index].chat:hide()
    chats[next].chat:open()

    to =>

    chats[next].chat:open()
    chats[index].chat:hide()
```

I also noticed that the width of my chat window jumps. You should probably try to read the current width and height of the window (if the chat is visible) and pass it to the open function, then try to reuse it inside"

olimorris:
Bizarrely I only find the jumping occurs when the order is `open` and then `hide`. If you do `hide` and `open` it seems to be fine.

I believe that's because it's trying to open a chat buffer in the already open split. I'd advise playing around with the minimal.lua config in the bug report and seeing if that works. Could be a conflicting plugin. Also, I'm testing this on Nightly so may be something upstream in Neovim that is fixed.

niba:
Hmm, that's weird. I want to clarify:
We have two different problems:

1. Flickering Treesitter context data
   - This is fixed by swapping hide and open.
2. Width and height of the chat window jumps
   - This started appearing when I swapped hide and open. I hadn't realized that I caused this bug, sorry.
  
When you mentioned 'jumping,' did you mean the second problem?

PR #94, opened 2024-08-14
feat(inline): support stop_context_insertion  logic in inline strategy

SDGLBL:
support stop_context_insertion  logic in inline strategy

Sometimes we need modify the visual selected code sended to model.


olimorris:
Ahhh good spot! Thank you

PR #95, opened 2024-08-17
fix(inline): fix issue with repeated delete keymap

SDGLBL:
image

image


olimorris:
Yes good spot. I shouldn't leave these groups around.

PR #97, opened 2024-08-17
fix(inline): Add safe_undojoin function and integrate into inline str…

SDGLBL:
Snipaste_2024-08-18_00-15-19

I still don't know how this bug is triggered. Currently, I suspect it's caused by the custom cmp configuration. However, the fix should be effective.

```lua
-- This is a better implementation of `cmp.confirm`:
--  * check if the completion menu is visible without waiting for running sources
--  * create an undo point before confirming
-- This function is both faster and more reliable.
---@param opts? {select: boolean, behavior: cmp.ConfirmBehavior}
function M.confirm(opts)
  local cmp = require "cmp"
  opts = vim.tbl_extend("force", {
    select = true,
    behavior = cmp.ConfirmBehavior.Insert,
  }, opts or {})
  return function(fallback)
    if cmp.core.view:visible() or vim.fn.pumvisible() == 1 then
      utils.create_undo()
      if cmp.confirm(opts) then
        return
      end
    end
    return fallback()
  end
end
```

### DIFF
--- a/lua/codecompanion/strategies/inline.lua
+++ b/lua/codecompanion/strategies/inline.lua
@@ -1,6 +1,7 @@
 local adapters = require("codecompanion.adapters")
 local client = require("codecompanion.client")
 local config = require("codecompanion").config
+local util = require("codecompanion.utils.util")
 
 local hl = require("codecompanion.utils.highlights")
 local keymaps = require("codecompanion.utils.keymaps")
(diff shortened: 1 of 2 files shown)


olimorris:
Thanks for this.

This is a really strange one. I have never seen this occur in the plugin and I use the inline assistant all the time.

I'm going to update the CONTRIBUTING guide to say I'll only accept PRs if we have a matching bug report. Otherwise I'm just adding code with no record of how or why the bug occurs.

SDGLBL:
Sure, if I find the cause, I'll submit a bug issue.

PR #119, opened 2024-08-25
feat: Impl slash command like zed

SDGLBL:
Please note that this is only a demonstration pull request. It is not intended to be merged into this codebase, but rather to provide a possible approach, as I am not certain of the best way to implement this feature.

## Description

Impl slash command like zed

### now

CleanShot 2024-08-25 at 23 15 05

### file

CleanShot 2024-08-25 at 23 16 37

### diagnostics

CleanShot 2024-08-25 at 23 18 14

### prompt

CleanShot 2024-08-25 at 23 22 02

### terminal

Not impl yet

## Checklist

- [ x] I've read the [CONTRIBUTING](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md) guidelines.

### DIFF
--- a/lua/cmp_codecompanion/slash_command.lua
+++ b/lua/cmp_codecompanion/slash_command.lua
@@ -1,0 +1,101 @@
+local source = {}
+
+function source.new()
+  return setmetatable({}, { __index = source })
+end
+
+function source:is_available()
+  local chat = require("codecompanion.strategies.chat").last_chat()
+  return vim.bo.filetype == "codecompanion" and chat ~= nil and chat.bufnr == vim.api.nvim_get_current_buf()
+end
+
+source.get_position_encoding_kind = function()
+  return "utf-8"
+end
+
+function source:get_trigger_characters()
+  return { "/", " " }
+end
+
+---@param params cmp.SourceCompletionApiParams
+---@param callback fun(response: CodeCompanion.SlashCommandCompletionResponse|nil)
+function source:complete(params, callback)
+  local input = params.context.cursor_before_line
+  local chat = require("codecompanion.strategies.chat").last_chat()
+
+  if not chat or not chat.slash_command_manager then
+    return callback()
+  end
+
+  local command = input:match("/(%w+)")
+  local cmd = chat.slash_command_manager:get(command)
+
+  if command and cmd then
+    if input:match("/" .. command .. "%s$") then
+      if cmd and cmd.complete then
+        return cmd:complete(params, callback)
+      end
+    end
+
+    return callback()
+  else
+    local items = {}
+
+    for name, c in pairs(chat.slash_command_manager.commands) do
+      local item = {
+        label = name,
+        kind = require("cmp").lsp.CompletionItemKind.Function,
+        documentation = c.description,
+        -- Because when executing source:execute, the corresponding command needs to be obtained through slash_command_name
+        -- For the 'now' and 'terminal' commands, since they are executed directly, slash_command_name needs to be set
+        -- Ensure that the corresponding command:execute is called
+        slash_command_name = (name == "now" or name == "terminal") and name or nil,
+      }
+
+      table.insert(items, item)
+    end
+
+    return callback({ items = items, isIncomplete = false })
+  end
+end
+
+---Executed after the item was selected.
+---@param completion_item CodeCompanion.SlashCommandCompletionItem
+---@param callback fun(completion_item: CodeCompanion.SlashCommandCompletionItem|nil)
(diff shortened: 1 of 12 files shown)


olimorris:
This is looking awesome btw! Thanks so much for your hardwork. Hoping I can begin implementing this in the next couple of days.

S1M0N38:
Great work! However, it seems that the prompts you are using are not the same as those used in Zed. The Zed team collaborates with engineers at Anthropic (https://zed.dev/blog/zed-ai), so I assume they are trying to use best practices for effectively prompting LLMs.

I have intercepted some LLM requests sent by Zed and collected notes about their slash commands, which you can find here:
https://gist.github.com/S1M0N38/1c9c333874e1061d199726c28df169ae
(Note: the /workflow command is missing at the moment.)
I hope this will help refine the implementation of slash commands.

If you are interested in learning how requests can be intercepted, feel free to DM me on Discord: https://discord.com/users/S1M0N38#0317

olimorris:
Very, interesting. So they're sending over the content in much the same way we are right now. Though I note we send over line numbers too...

olimorris:
https://github.com/user-attachments/assets/e11628b8-4d12-4d6b-a7cc-ae6f919154a2

This PR has been an awesome reference. Thank you!

My current plan is to introduce a new section in the config:

```lua
chat = {
  slash_commands = {
    ["buffer"] = {
      callback = "helpers.slash_commands.buffer",
      description = "Share a buffer's contents with the LLM",
      key = "Buffer",
      opts = {
        icon = "",
        contains_code = true,
      },
    },
  },
}
```

This would allow users to extend and build their own slash commands, and ideally only having to build one function. I'll then use Tree-sitter to parse the message and get the block quotes.

For the `/files` slash command I think it makes sense to leverage Telescope with a code preview window.

dsully:
Please include `fzf-lua` support as well as `telescope` if possible.

SDGLBL:
> 2024-08-31.16_42_50.-.WezTerm.mp4 
> 
> This PR has been an awesome reference. Thank you!
> 
> My current plan is to introduce a new section in the config:
> 
> ```lua
> chat = {
>   slash_commands = {
>     ["buffer"] = {
>       callback = "helpers.slash_commands.buffer",
>       description = "Share a buffer's contents with the LLM",
>       opts = {
>         icon = "",
>         contains_code = true,
>       },
>     },
>   },
> }
> ```
> 
> This would allow users to extend and build their own slash commands, and ideally only having to build one function. I'll then use Tree-sitter to parse the message and get the block quotes.
> 
> For the `/files` slash command I think it makes sense to leverage Telescope with a code preview window.

Personally, I prefer the cmp implementation because it operates exactly like code completion in the code buffer, which feels more intuitive to me. However, I acknowledge that this approach might be more cumbersome for users who want to extend functionality. Therefore, I’ve also introduced ⁠BaseSlashCommand to facilitate easier extensions. 

 In fact `lua/codecompanion/slash_commands/prompt.lua`, it’s quite a simple implementation.

olimorris:
> Personally, I prefer the cmp implementation because it operates exactly like code completion in the code buffer, which feels more intuitive to me. However, I acknowledge that this approach might be more cumbersome for users who want to extend functionality. Therefore, I’ve also introduced ⁠BaseSlashCommand to facilitate easier extensions.

My concern is how it scales to repositories with lots of files. Telescope or `vim.ui.select` make it much easier without needing to run jobs in the background.

I love your approach to folding the code in the chat buffer, like Zed AI. I just can't find a way to make `manual` folds flexible. The minute you insert code above the fold, it breaks. I'm trying to leverage tree-sitter for this but have spent the last 1 hour failing at that. I want to keep the chat buffer as clean as possible so folding is essential to me.

Once again, thanks for this PR. It's been a lighthouse for me. I'm hoping to release slash commands tomorrow.

olimorris:
@dsully - I've just added `fzf_lua` support. Can you try it out?

olimorris:
@S1M0N38 - I'm going to get around to adding the other slash commands at some point soon. Thanks for the gist.

dsully:
A cursory check looks good. Thank you!

olimorris:
I've now adde in the `terminal` slash command. I went for a different approach whereby I only pull in the last terminal buffer's worth of content

olimorris:
Thanks for all the hard work that went into this PR. As I said before, it was a wonderful reference to use so early on in the implementation of slash commands.

With version `8.0.0` adding the ability to have slash commands from the prompt library then I'll close this PR down as I think slash commands are exactly where I want them to be.

PR #148, opened 2024-08-29
feat: send keymap use submit function instead of writing buffer

itsfrank:
## Description

related discussion: #147

The buffer sending on write is nice and convenient, but forcing all implementations of send to use this mechanism can be limiting.

Implementing send using the `submit` function on the Chat object makes it possible to set `buftype=nofile` which greatly improves the experience for users of the `hidden` setting

Currently, with `set hidden`, when the chat window is toggled off after being modified, quitting neovim will be blocked by the `No write since last change` error which is very annoying, as the user must now force quit, or re-open the chat buffer and close it definitively (e.g. using `<C-c>`)

With this change, my toggle keymap looks like this and I get no issues when closing neovim:
```lua
vim.keymap.set("n", "<leader>gpt", function()
    codecompanion.toggle()
    local current_buffer = vim.api.nvim_get_current_buf()
    local buffer_name = vim.api.nvim_buf_get_name(current_buffer)
    if string.find(buffer_name, "%[CodeCompanion%]") then
        vim.api.nvim_buf_set_option(0, "buftype", "nofile")
    end
end)
```

## Checklist

### DIFF
--- a/lua/codecompanion/keymaps.lua
+++ b/lua/codecompanion/keymaps.lua
@@ -161,8 +161,8 @@ M.options = {
 }
 
 M.send = {
-  callback = function()
-    vim.cmd("w")
+  callback = function(chat)
+    chat:submit()
   end,
 }
 


olimorris:
Absolutely down for this. I'll try it out tomorrow.

olimorris:
Presume we can remove the set hidden, and set the buftype=nofile with this approach?

itsfrank:
> Presume we can remove the set hidden, and set the buftype=nofile with this approach?

`set hidden` is part of my config and behavior I like neovim to have. `buftype=nofile` is an option I want the chat buffer to have so that neovim doesn't warn me when I close it without saving (in fact it makes it impossible to save the buffer at all).

Those two options are my personal preferences, the reason I submitted this change, is so that the chat buffer still works if a user has chose to set  `buftype=nofile` in the buffer.

An easy way to test this out, is to open that chat window and run:
`:lua vim.api.nvim_buf_set_option(0, "buftype", "nofile")`

So if this PR is merged, I would still keep the keymap I have above, which is applying my personal preferences to the buffer.

If other people want this behavior as well they can use a similar keymap, it might also be possible to do this with some autocmd but this worked for me.

In the linked discussion I suggested as a follow-up it could be interesting to discuss adding a config setting that would make the chat buffer `nofile`. If that were added then I could remove everything from my keymap and just use `:CodeCompanionToggle` directly

itsfrank:
Just tested, and interestingly, even after making the buffer `'nofile'`, the `:w` behavior still works, sending the chat to the LLM!

olimorris:
So doing the following:

- Changing the keymap as per this PR
- Removing the line `api.nvim_buf_set_option(bufnr, "buftype", "acwrite")` from chat.lua
- And removing the `BufWriteCmd`:
```lua
  api.nvim_create_autocmd("BufWriteCmd", {
    group = aug,
    buffer = bufnr,
    desc = "Submit the CodeCompanion chat buffer",
    callback = function()
      self:submit()
    end,
  })
```

Keeps the chat buffer hidden from my buffer lists, keeps functionality of the chat buffer the same and stops the annoying error message `E162: No write since last change for buffer "[CodeCompanion] 2646937"`

itsfrank:
Apologies if I misunderstood you in my last reply!

Yes, I think that should work, if you want to put the change on a branch I'm happy to try it out.

FWIW I'm not sure it's necessary to remove the `BufWriteCmd`? In my branch `:w` still sends to the LLM, if this is an interaction type you think is desirable it may be possible to keep it.

olimorris:
Awesome. Have done so [here](https://github.com/olimorris/codecompanion.nvim/tree/refactor/change-buftype).

itsfrank:
Been using these changes locally for a few days and they work like a charm. Of the removal of bufwrite doesn't break any existing flows then I think this is a great change!

olimorris:
Thanks so much for testing it out

PR #155, opened 2024-08-31
fix(keymaps): update next and previous header keybindings

S1M0N38:
## Description

Updated the keybindings for navigating to the next and previous headers from `]` to `]]` and from `[` to `[[` respectively in both the README and configuration files.

It's pretty common to use `[` and `]` as first char in Vim/Neovim motion keymaps. So the editor usually pause for a couple of milliseconds before process `[` and `]` as standalone motion.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -257,7 +257,7 @@ require("codecompanion").setup({
         },
         next_header = {
           modes = {
-            n = "]",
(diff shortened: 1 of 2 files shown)


olimorris:
Great spot! Thank you

Issue #197, opened 2024-09-10
[Bug]: Autocomplete for slash-command


S1M0N38:
### Your `minimal.lua` config

Same minimal.lua provided in the issue template

```lua
local root = vim.fn.fnamemodify("./.repro", ":p")

-- set stdpaths to use .repro
for _, name in ipairs({ "config", "data", "state", "cache" }) do
  vim.env[("XDG_%s_HOME"):format(name:upper())] = root .. "/" .. name
end

-- set config path to enable the copilot adapter to work
vim.env["CODECOMPANION_TOKEN_PATH"] = vim.fn.expand("~/.config")

-- bootstrap lazy
local lazypath = root .. "/plugins/lazy.nvim"
if not vim.loop.fs_stat(lazypath) then
  vim.fn.system({
    "git",
    "clone",
    "--filter=blob:none",
    "--single-branch",
    "https://github.com/folke/lazy.nvim.git",
    lazypath,
  })
end
vim.opt.runtimepath:prepend(lazypath)

-- Your CodeCompanion setup
local plugins = {
  {
    "olimorris/codecompanion.nvim",
    dependencies = {
      { "nvim-treesitter/nvim-treesitter", build = ":TSUpdate"},
      "nvim-lua/plenary.nvim",
      "hrsh7th/nvim-cmp",
      {
        "stevearc/dressing.nvim",
        opts = {},
      },
      "nvim-telescope/telescope.nvim",
    },
    opts = {
      -- Refer to: https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/config.lua
      strategies = { -- Change the adapters as required
        chat = { adapter = "openai" },
        inline = { adapter = "openai" },
        agent = { adapter = "openai" },
      },
      opts = {
        log_level = "DEBUG",
      },
    },
  }
}

require("lazy").setup(plugins, {
  root = root .. "/plugins",
})

-- Setup Tree-sitter
local ok, treesitter = pcall(require, "nvim-treesitter.configs")
if ok then
  treesitter.setup({
    ensure_installed = { "lua", "markdown", "markdown_inline", "yaml" },
    highlight = { enable = true },
  })
end

-- Setup completion
local ok, cmp = pcall(require, "cmp")
if ok then
  cmp.setup({
    mapping = cmp.mapping.preset.insert({
      ["<C-b>"] = cmp.mapping.scroll_docs(-4),
      ["<C-f>"] = cmp.mapping.scroll_docs(4),
[...]

PR #200, opened 2024-09-11
ensure callback execution in source:execute

DanielEliasib:
## Description

Callback execution in source.execute() for slash commands so that cmp ends correctly.

## Related Issue(s)

#197 

## Screenshots

image

## Checklist

### DIFF
--- a/lua/cmp_codecompanion/slash_commands.lua
+++ b/lua/cmp_codecompanion/slash_commands.lua
@@ -54,6 +54,9 @@ end
 ---@return nil
 function source:execute(item, callback)
   vim.api.nvim_set_current_line("")
+  if callback then
+	  callback()
+  end
   return SlashCommands:execute(item)
 end
 


S1M0N38 (changes requested):
I think that this is sufficient
```lua
---Execute selected item
---@param item table The selected item from the completion menu
---@param callback function
---@return nil
function source:execute(item, callback)
  vim.api.nvim_set_current_line("")
  SlashCommands:execute(item)
  callback(item)
end
```
No, need to check for nil value.

As example, from cmp docs:
```lua
---Executed after the item was selected.
---@param completion_item lsp.CompletionItem
---@param callback fun(completion_item: lsp.CompletionItem|nil)
function source:execute(completion_item, callback)
  callback(completion_item)
end
```
---
- Please reword the commit following conventional commits standards and ref issue #197 
- Use 2 "spaces" for indedetion

olimorris:
Just tested this and works perfectly. Awesome fix. Thanks both.

I'll merge as-is and sort the indentation later

PR #210, opened 2024-09-12
feat: adding mini.diff module as inline diff mechanism

bassamsdata:
## Description

Adding the Mini.diff module as an option for the inline feature, offering an alternative to Neovim's original diff mode, which works between two buffers. With this, the diff will appear inline in the same buffer.

## Related Issue(s)

The initial discussion happened here: #154  
The PR is not finished yet; this is just a starting point. I placed the module in a new file due to its size, making it easier to maintain.

Currently, this PR has the following issues that aren’t working properly:
- The diff starts but seems to use the wrong original text. It appears the variable `original_buffer_content` doesn’t capture the correct buffer to compare against (possibly can be retrieved from `self.context.lines`).
- As a result, the accept and reject functions don’t work as expected.

## Checklist


lua/codecompanion/config.lua
olimorris, line 1: I think we should give users the choice of a provider i.e. `default` or `mini_diff`. I think we should hardcode the `revert_delay` in too.

olimorris:
This looks really promising. Very exciting to have in the plugin.

Once you're happy with the PR I'll merge into `main` and then at some point in the future refactor to a Diff class and ensure both the `default` provider and `mini_diff` implement the same methods.

lua/codecompanion/config.lua
bassamsdata: I added a setting called `config.display.inline.diff.diff_method` that can be set to either `default` or `mini_diff`. I'm not sure if this is what you asking for. 

I also removed the `revert_delay` setting and replaced it with a hardcoded value.

bassamsdata:
This PR is ready unless there're other reviews. Here are some notes:

1. All features like diff against LLMs, `gr` or `ga` for reject and accept, work perfectly.
2. I've added some APIs, like `force_git` and `force_codecompanion`, and an API for the statusline diff source (Readme updated).
3. I've added comments to make the code easier to understand, as well as Lua annotations.
> at some point in the future refactor to a Diff class and ensure both the default provider and mini_diff implement the same methods.
4. I agree. In the future, this module should be more integrated into the plugin. (I've already started a refactor but it will take some time).
5. I'll be adding the `mini.nvim` module because some people use the entire mega plugin. (This will be in a separate PR since we need it for `mini.pick` as well.)

Thank you!

S1M0N38:
Great work!

However, I am wondering why CodeCompanion decides to define custom keymaps for reject and accept (`gr` and `ga`) instead of using the [default ones](https://neovim.io/doc/user/diff.html#_4.-diff-copying) for operating with diff, i.e. `do` and `dp`.

`dp` does not reject a suggestion but updates the suggestion with the content of the original buffer. This is a way to update the chat context and sync it to the content of the user buffer, so that future requests are based on the correct file content.

Thoughts?

olimorris:
@bassamsdata this look's awesome. Only non-value add ask is if we can rename references such as `miniDiff` to `mini_diff` for consistency reasons?

@S1M0N38 haha because the author has absolutely no idea about using diff's in Neovim 😆.

bassamsdata:
> if we can rename references such as miniDiff to mini_diff for consistency reasons?

Sure, I made the change right away. I'm juggling multiple things at the moment, including Python, R, Golang, and Lua, so my brain got a bit mixed up since each language prefers a different naming convention. My bad!
So now, everything related to this module is named `mini_diff`. I kept the `mini.diff` plugin as `MiniDiff` since it's part of the plugin’s internal structure, and to help distinguish it from this module.

@S1M0N38 This is cool! I’ve been using Neovim for 2 years, but I haven’t really done any diffing in the editor. It feels like there’s always something new to learn in Neovim.

olimorris:
@bassamsdata thanks so much. I'll merge now. I've been using Neovim for the best part of 4 years and never even knew diffing could be done natively until I saw @S1M0N38's dante plugin the other week 😆.

olimorris:
Hey @bassamsdata...I've reworked how we handle diff's in version 3.0.0 of the plugin...the result is I've broken your `mini_diff` integration so I will endeavor to fix that in the coming days!

PR #212, opened 2024-09-12
feat: add /help slash-command

S1M0N38:
## Description

I've start to work on implementing the simple version of the `/help` slash command as described in discussion https://github.com/olimorris/codecompanion.nvim/discussions/207.

This is the simple version (not the rag version), that simply use tags to select the docs file.

Some doc/*.txt can be quite long so I'd like to select only a couple of subsetion of the whole file (e.g. the first section and the tag's section). 

I'm thinking about filtering using regex or make use of treesitter. Feedbacks?

## Demo

https://github.com/user-attachments/assets/7bcd4c62-52c9-494e-bbf1-51b7db1d3129

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -144,6 +144,7 @@ _Slash commands_, accessed via `/`, run commands to insert additional context in
 - `/file` - Insert a file
 - `/now` - Insert the current date and time
 - `/symbols` - Insert symbols for the active buffer
+- `/help` - Insert content from help tags
 
 _Tools_, accessed via `@`, allow the LLM to function as an agent and carry out actions:
 
@@ -322,6 +323,13 @@ require("codecompanion").setup({
         ["symbols"] = {
           callback = "helpers.slash_commands.symbols",
           description = "Insert symbols for the active buffer",
+          opts = {
+            contains_code = true,
+          },
+        },
+        ["help"] = {
+          callback = "helpers.slash_commands.help",
+          description = "Insert content from help tags",
           opts = {
             contains_code = true,
           },
--- a/doc/codecompanion.txt
+++ b/doc/codecompanion.txt
@@ -122,6 +122,7 @@ into the chat buffer:
 - `/file` - Insert a file
 - `/now` - Insert the current date and time
 - `/symbols` - Insert symbols for the active buffer
+- `/help` - Insert content from help tags
 
 _Tools_, accessed via `@`, allow the LLM to function as an agent and carry out
 actions:
--- a/lua/codecompanion/config.lua
+++ b/lua/codecompanion/config.lua
@@ -74,6 +74,13 @@ return {
             contains_code = true,
           },
         },
+        ["help"] = {
+          callback = "helpers.slash_commands.help",
+          description = "Insert content from help tags",
+          opts = {
+            contains_code = true,
+          },
+        },
       },
       keymaps = {
         options = {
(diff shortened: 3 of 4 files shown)


olimorris:
Firstly, I am very, very excited by this. I actually could have done with this when I was troubleshooting cmp issues the other day.

Regarding length of the help file, that won't matter if you're using Copilot or Anthropic. I can't think how you'd parse the help file and workout which parts to omit. I have a crude [tokens](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/utils/tokens.lua) file which allows you to estimate how many tokens might be consumed by a message. Would displaying a warning to a user if a helpfile exceeds a threshold be worthwhile?

S1M0N38:
Yep, the idea come to me when I was bored to paste the content of help file into codecompletion chat when developing/debugging Neovim plugin. 

[token](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/utils/tokens.lua) seem pretty useful.

Btw, not mentioned in the discussion, by compare to vscode all the installed plugin docs is searchable.

---

I'd like to test a couple of implementations so do not expect quick results. I think that provide the right ctx in chat is really important for this use case.

S1M0N38:
https://github.com/user-attachments/assets/0969a5b5-fdb8-49f3-8e39-6702d581eccc



I'm trying to extract the selected tag's section using treesitter query

```scm
(help_file
  (block
    (line
      (h1) @start))
  (block
    (line
    (tag) @tag
    (#eq? @tag "*lazy.nvim-🚀-getting-started-✨-features*")))
  (block
    (line
      (h1) @end))
)
```

Im testing this on the file opened by `:help lazy.nvim-🚀-getting-started-✨-features`
(doc/lazy.nvim.txt)

### Issue

However I'd like to select just the previous and next h1 relative to the matched tag. Right now the query match all h1 before, then the tag and finally all h1 after. Any idea how to refine this query?

I think that including tag's section with /help slash cmd it's a good trade off between overload the llm ctx with all the file and send to little ctx.

olimorris:
My experience with Tree-sitter is you're best to parse the whole document and obtain all of the matches to your query, before you filter them down to say the previous and next occurrences

Looking at the tree for a help files, it doesn't look like the hierarchy easily supports selecting everything under a tag as everything is just a `block`.

Perhaps you could have a query like:

```scm
(
 (heading)
 (tag) @tag
)
(block) @block
```

and you process all of the blocks until you come to a tag which resides next to a heading?

S1M0N38:
At the end I avoid the headers shenanigans and use TS just to get the tag row. 
I think that providing a `MAX_LINES` is good enough.

Using TS to parse and select the content to sent to LLM can be quite challenging:
- documentation is written in different format (some are markdown) 
- and different styles (plugin authors preferences)

olimorris:
Firstly this looks awesome! Will be a worth edition to the plugin.

I want to send over some context about `nvim-cmp`. So Telescope opens and I type `nvim-cmp` and I select the first entry:

2024-09-13 19_13_12 - WezTerm@2x

Do the dots at the bottom of the output indicate the help file has been truncated?

2024-09-13 19_14_51 - WezTerm@2x

(I've purposefully not looked at the code so I can observe it with clean eyes 😄)

S1M0N38:
Yes, if only a subset of lines is sent to the buffer, `...` is added at the beginning and the end. A slight improvement could be to remove `...` when hitting file boundaries (i.e., the first and last lines).

olimorris:
Thanks again for this! A fantastic addition to the plugin.

PR #220, opened 2024-09-14
fix(adapter): add notification for missing GitHub Copilot token

jellydn:
## Description

- Show error message if missing Github token
- Rework the logic for getting Github token (it doesn't work on my setup even CopilotChat.nvim works fine)

## Checklist


lua/codecompanion/adapters/copilot.lua
```diff
@@ -10,55 +10,55 @@ local _oauth_token
...
-    local f = io.open(path, "r")
-    if not f then
-      return log:error("Could not open file: %s", path)
-    end
+---Get the Copilot OAuth token
+--- The function first attempts to load the token from the environment variables,
+--- specifically for GitHub Codespaces. If not found, it then attempts to load
+--- the token from configuration files located in the user's configuration path.
+---@return string|nil
+local function get_github_token()
```
olimorris, line 40: It looks like we're hitting the filesystem every time we make a request with this adapter?

Previously I'd set the token to a locally scoped var at the top of the page and add a check:

```lua
if _oauth_token then
  return _oauth_token
end
```

lua/codecompanion/adapters/copilot.lua
```diff
@@ -129,9 +129,13 @@ return {
     ---@param self CodeCompanion.AdapterArgs
     ---@return boolean
     setup = function(self)
-      _oauth_token = get_oauth_token()
+      _oauth_token = get_github_token()
       if not _oauth_token then
         log:error("No GitHub Copilot token found. Please refer to https://github.com/github/copilot.vim")
+        vim.notify(
```
olimorris, line 135: Let's remove the `vim.notify` as the user is already being notified via `log:error` (see `utils.log`).

cspell-tool.txt
olimorris, line 1: No idea what these. Please remove.

olimorris:
Thanks for the PR. I've added in some comments. Main concern is with `get_github_token`. I think we should be storing the token to a variable to prevent expensive file reads.

jellydn:
Thank you @olimorris for quick review.

>  1) The Copilot adapter not working
I don't think so. As I mentioned, it works fine for me with `CopilotChat.nvim`

> 2) Changing the notification type for log:error.

The reason I changed to notify as I didn't know what's going on until I checked the log

```sh
- Log file: /Users/huynhdung/.local/state/nvim/codecompanion.log
```

lua/codecompanion/adapters/copilot.lua
jellydn: I have just added this back @olimorris 

> Previously I'd set the token to a locally scoped var at the top of the page and add a check

What if the token expired?

lua/codecompanion/adapters/copilot.lua
jellydn: I would prefer to keep this as it's notified to user right away without checking the log. Let me know if that's okay or have to drop this change. Thanks.

lua/codecompanion/adapters/copilot.lua
olimorris: I believe the lifespan of the oauth token is looooong. Mine is over two years old! So the likelihood of incurring this seems low.

lua/codecompanion/adapters/copilot.lua
olimorris: The user will be getting doubley notified so please remove. `Log:error` outputs the error to the user using `vim.api.nvim_echo`.

olimorris:
> Thank you @olimorris for quick review.
> 
> > 1. The Copilot adapter not working
> 
> I don't think so. As I mentioned, it works fine for me with `CopilotChat.nvim`

I'm referring to the Copilot adapter in CodeCompanion.

olimorris:
Awesome thanks for this. Really appreciate you taking the time to solve this. Imagine there have been a few other people that have been caught out by it.

Could I just ask you to drop the README.md changes? then I can merge this.

jellydn:
Sure. I will drop that change when I am back to my desk.

PR #221, opened 2024-09-14
Update minimal.lua in bug report

S1M0N38:
Using lazy.nvim for repro as suggested in the [official documentation](https://lazy.folke.io/developers#reprolua)


olimorris:
Awesome thanks for this. Really cleans everything up.

I'm hoping to get Version 3.0 released soon and I want the PR for it to be clean so I'll leave this unmerged until then.

PR #230, opened 2024-09-16
feat: add back mini.diff

olimorris:
https://github.com/user-attachments/assets/f286893f-1d24-4e87-9ab1-8b499fad11cf

@bassamsdata - This is the output from my initial attempt to add back `mini.diff`. Is this how you expect it to look?

### DIFF
--- a/README.md
+++ b/README.md
@@ -1221,6 +1221,8 @@ The plugin fires many events during its lifecycle:
 - `CodeCompanionInlineFinished` - Fired at the end of the Inline strategy
 - `CodeCompanionRequestStarted` - Fired at the start of any API request
 - `CodeCompanionRequestFinished` - Fired at the end of any API request
+- `CodeCompanionDiffAttached` - Fired when in Diff mode
+- `CodeCompanionDiffDetached` - Fired when exiting Diff mode
 
 > [!TIP]
 > Some events are sent with a data payload which can be leveraged.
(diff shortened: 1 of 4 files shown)


bassamsdata:
Thanks for reworking it. The video is a bit short, but if the LLM refactored it by deleting all the function logic (shown in the red lines), then yes, it worked.

I’m noticing that the integration has much less code now, so I’m curious to try it. Usually, if someone is using mini.diff (which is almost a direct replacement for gitsigns.nvim), we use it to diff against the git source. The previous implementation automatically switched between codecompanion inline diff and git diff.

bassamsdata:
Thank you again for the heads-up!

I tested this PR a bit, and it’s working as expected. However, I have two small requests.

Could you please add back a buffer variable like `vim.b[bufnr].diffCompGit`, so we can know what the source of the diff is (either git or LLM)? It’s really useful, and I use it in my statusline to track which diff I’m comparing against.

I also noticed that we return to git diff after the window loses focus, which is great. Could we have the same behavior happen after accepting or rejecting the changes please?

bassamsdata:
thank you for adding these additions. they work flawlessly.

olimorris:
Glad to hear it. I've actually started using mini.diff myself, so nice being able to see the diff inline and so cleanly.

PR #231, opened 2024-09-17
Allow to customize prompt

chrishrb:
## Description

I don‘t use dressing for inputs so I have the problem that the input is missing a space at the end of the "Prompt" input. A solution would be to just add the space everywhere, but I think it's also fun to have the ability to customize the prompt field. I also deleted the warning for the input, mainly to stay consistent for all inputs.

## Screenshots

The current implementation doesn‘t look good if you don‘t have dressing enabled for input fields:
image

## Checklist

### DIFF
--- a/lua/codecompanion/config.lua
+++ b/lua/codecompanion/config.lua
@@ -601,6 +601,7 @@ Use Markdown formatting and include the programming language name at the start o
     action_palette = {
       width = 95,
       height = 10,
+      prompt = "Prompt ", -- Prompt used for interactive LLM calls
     },
     chat = {
       window = {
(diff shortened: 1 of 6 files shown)


lua/codecompanion/config.lua
```diff
@@ -643,6 +643,9 @@ Use Markdown formatting and include the programming language name at the start o
   opts = {
     log_level = "ERROR", -- TRACE|DEBUG|ERROR|INFO
 
+    -- Prompt used for interactive LLM calls
+    command_prompt_prefix = "Prompt ",
```
olimorris, line 647: Can we move this to `display.action_palette` and just call it `prompt`?

olimorris:
Hi @chrishrb this is a great spot, thank you. Just a minor comment regarding the location of the prompt in the config

chrishrb:
resolved and while testing it I also found the bug that the value of prompt is evaluated too early if you require the config in the file, so I moved it to inline. It doesn‘t look that good, but it works 👍

olimorris:
Awesome thanks for this!

PR #232, opened 2024-09-17
Fix diff provider in editor tool

chrishrb:
## Description

The `config.display.inline.diff.provider` option is nil, so I guess it should be `config.opts.diff.provider`

## Screenshots

image

## Checklist


olimorris:
Great spot! Thanks

PR #241, opened 2024-09-19
fix(chat): don't index an empty table

MariaSolOs:
## Description

With cd6ac62a82e98527850ad419ef59165197629f10, a null access happens when running `:CodeCompanionChat`:
image

This PR fixes that by making sure that `fargs` isn't empty before indexing into it.

## Checklist


olimorris:
Great spot, thanks @MariaSolOs!

PR #250, opened 2024-09-22
feature(pickers): Add telescope picker for codecompanion actions

Ajaymamtora:
## Description

  As described here: https://github.com/olimorris/codecompanion.nvim/discussions/248 -- I would prefer the ability to configure the layout of the telescope picker, which cant be done currently. The current implementation seems to use vim.ui.select which I've overriden with Telescope using dressing.nvim. 
  
If we create a telescope picker and export it as an extension, you can pass opts directly to the picker and configure the layout/bindings etc.

My default telescope layout takes up most of the screen, but for pickers with not many options like codecompanion actions, I want to reduce the size.

## Related Issue(s)
https://github.com/olimorris/codecompanion.nvim/discussions/248

## Screenshots

image

As you can see the layout is now as I wanted when compared to the discussion above, using:

```lua
local small_no_preview = {
  layout_strategy = "vertical",
  previewer = false,
  layout_config = {
    vertical = {
      size = {
        height = "40%",
        width = "40%",
      },
    },
  },
}
```

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ There are keymaps available to accept or reject edits from the LLM in the [inlin
 
 <!-- panvimdoc-ignore-end -->
 
-Run `:CodeCompanionActions` or `:CodeCompanionTelescope` to open the action palette, which gives you access to all of the functionality in the plugin. This is where core actions and the [pre-defined prompts](#clipboard-pre-defined-prompts) are listed.
+Run `:CodeCompanionActions` or `:lua require('telescope').extensions.codecompanion.codecompanion()` to open the action palette, which gives you access to all of the functionality in the plugin. This is where core actions and the [pre-defined prompts](#clipboard-pre-defined-prompts) are listed.
 
 > [!NOTE]
 > Some actions and prompts will only be visible if you're in _Visual mode_.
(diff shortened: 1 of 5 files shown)


olimorris:
Thanks @Ajaymamtora this looks epic!

I'll test this out tonight and my thinking is that perhaps this should be the default behaviour. If a user runs `CodeCompanionActions` and they have Telescope installed, then use the extension...otherwise default to the current one.

S1M0N38:
I think registering a proper extension for Telescope would be a nice touch. 

However, I'm not sure about cluttering the CodeCompanion command namespace with the command `CodeCompanionTelescope`.

It might be better to avoid creating user commands tied to specific dependencies because, following that logic, we would need to implement separate commands for different pickers (e.g., fzf-lua, mini.pick, etc.).

olimorris:
@S1M0N38 is Telescope best practice to still manually register the extension?

S1M0N38:
> @S1M0N38 is Telescope best practice to still manually register the extension?

I don't know. I was only point out some design choice about user commands.

olimorris:
Sorry I've only got around to reviewing this now.

I agree with @S1M0N38's point that I'd like to keep the command at just `:CodeCompanionActions`. I've noticed that this PR doesn't show any prompts that are [bound to visual mode](https://github.com/olimorris/codecompanion.nvim/blob/15d0aee6f8a029cb6b6c2bd34894473d117998c9/lua/codecompanion/config.lua#L321).

An easier way to obtain all of the actions is with:

```lua
local context_utils = require("codecompanion.utils.context")

local actions = require("codecompanion.actions")
local context = context_utils.get()
actions.add_default_prompt_library(context)
```

I notice that in [CopilotChat.nvim](https://github.com/CopilotC-Nvim/CopilotChat.nvim/blob/canary/lua/CopilotChat/integrations/telescope.lua) they pass the actions to a telescope function:

```lua
require("CopilotChat.integrations.telescope").pick(actions.prompt_actions({
    selection = select.visual,
}))
```

I consider that advantageous as a user doesn't need to register an extension.

olimorris:
This has made me realise that I should probably refactor the whole actions and prompts anyway...

olimorris:
@Ajaymamtora I'm in the process of massively refactoring the actions / prompts.

Should be done tomorrow (hopefully) and I'll make sure I add Telescope support.

I forgot how horrible and hacky my default pickers were...this was before I expected anyone to ever use the plugin 😂.

olimorris:
This should now be in `v4.0.0`. If you run `:CodeCompanionActions Telescope` it should open.

olimorris:
I'm wondering if it's because I didn't register the extension. Can you check #267 ? I'll also move all the conversation over there as well.

PR #267, opened 2024-09-28
fix: telescope extension is registered

olimorris:


olimorris:
Okay this should now be fixed...I forgot I needed to create the extension.

Running `:Telescope codecompanion` will load it up or you can change the default provider in the `display.action_palette.provider`

olimorris:
So is it a case of just passing in opts to the extension to allow for customisation and overrides? Like:

```lua
return require("telescope").register_extension({
  exports = {
    codecompanion = function(opts)
      return require("codecompanion").actions({ provider = { name = "telescope", opts = opts } })
    end,
  },
})
```

I may be missing the level of customisation you need as I thought that the extension will inherit the defaults you've set for Telescope like it does for me:

2024-09-29 19_52_00 - WezTerm@2x

olimorris:
Cool. Can you give me some example opts that you'd pass as part of a command then I can make sure I'm getting them right? Say `:Telescope codecompanion { layout_config = "vertical" }`

olimorris:
I've been able to get it to work with:

```lua
function()
  local opts = {
    layout_strategy = "vertical",
    previewer = false,
    layout_config = {
      vertical = {
        width = 0.7,
        height = 0.7,
      },
    },
  }

  require("telescope").extensions.codecompanion.codecompanion(opts)
end,
```

PR #270, opened 2024-09-29
fix(adapters/openai): empty content but has role in response

gh-liu:


olimorris:
Thanks for the PR.

Can you share the output from OpenAI that necessities this? I've never come across this before.

olimorris:
Also, where possible, it's my preference to have `fix` PRs that reference an issue that I can test.

gh-liu:
> reference an issue

Thanks for you reply, for now ,I can not reproduce the issue, I will open a issue when it occurs.

PR #272, opened 2024-09-30
feat: add mini pick to actions

bassamsdata:
## Description

This pull request introduces the `mini.pick` plugin as a provider for the `CodeCompanionActions` command, aligning with the Telescope integration in v4 (#260).

**Note:** The only issue with this implementation is that after selecting an item, the cursor remains in the current buffer instead of jumping to the chat buffer. I attempted several fixes without success, so it might need revisiting. However, it's not a major concern and could be addressed later by me or someone else.

Thank you!

## Related Issue(s)

No issues

## Screenshots

Screenshot 2024-09-29 at 10 09 30 PM

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -192,7 +192,7 @@ There are keymaps available to accept or reject edits from the LLM in the [inlin
 
 <!-- panvimdoc-ignore-end -->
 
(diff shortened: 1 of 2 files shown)


olimorris:
Fantastic thank you!

PR #297, opened 2024-10-06
make lazy.vim instructions more clear

hubrix:
confused me when i tried to install, since lazy recommends structured config and the config in the readme showed for the older single file config

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,7 +73,7 @@ Install the plugin with your preferred package manager:
   config = true
 }
 
--- plugins/telescope.lua:
+-- plugins/codecompanion.lua:
 return {
   "olimorris/codecompanion.nvim",
   dependencies = {
--- a/README.md
+++ b/README.md
@@ -59,7 +59,9 @@ Install the plugin with your preferred package manager:
 **[Lazy.nvim](https://github.com/folke/lazy.nvim)**
 
 ```lua
-{
+
+-- init.lua:
+    {
   "olimorris/codecompanion.nvim",
   dependencies = {
     "nvim-lua/plenary.nvim",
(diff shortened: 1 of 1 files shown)


hubrix:
I was confused trying to set this up, so once I figured out what was confusing me I tried to rewrite the instructions to be more clear as I have seen from other plugins.

olimorris:
Thanks for the PR. Can you also share the lazy guidelines you're referring to?

hubrix:
Take a look here: https://lazy.folke.io/installation

S1M0N38:
The majority of Neovim plugins listing lazy.nvim installation instruction in their readme.md do not include the `return` keyword. Even folke's plugins do not include it (e.g. [folke/tokionight](https://github.com/folke/tokyonight.nvim?tab=readme-ov-file#-installation)).

hubrix:
Take a look at the documentation on "structuring"
https://lazy.folke.io/usage/structuring

That includes the return keyword in the example. And it is required for the
structured plugins setup that lazy seems to recommend as the default. The
configuration doesn't work without it

S1M0N38:
You can have Structured Setup or Single File Setup (see https://lazy.folke.io/installation).

- When using structured setup, the config is split across multiple files, and every file is a source using the `require` keyword, so you have to return the plugin config as a Lua table.

- With a single file set, you put the plugin config table directly in the lazy.nvim setup function, and you do not need the return keyword.

I think that the README should only describe the lazy.nvim plugin config format.

hubrix:
Listen, it's your project. I was just trying to be helpful since I was
confused getting it set up given the instructions provided. Feel free to
kill the change.

olimorris:
Thanks so much @hubrix for your suggestion. For now, I'm going to leave it as-is and keep it as simple as possible.

PR #320, opened 2024-10-16
fix(tools): user tool fails to load

e2r2fx:
## Description

I'm implementing an image tool using OpenAI vision specifically. While doing so I noticed that 
currently loading a user tool with loadfile fails in line
https://github.com/olimorris/codecompanion.nvim/blob/3e0995f53757b637f54f6c4ef227cb22ec63d91c/lua/codecompanion/strategies/chat.lua#L705
due to loadfile loading as a function rather than directly as a module

For context:
* My tool file is located at `tools/image.lua` related to my config directory
* Here's the relevant piece of my config that loads the tool:
```lua
    config = function()
        strategies = {
          agent = {
            adapter = "openai",
            tools = {
              ["img"] = {
                callback = "tools.image",
                description = "Image analysis",
              },
            },
          },
          chat = {
            adapter = "openai",
            tools = {
              ["img"] = {
                callback = "tools.image",
                description = "Image analysis",
              },
            },
            keymaps = {
```

And how the module is exported:

```lua
---@class CodeCompanion.Tool
return {
  name = "img",
  cmds = {
    {
      "curl",
      "${url}",
      "-H",
      "Accept: application/json",
      "-H",
      "Content-Type: application/json",
      "-H",
      "Authorization: Bearer ${api_key}",
      "-d",
      '{ "model": "gpt-4o-mini", "messages": [ { "role": "user", "content": [ { "type": "text", "text": "${query}" }, { "type": "image_url", "image_url": { "url": "${image_url}" } } ] } ], "max_tokens": 300 }',
    },
  },
  schema = {
    tool = {
      _attr = { name = "img" },
      action = {
        url = "https://media.istockphoto.com/id/1944476521/photo/portrait-of-a-pretty-blue-merle-australian-shepherd-dog-looking-straigth-at-the-camera-with.jpg?s=1024x1024&w=is&k=20&c=cUz-C8K91qUiT5PXys3mu5cYGFLwRYShqg8jH0V89jQ=",
        query = "Describe the image",
      },
    },
```

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/tools.lua
+++ b/lua/codecompanion/strategies/chat/tools.lua
@@ -385,7 +385,7 @@ function Tools.resolve(tool)
   local ok, module = pcall(require, "codecompanion." .. callback)
   if not ok then
     -- Try loading the tool from the user's config
-    ok, module = pcall(loadfile, callback)
+    ok, module = pcall(require, callback)
   end
   if not ok then
     return log:error("Could not resolve tool: %s", callback)


olimorris:
Loadfile requires the full path. Require will look relative to the plugin's directory.

olimorris:
This doesn't seem correct.

You need to pass the full path to loadfile.

[Reference](https://www.gammon.com.au/scripts/doc.php?lua=loadfile)

e2r2fx:
even when passing full path in config, the behavior of loadfile is to return the module as a function as stated in the cited docs.

This then fails in https://github.com/olimorris/codecompanion.nvim/blob/3e0995f53757b637f54f6c4ef227cb22ec63d91c/lua/codecompanion/strategies/chat.lua#L705

because when calling `resolved.schema` directly the code is assuming `resolved` has the module instead of being a function that returns the module.

I've tried to config the tool by providing the full path and the result is the same

```lua
        strategies = {
          agent = {
            adapter = "openai",
            tools = {
              ["img"] = {
                callback = os.getenv("HOME") .. "/.config/nvim/lua/tools/image.lua",
                description = "Image analysis",
              },
            },
          },
```

PR #324, opened 2024-10-17
Use default diff in `/commit` command

GordianDziwis:
## Description

Always using the default diff is more stable, users who want to use another diff program can create a new slash command.

## Checklist


GordianDziwis:
You can define your own executables which render the diff for `git diff` in your git config.
For example [difftastic](https://github.com/Wilfred/difftastic). Those can or cannot be more useful for creating a commit message. With the `--no-ext-diff` parameter, `git diff` always uses the default diff viewer.

PR #353, opened 2024-10-23
feat(anthropic): update to latest anthropic models

GordianDziwis:
Use the [latest tag](https://docs.anthropic.com/en/docs/about-claude/models) instead of a specific Version.


olimorris:
You're on fire with these PRs 😉. Thanks again

PR #359, opened 2024-10-24
fix(diff): diff split window wrongly opened in chat window

Bekaboo:
## Description

This patch fixes the issue that the diff split is opened in chat window instead of the source window when the diff layout is "horizontal" and the chat window layout is "vertical".

## Related Issue(s)

N/A

## Screenshots

Before (diff split window opened in chat window):

Screenshot_20241024_025911

After (diff split window opened below source window):

Screenshot_20241024_030027

## Checklist


olimorris:
Great spot and thank you for the PR.

PR #365, opened 2024-10-24
feat(keymaps): `yank_code` copy focused or the last codeblock.

antosha417:
## Description
Thank you for the amazing plugin, I love it! Works amazing with the claude for me. 

I wanted to make a small improvement to the `yank code` functionality:

When cursor is inside the code block `yank code` keymap would copy focused codeblock instead of the last one. 
When user wants to copy the last one they would easily jump out of the current code block and press `gy` to copy the last codeblock.

We can also make this addition opt in or opt out in the config. 

Personally I also use `vim.highlight.on_yank`, and it is a bit sad that this is not working with the `yank code` function. 
What do you think about actually jumping to the codeblock, and yanking it with the build in vim yank instead of `vim.fn.setreg` ?

## Checklist

Decided not to open a discussion since this this a very small change and I'll spend more time opening a discussion rather than implementing it.

### DIFF
--- a/lua/codecompanion/config.lua
+++ b/lua/codecompanion/config.lua
@@ -239,6 +239,7 @@ local defaults = {
       },
       opts = {
         register = "+", -- The register to use for yanking code
+        yank_jump_delay_ms = 400, -- Delay in milliseconds before jumping back from the yanked code
       },
     },
     -- INLINE STRATEGY --------------------------------------------------------
--- a/lua/codecompanion/keymaps.lua
+++ b/lua/codecompanion/keymaps.lua
@@ -217,15 +217,31 @@ M.codeblock = {
   end,
 }
 
+---@param node TSNode to yank text from
+---@param register string register to yank to
+local function yank_node(node, register)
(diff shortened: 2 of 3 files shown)


olimorris:
Firstly, absolutely love this as an idea. Agree it's a much better workflow. I actually left this out until someone raised it as a proposal so thanks for picking it up.

> Personally I also use vim.highlight.on_yank, and it is a bit sad that this is not working with the yank code function.
What do you think about actually jumping to the codeblock, and yanking it with the build in vim yank instead of vim.fn.setreg ?

I would love to get this to work. I too use the same. We return the range from Tree-sitter so could make a visual selection of the range, then yank, then return the cursor back to the original position. Would be totally open for that

antosha417:
Hey @olimorris fixed the pr, now `vim.highlight.on_yank` also works. Please tell me what do you think.

For me it was jumping back really fast so I could not get a glimpse of the code that I'm yanking if the chat buffer is long enough, so I introduced some configurable delay there.

olimorris:
Just tried this out and it's amazing. Great job! A real value add feature

PR #367, opened 2024-10-25
fix(editor,config): use actual line breaks for new lines to avoid '\n…

### LINKED ISSUE #362: [Bug]: Special chars like `\n` not escaped
### Your `minimal.lua` config

-- Same as the provided minimal.lua config

### Error messages

N/A

### Log output

N/A

### Health check output

N/A

### Describe the bug

Special characters, e.g. `\n` in the LLM output is not escaped and is inserted as an actual line break in the buffer after applying changes. For example, when I ask CodeCompanion (Claude 3.5 Sonnet backend) to write a bash script for me, it generates:

```bash
    if [ "$format" = "json" ]; then
        json_output="{\n"
        json_output+="  \"total_arguments\": $#,\n"
        json_output+="  \"arguments\": [\n"
     ...
```

However after applying the change to the source buffer using `@editor`, it becomes:

```bash
    if [ "$format" = "json" ]; then
        json_output="{
"
        json_output+="  \"total_arguments\": $#,
"
        json_output+="  \"arguments\": [
"
```

I noticed that `\n` is deliberately "unescaped" here https://github.com/olimorris/codecompanion.nvim/blob/efe90f61b05f6c11009dc7af204ad0e201578552/lua/codecompanion/helpers/tools/editor.lua#L15

which might be the cause of the issue. Can you provide an option to disable it or can we improve the prompt to let LLMs use actual new lines instead of `\n` in the genarated code so that we don't need this trick?

### Reproduce the bug

_No response_

### Final checks

Bekaboo:
…' substitution trick (#362)

## Description

Fix #362. Turns out that we just need to tell LLM to use actual linebreaks in the response and remove the string replacement trick.

## Related Issue(s)

#362 

## Screenshots

Before:

Screenshot_20241025_014606

After:

Screenshot_20241025_014636

## Checklist

### DIFF
--- a/lua/codecompanion/config.lua
+++ b/lua/codecompanion/config.lua
@@ -833,6 +833,8 @@ You must:
 - Avoid line numbers in code blocks.
 - Avoid wrapping the whole response in triple backticks.
 - Only return code that's relevant to the task at hand. You may not need to return all of the code that the user has shared.
+- Use actual line breaks instead of '\n' in your response to begin new lines.
+- Use '\n' only when you want a literal backslash followed by a character 'n'.
 
 When given a task:
 1. Think step-by-step and describe your plan for what to build in pseudocode, written out in great detail, unless asked not to do so.
--- a/lua/codecompanion/helpers/tools/editor.lua
+++ b/lua/codecompanion/helpers/tools/editor.lua
@@ -8,17 +8,6 @@ local xml2lua = require("codecompanion.utils.xml.xml2lua")
 
 local api = vim.api
 
----Sometimes the LLM will insert new lines as `\n` which are then escaped by
----the XML library. This function unescapes them
----@param str string
----@return string
-local function unescape_breaks(str)
-  if not str then
-    return str
-  end
-  return str:gsub("\\n", "\n")
-end
-
 -- To keep track of the changes made to the buffer, we store them in this table
 local deltas = {}
 local function add_delta(bufnr, line, delta)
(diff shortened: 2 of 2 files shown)


olimorris:
Awesome. I love it when a fix is so simple 😄.

I've just asked Co-Pilot, OpenAI, Anthropic and xAI the following question:

_"If I didn't want you to use `\n` in any responses to me that include code, what should I ask you?"_

The results were:

**Co-Pilot**:
- "Provide code without using `\n`."
- "Show me how to do this without using newline characters."

**OpenAI**:
- "avoid using newline characters `\n` in code responses."

**Anthropic**:
- "use explicit newlines instead of escape sequences"

**xAI**:
- "Please do not use newline escape characters (`\n`) in your code examples"

Could you amend the system prompt to say something like:

`Use actual line breaks and newlines instead of newline escape characters (`\n`) in your responses`

That seems to cover everything from the core LLMs.

Bekaboo:
> "If I didn't want you to use \n in any responses to me that include code, what should I ask you?"
> Use actual line breaks and newlines instead of newline escape characters (\n) in your responses

But sometimes we do want to include literal `\n` in the code, for example when generating code with `printf()` as shown in the screenshot. Is it better to tell LLM when to use actual linebreaks and when to use `\n` instead of telling it to always use actual linebreaks?

olimorris:
> Is it better to tell LLM when to use actual linebreaks and when to use \n instead of telling it to always use actual linebreaks?

@Bekaboo yes good point!

PR #386, opened 2024-11-02
feat(adapters): ✨ add Azure OpenAI

strayer:
## Description

As discussed in https://github.com/olimorris/codecompanion.nvim/discussions/383 I added a new adapter using the OpenAI adapter handlers comparable to xAI and Copilot.

The adapter uses these env vars:

```lua
  env = {
    api_key = "AZURE_OPENAI_API_KEY",
    endpoint = "AZURE_OPENAI_ENDPOINT",
    api_version = "2024-06-01",
    deployment = "schema.model",
  }
```

The first two are also used in the [official OpenAI Python library](https://learn.microsoft.com/en-us/azure/ai-services/openai/quickstart?tabs=command-line%2Cjavascript-keyless%2Ctypescript-keyless%2Cpython-new&pivots=programming-language-python#create-a-new-python-application), so they should be pretty standard. The api_version shouldn't change too often, maybe only for newly released models.

I'm not sure about the deployment. Azure OpenAI does not select the model based on a parameter as OpenAI-compatible APIs usually do. Instead each model can be deployed multiple times under custom names. As far as I know usually these deployment names are set the same as the actual OpenAI model name, but I guess not everyone would do that. Because the rest of codecompanion.nvim assumes the model parameter to be set I decided to keep it in there and use it as the configuration for the deployment name.

Note that this is also very similar to the official OpenAI Python library:

```python
import os
from openai import AzureOpenAI
    
client = AzureOpenAI(
    api_key=os.getenv("AZURE_OPENAI_API_KEY"),  
    api_version="2024-02-01",
    azure_endpoint = os.getenv("AZURE_OPENAI_ENDPOINT")
    )
    
deployment_name='REPLACE_WITH_YOUR_DEPLOYMENT_NAME' #This will correspond to the custom name you chose for your deployment when you deployed a model. Use a gpt-35-turbo-instruct deployment. 
    
# Send a completion call to generate an answer
print('Sending a test completion job')
start_phrase = 'Write a tagline for an ice cream shop. '
response = client.completions.create(model=deployment_name, prompt=start_phrase, max_tokens=10)
print(start_phrase+response.choices[0].text)
```

Please check if I used the schema value replacement for the env var correctly. I was a bit confused because the docs always mentioned `schema.model.default` when talking about this functionality. I'm also not really sure what to put into the choices for the parameter - the default ones of OpenAI really don't make sense here because of the deployment stuff. I guess I should change it to type string with no default? Open to suggestions.

Example usage:

```lua
  {
    dependencies = {
      -- ...
    },
    config = {
      strategies = {
        chat = {
          adapter = "azure_openai",
        },
        inline = {
          adapter = "azure_openai",
        },
      },
      adapters = {
        azure_openai = function()
          return require("codecompanion.adapters").extend("azure_openai", {
            env = {
              api_key = 'foobar',
              endpoint = 'https://foobar.openai.azure.com/',
            },
            schema = {
              model = "gpt-4o",
            },
          })
        end,
      },
    },
  },
```

## Related Issue(s)

https://github.com/olimorris/codecompanion.nvim/discussions/383

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@
 </p>
 
 <p align="center">
-Currently supports: Anthropic, Copilot, Gemini, Ollama, OpenAI and xAI adapters<br><br>
+Currently supports: Anthropic, Copilot, Gemini, Ollama, OpenAI, Azure OpenAI and xAI adapters<br><br>
 New features are always announced <a href="https://github.com/olimorris/codecompanion.nvim/discussions/categories/announcements">here</a>
 </p>
 
@@ -28,7 +28,7 @@ Thank you to the following people:
 ## :sparkles: Features
 
 - :speech_balloon: [Copilot Chat](https://github.com/features/copilot) meets [Zed AI](https://zed.dev/blog/zed-ai), in Neovim
-- :electric_plug: Support for Anthropic, Copilot, Gemini, Ollama, OpenAI and xAI LLMs (or bring your own!)
+- :electric_plug: Support for Anthropic, Copilot, Gemini, Ollama, OpenAI, Azure OpenAI and xAI LLMs (or bring your own!)
 - :rocket: Inline transformations, code creation and refactoring
 - :robot: Variables, Slash Commands, Agents/Tools and Workflows to improve LLM output
 - :sparkles: Built in prompt library for common tasks like advice on LSP errors and code explanations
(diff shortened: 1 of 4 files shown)


strayer:
I realized I forgot to update the codecompanion.txt doc and the README. I'll do that soon!

olimorris:
Thanks so much this! Pretty sure this will be of huge use to many people.

I have no way of testing it but it all looks good from my side.

strayer:
Thanks for checking the PR! I updated the README.md and regenerated the doc/codecompanion.txt, but I still need to update the model parameter as its current copied version is kind of confusing. I'll mark the PR ready for review as soon as I've been able to find time for that!

strayer:
Updated the model schema to type string and non-optional. Not really sure if it makes sense like this, please make sure to review that! Apart from that I assume this PR is ready for review - I've been using the plugin like this for a few days now (albeit not too often) and everything seems to work fine.

olimorris:
Thanks again for this.

olimorris:
@strayer heads up. I may have broken your personal azure_openai config in `v10.0.0`.

#458 was raised today and as I fixed that, I noticed that the recommended use of the Azure adapter differs slightly from the rest of the adapters. Namely that the model should be set with `schema.model.default`.

strayer:
Thanks for letting me know! I actually noticed this and mentioned it in the PR description:

> Please check if I used the schema value replacement for the env var correctly. I was a bit confused because the docs always mentioned schema.model.default when talking about this functionality.

The code confused me a bit, sorry that it caused an issue for you!

olimorris:
Not at all. The joys of working with different LLMs and APIs 😄

PR #390, opened 2024-11-03
Add doctoring prompt to library

tim-hilde:
## Description

This PR adds a prompt to generate docstrings to the prompt_library. It uses treesitter to get the content of the function the cursor is on. It prompts the user to input a specific style or uses the languages typical one.

## Related Issue(s)

https://github.com/olimorris/codecompanion.nvim/discussions/364

## Screenshots

image

image

## Checklist

### DIFF
--- a/lua/codecompanion/config.lua
+++ b/lua/codecompanion/config.lua
@@ -762,7 +762,7 @@ This is the code, for context:
         is_default = true,
         is_slash_cmd = false,
         modes = { "n" },
-        short_name = "Docstrings",
+        short_name = "docstrings",
         auto_submit = true,
         stop_context_insertion = true,
         user_prompt = true,
@@ -785,7 +785,7 @@ This is the code, for context:
               .. "\n"
               .. function_content
               .. "\n```\n"
-              .. "Generate only the docstrings and nothing else using tabs or spaces depending on the current implementation. Do not change anything else in the code. Give only the docstrings. Use the following style to generate the docstrings. If I don't give you a style, use the typical style of the language."
+              .. "Generate only the docstrings and nothing else using tabs or spaces depending on the current implementation. Do not change anything else in the code. Give only the docstrings. Use the following style to generate the docstrings. If I don't give you a function, do nothing. If I don't give you a style, use the typical style of the language."
           end,
         },
       },
--- a/lua/codecompanion/config.lua
+++ b/lua/codecompanion/config.lua
@@ -754,6 +754,42 @@ This is the code, for context:
         },
       },
     },
+    ["Generate docstrings"] = {
+      strategy = "inline",
+      description = "Generate docstrings for this function",
+      opts = {
+        index = 11,
+        is_default = true,
+        is_slash_cmd = false,
+        modes = { "n" },
+        short_name = "Docstrings",
+        auto_submit = true,
+        stop_context_insertion = true,
+        user_prompt = true,
+        placement = "add",
+      },
+      prompts = {
+        {
+          role = constants.SYSTEM_ROLE,
+          content = function(context)
+            local get_function_content = require("codecompanion.utils.treesitter").get_function_at_cursor
+            local bufnr = context.bufnr
+            local cursor = context.cursor_pos
+            local function_content = get_function_content(bufnr, cursor)
+
+            return "I want you to act as a senior "
+              .. context.filetype
+              .. " developer. Here is a function I want you to generate the docstrings for: "
+              .. "\n```"
+              .. context.filetype
+              .. "\n"
+              .. function_content
+              .. "\n```\n"
+              .. "Generate only the docstrings and nothing else using tabs or spaces depending on the current implementation. Do not change anything else in the code. Give only the docstrings. Use the following style to generate the docstrings. If I don't give you a style, use the typical style of the language."
+          end,
+        },
+      },
+    },
   },
   -- DISPLAY OPTIONS ----------------------------------------------------------
   display = {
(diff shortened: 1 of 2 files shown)


olimorris:
This looks really great so thanks for taking the time to PR this.

https://github.com/user-attachments/assets/1eb4916b-860f-43fc-a839-c0ea5cddcf9a

A few thoughts:

- What do you type into the user_prompt? Setting it to false is a much cleaner experience for me.
- I notice that for me it places the docstring after the cursor...I guess this is on me to add a `before` placement? (Someone else has asked for this and it's a quick addition)
- `short_name` should be lowercase, it's what you type in the cmdline to reach to prompt.
- If you're not in a function, you get this error: `Error executing vim.schedule lua callback: ...e/Neovim/codecompanion.nvim/lua/codecompanion/config.lua:787: attempt to concatenate local 'function_content' (a nil value)`
- What do you think about adding a visual selection option?

tim-hilde:
Thanks for the quick feedback!


> * What do you type into the user_prompt? Setting it to false is a much cleaner experience for me.
> * I notice that for me it places the docstring after the cursor...I guess this is on me to add a `before` placement? (Someone else has asked for this and it's a quick addition)
> * `short_name` should be lowercase, it's what you type in the cmdline to reach to prompt.
> * If you're not in a function, you get this error: `Error executing vim.schedule lua callback: ...e/Neovim/codecompanion.nvim/lua/codecompanion/config.lua:787: attempt to concatenate local 'function_content' (a nil value)`
> * What do you think about adding a visual selection option?

As for your thoughts:
1. as there are a lot of different styles for docstrings, I thought about using the user prompt to specify the style. e.g. in Python there is numpy or google style etc. If no input is given, the language's typical style is used
2. Positioning is kind of tricky I guess, as it seems to depend on the language. Do you have any suggestions?
3. This is due to how the prompt works. I've implemented a function that gets the text of a function via treesitter. What do you think should happen, when the cursor is not in a function? Just nothing?
4. Does it make sense to implement a visual selection? In my experience you want the docstrings when you're at the top of the function and not hassle with selecting anything

PLAZMAMA:
What a coincidence! I was going to try creating my doc string generator for this plugin because it wasn't a thing yet.
@tim-hilde I'm looking forward to using this once it gets merged. Thanks!!!

olimorris:
@tim-hilde I've added a "before" placement that should hopefully solve the issue of the doc string output appearing after the cursor. Let me know if it works...it's from many months ago so may need a little tweaking.

marshallformula:
I just found this PR when trying to implement my own docstring inline function. One issue I'm running into is that [the inline strategy adds the CODE_ONLY_PROMPT](https://github.com/olimorris/codecompanion.nvim/blob/331d95cd86627e2ed998a2ca012d6fed01080486/lua/codecompanion/strategies/inline.lua#L283) which instructs the model to output nothing if it can't respond with code only. 

When running specifically with the ollama adapter and `llama3.1` or `llama3.2` models it seems to follow those instructions and it outputs nothing.  Other adapters seem to ignore it and still give me output. Even other ollama models give me something - but llama3 gets nothing.  There are no errors from the ollama api end.

When I switch the strategy to `chat` it will give me output - which makes me believe that the `CODE_ONLY_PROMPT` is the issue.  

I'm not sure if this needs to be addressed as a separate issue, but it does seems specifically problematic with trying to write a docstring - which is not code - with an inline strategy. Should there be an option to disable that `CODE_ONLY_PROMPT` for inline strategies?

tim-hilde:
After a lot back and forth I'm sorry to inform you, that I cannot manage to get the function working as I would expect it to:
- be language dependent in the positioning of the doctoring
- give the user the option to specify a style

For now I will close this PR and use my own prompt as I mainly use python. For those interested:

```lua
prompt_library = {
	["Docstring"] = {
		strategy = "inline",
		description = "Generate docstring for this function",
		opts = {
			modes = { "v" },
			short_name = "docstring",
			auto_submit = true,
			stop_context_insertion = true,
			user_prompt = false,
		},
		prompts = {
			{
				role = "system",
				content = function(context)
					return "I want you to act as a senior "
						.. context.filetype
						.. " developer. I will send you a function and I want you to generate the docstrings for the function using the numpy format. Generate only the docstrings and nothing more. Put the generated docstring at the correct position in the code. Use tabs instead of spaces"
				end,
			},
			{
				role = "user",
				content = function(context)
					local text = require("codecompanion.helpers.actions").get_code(context.start_line, context.end_line)

					return text
				end,
				opts = {
					placement = "add",
					contains_code = true,
				},
			},
		},
	},
},

```

olimorris:
Thanks so much @tim-hilde. Really appreciate your efforts. I'll add this to the roadmap for sometime soon.

rbhanot4739:
> prompt_library = {
> 	["Docstring"] = {
> 		strategy = "inline",
> 		description = "Generate docstring for this function",
> 		opts = {
> 			modes = { "v" },
> 			short_name = "docstring",
> 			auto_submit = true,
> 			stop_context_insertion = true,
> 			user_prompt = false,
> 		},
> 		prompts = {
> 			{
> 				role = "system",
> 				content = function(context)
> 					return "I want you to act as a senior "
> 						.. context.filetype
> 						.. " developer. I will send you a function and I want you to generate the docstrings for the function using the numpy format. Generate only the docstrings and nothing more. Put the generated docstring at the correct position in the code. Use tabs instead of spaces"
> 				end,
> 			},
> 			{
> 				role = "user",
> 				content = function(context)
> 					local text = require("codecompanion.helpers.actions").get_code(context.start_line, context.end_line)
> 
> 					return text
> 				end,
> 				opts = {
> 					placement = "add",
> 					contains_code = true,
> 				},
> 			},
> 		},
> 	},
> },
> 

@tim-hilde Did you get the positioning of the docstring correct? I was doing the similar thing, but the docstring is generated above the function for even Python files.

PR #463, opened 2024-11-20
fix(diff): last_chat referenced is_visible() directly instead of last_chat.ui

GitMurf:
## Description

Looks like just a couple items got missed in the refactor to using Chat:ui

## Checklist


GitMurf:
As an FYI, this was the error I was getting:

image

PR #482, opened 2024-11-25
feat(prompt): change adapter before custom prompt execute

nangua:
1. change adapter for slash cmd style custom prompt, if specified
2. fix custom prompt register for completion. (After nightly lazy update earlier this moning, i found my custom slash cmd lost, and with he powerful help of AI(this plugin), i found the vim.iter problem and try to fixed it)

hope it is helpful.

## Checklist

Not involved.

### DIFF
--- a/lua/codecompanion/completion.lua
+++ b/lua/codecompanion/completion.lua
@@ -1,6 +1,7 @@
 local SlashCommands = require("codecompanion.strategies.chat.slash_commands")
 local config = require("codecompanion.config")
 local strategy = require("codecompanion.strategies")
+local util = require("codecompanion.utils")
 
 local trigger = {
   agents = "@",
@@ -28,12 +29,8 @@ function M.slash_commands()
     end)
     :totable()
 
-  vim
-    .iter(config.prompt_library)
-    :filter(function(_, v)
-      return v.opts and v.opts.is_slash_cmd and v.strategy == "chat"
-    end)
-    :map(function(_, v)
+  for _, v in pairs(config.prompt_library) do
+    if v.opts and v.opts.is_slash_cmd and v.strategy == "chat" then
       table.insert(slash_commands, {
         label = "/" .. v.opts.short_name,
         detail = v.description,
@@ -41,7 +38,8 @@ function M.slash_commands()
         type = "slash_command",
         from_prompt_library = true,
       })
-    end)
+    end
+  end
 
   return slash_commands
 end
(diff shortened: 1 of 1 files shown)


olimorris:
Thanks for this.

I don't want any adapter change logic to be implemented via slash commands or the plugin's completion logic. I'm happy to have some custom events which fire, allowing you to hook into them and change the adapter. But your current implementation is making the single completion class responsible for too much.

Can you share your custom slash command btw? I'm curious why you'd have a slash command that would change the adapter in the chat buffer when you could open up a chat buffer with that adapter?

nangua:
> Thanks for this.
> 
> I don't want any adapter change logic to be implemented via slash commands or the plugin's completion logic. I'm happy to have some custom events which fire, allowing you to hook into them and change the adapter. But your current implementation is making the single completion class responsible for too much.
> 
> Can you share your custom slash command btw? I'm curious why you'd have a slash command that would change the adapter in the chat buffer when you could open up a chat buffer with that adapter?

Or is it My personal preference?  I thought it acts the same as mini_picker.

According to the guide [Other Configuration Options](https://github.com/olimorris/codecompanion.nvim/blob/main/doc/RECIPES.md#other-configuration-options), it states "Specifying an Adapter and Model", I thought the adapter configuration item would automatically switch the adapter.

If this is not your original design, please ignore this PR.

BTW, in my ofen usage, i need to chat with seriese AI `Wrapper` like Dify, and i had some special prompt for specIfed adapter, like:

```lua
{
	adapters = {
		my_adapter = function()
			return require("codecompanion.adapters").extend(
				require("my.plugins.for.coding.codecompanion.adapter.dify"),
				{
					name = "my_adapter",
					env = {
						api_key = "APP_API_KEY_FOR_A",
					},
					parameters = {
						inputs = {
							foo = "A",
						},
					},
				}
			)
		end,
	},
	prompt_library = {
		["Prompt for Purpose A"] = {
			strategy = "chat",
			description = "Prompt for Purpose A",
			opts = {
				short_name = "chat_for_purpose_a",
				adapter = {
					name = "my_adapter",
					model = "Qwen2.5-14B",
				},
				is_default = true,
				is_slash_cmd = true,
				auto_submit = false,
				user_prompt = true,
				stop_context_insertion = true,
				contains_code = false,
			},
			prompts = {
				{
					role = "user",
					content = " Please do it for me:\n",
				},
			},
		},
	},
}
```

With the above configuration, I can switch to Dify Wrapper `my_adapter` through slash command `/chat_for_purpose_2`, and chat with Dify with specified `parameters`: inputs.foo=A.

I don't know if this is correct.

olimorris:
Thanks for spotting the issue of prompt library slash commands not appearing in the chat buffer. That is now fixed.

I take your point about your slash command having a different adapter which isn't being applied in the chat buffer. I'll have a think about what to do with this.

nangua:
Appreciate, and I will keep My fork syncing up-to-date.

PR #485, opened 2024-11-26
feat(keymaps): add chat buffer window hide keymap

GitMurf:
## Description

I really like to use the `CodeCompanionChat Toggle` command to show and hide a chat window. The problem with the keymaps for the chat window is that it only has a "close" mapping which completely "destroys" the chat. I often find myself wanting to quickly hide the chat window just like if I were to do "Toggle" cmd while a chat window is open.

## Checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -581,6 +581,7 @@ When in the chat buffer, press `?` to bring up a menu that lists the available k
 - `gc` to insert a codeblock in the chat buffer
 - `gd` to view/debug the chat buffer's contents
 - `gf` to fold any codeblocks in the chat buffer
+- `gh` to hide the chat buffer window
 - `gr` to regenerate the last response
 - `gs` to toggle the system prompt on/off
 - `gx` to clear the chat buffer's contents
--- a/lua/codecompanion/config.lua
+++ b/lua/codecompanion/config.lua
@@ -259,6 +259,14 @@ local defaults = {
           callback = "keymaps.toggle_system_prompt",
           description = "Toggle the system prompt",
         },
+        hide = {
+          modes = {
+            n = "gh",
+          },
+          index = 16,
+          callback = "keymaps.hide",
+          description = "Hide Chat",
+        },
       },
       opts = {
         register = "+", -- The register to use for yanking code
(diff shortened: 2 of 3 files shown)


GitMurf:
> Could you not map `CodeCompanionToggle` to a keybinding? I use `<localleader>a` in my dot files so I can then restore it from any other buffer

Yes totally and that is what I do. But I want to remap the builtin hide to just "q" to hide the window (this is a common keymap for me) as it is quick and muscle memory instead of having to do "<localleader>C" to close when in chat window.

Not so much for me, as I can also add my own "q" keymap with a filetype autocmd or something... but for the "average user" this seems like a keymap that would be available when you have things like Close (destroys chat), Stop/kill etc. There is no builtin keymap that allows you to "close" the chat window without destroying it. And although I have multiple options to do this myself, I am on a neovim config kick right now on minimizing as much as possible my customizations to plugins so I don't have to keep tweaking / fixing things when plugins change or add new features ;)

No biggie if you don't want this added. Just had been thinking about it for a while and it was a quick / easy PR to propose.

oVerde:
Kind of related https://github.com/olimorris/codecompanion.nvim/discussions/486

olimorris:
Yep totally get where you're coming from. Given this is possible with `CodeCompanionToggle` and is my preferred way of working with the chat buffer, I'll reject this PR but have just added a feature to enable you to add custom keymaps:

```lua
require('codecompanion').setup({
  strategies = {
    chat = {
      keymaps = {
        hide = {
          modes = {
            n = "gh",
          },
          callback = function(chat)
            chat.ui:hide()
          end,
          description = "Hide the chat buffer",
        },
      }
    }
  }
})
```

There's likely soooo much more that can be done by giving users direct access to the chat buffer like this so feels like a sensible move and still allows you to have your desired functionality.

GitMurf:
> but have just added a feature to enable you to add custom keymaps:

Funny! I was waiting to see if you decided to reject the PR and then I was going to propose custom keymaps. Honestly that is my first preference anyways ;) nice job!

PR #527, opened 2024-12-11
feat(adapters): Adding Huggingface adapter

bassamsdata:
## Description

New adapter for the HuggingFace serverless inference API. It is somewhat OpenAI API-compatible, but not fully.

## Related Issue(s)

related: https://github.com/olimorris/codecompanion.nvim/discussions/517#discussioncomment-11517785

Thank you

## Checklist


lua/codecompanion/adapters/huggingface.lua
```diff
@@ -0,0 +1,212 @@
...
+      end
+      -- Set model in env_replaced for URL construction
+      self.env_replaced = self.env_replaced or {}
+      self.env_replaced.model = self.parameters.model or self.schema.model.default
+
+      -- Add headers with string values
+      if self.parameters.use_cache then
+        self.headers["x-use-cache"] = self.parameters.use_cache -- Already a string
+      end
+      if self.parameters.wait_for_model then
```
olimorris, line 49: Does Huggingface require us to set a parameter and a header for this?

lua/codecompanion/adapters/huggingface.lua
```diff
@@ -0,0 +1,212 @@
...
+  handlers = {
+    ---@param self CodeCompanion.Adapter
+    ---@return boolean
+    setup = function(self)
+      if self.opts and self.opts.stream then
+        self.parameters.stream = true
+      end
+      -- Set model in env_replaced for URL construction
+      self.env_replaced = self.env_replaced or {}
+      self.env_replaced.model = self.parameters.model or self.schema.model.default
```
olimorris, line 43: We shouldn't need to do work with the `env_replaced` table

olimorris:
Thanks so much for this. I've added a couple of comments.

Also, can we reuse the OpenAI adapter for the handlers? Like what I've done with `copilot` and `openai_compatible` adapters? It _may_ prevent us requiring an additional test and will make the adapter logic that much easier.

lua/codecompanion/adapters/huggingface.lua
bassamsdata: Actually, those are in the header (which they set it by default), and I've already set the same default values. so no need for those.

lua/codecompanion/adapters/huggingface.lua
bassamsdata: done, I removed this

bassamsdata:
> can we reuse the OpenAI adapter for the handlers? 

Sure, I actually started with that approach following the xAI adapter, but for some reason, it didn’t work. I might have had something else interfering with the code at the time. Now, everything runs using the OpenAI adapter.

bassamsdata:
One more thing: I resolved the conversations for the two reviews. I’m not sure if that’s the correct approach or if it’s something you’re supposed do the `resolve conversation`. Apologies if I overstepped, I don’t have much experience with PRs.

olimorris:
Looks fantastic, thank you. Feel free to drop the test as we have it all covered with the OpenAI one.

bassamsdata:
done, tests dropped.

olimorris:
Thanks so much for this. Feels great to have this in CodeCompanion finally.

PR #539, opened 2024-12-15
Fix documentation for system_prompt callback

gleachkr:
The argument to the system_prompt callback is an adapter, not an opt
table containing an adapter.

## Description

I tweaked the docs slightly to correct something I think might be off (it was a
minor papercut issue for me while tinkering with the system_prompt). Because
it's such a small change, I've skipped the discussion steps from the
CONTRIBUTING doc, I hope that's OK.

## Checklist

- [X] I've read the [CONTRIBUTING](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md) guidelines and have adhered to them in this PR
- [X] I've updated the README and ran the `make docs` command

### DIFF
--- a/README.md
+++ b/README.md
@@ -322,8 +322,8 @@ require("codecompanion").setup({
   opts = {
     ---@param adapter CodeCompanion.Adapter
     ---@return string
-    system_prompt = function(opts)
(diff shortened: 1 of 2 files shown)


olimorris:
Great spot and thanks for taking the time to fix this for others.

PR #553, opened 2024-12-20
docs: fix blink.cmp example

fredrikaverpil:
## Why this change?

When adding `blink.cmp` completion with codecompanion to [my config](https://github.com/fredrikaverpil/dotfiles/blob/baf5af27f6c50a49bc87d411d2e557ee8783eb9b/nvim-fredrik/lua/fredrik/plugins/codecompanion.lua#L105-L121), I noticed the docs showed an outdated setup example.

### What was changed?

I replaced the lua code snippet with a complete example.


olimorris:
Awesome thanks for this. Hard to keep up with Blink's changes at the minute. It's on fire.

fredrikaverpil:
@olimorris I know, it's moving really fast, breaking often. I've been waiting for cmdline support, and since that's in now - I'll just refer from updating in a while now, I think. 😅

PR #556, opened 2024-12-21
feat(chat): window position option

JonathanREmery:
## Description

I've added the option to choose the Chat Window position (left, right, top, bottom). This can be set in the config, and only works in either the vertical, or horizontal layout.

## Related Discussion

I've opened a discussion regarding this feature: https://github.com/olimorris/codecompanion.nvim/discussions/555

## Checklist

- [✓] I've read the [CONTRIBUTING](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md) guidelines and have adhered to them in this PR


olimorris:
Thanks so much for the PR. I tried this out and couldn't get it to work.

I'm in two minds about including this as I expect most users would want it to respect `vim.opt.splitright`.

JonathanREmery:
Ah, `vim.opt.splitright` and `vim.opt.splitbelow` were what I was missing. Thanks for letting me know! I've went ahead and added that into the positioning logic. The behavior should be as follows:

If a user does not specify a `position` in their `display.chat.window` config, the chat window will default depending on their `splitright` and `splitbelow`.

If a user specifies a `position`, but it is invalid for their `layout` (ex. `layout`: "vertical", `position`: "top"), then the chat window will default depending on their `splitright` and `splitbelow`.

If a user specifies a `position` that is valid for their `layout`, the chat window will use that position, ignoring `splitright` and `splitbelow`.

This would allow the user to set their `splitright` and `splitbelow` for splits, but override it with a custom position just for the CodeCompanion Chat if they desire. This allows that extra bit of customization without affecting the rest of their plugin's splits.

olimorris:
Working perfectly! Really, really appreciate this PR. A fantastic QOL feature to have.

PR #563, opened 2024-12-22
chore: update blink config in minimal.lua

S1M0N38:
## Description

`minimal.lua` make use of the old `sources.completion.enabled_providers` setting. Newer version of blink now use `sources.default`. This is the same as provided in the codecompanion.nvim README example.

## Checklist

- [ ] I've updated the README and ran the `make docs` command

### DIFF
--- a/minimal.lua
+++ b/minimal.lua
@@ -30,9 +30,7 @@ local plugins = {
             ["<Tab>"] = { "select_next", "fallback" },
           },
           sources = {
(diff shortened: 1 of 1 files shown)


olimorris:
Awesome thanks. I'll update the bug report later as well.

I've uninstalled blink from my personal config as the breaking changes are too frequent for me to keep track of.

PR #567, opened 2024-12-23
feat(chat): adding to chat initializes buffer if needed

JonathanREmery:
## Description

I often found that the first time I would use CodeCompanion would be to add some code to the chat and ask a question. I kept running into the issue where I had not initialized the chat window yet so I would get a "No chat buffer found" warning.

I figured it would be nice if it automatically created a new chat, if one does not already exist when the user calls `:CodeCompanionChat Add`

## Checklist

- [✓ ] I've read the [CONTRIBUTING](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md) guidelines and have adhered to them in this PR

### DIFF
--- a/lua/codecompanion/init.lua
+++ b/lua/codecompanion/init.lua
@@ -89,17 +89,21 @@ end
 ---@param args table
 ---@return nil
 M.add = function(args)
+  local context = context_utils.get(api.nvim_get_current_buf(), args)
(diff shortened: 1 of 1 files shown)


olimorris:
And thanks for this. Such an awesome QOL improvement.

adam-e-trepanier:
Thank you for this!

Issue #536, opened 2024-12-14
[Bug]: Gemini inline inserts code  tags


hoagy-davis-digges:
### Your `minimal.lua` config

```lua
---@diagnostic disable: missing-fields

--NOTE: Set config path to enable the copilot adapter to work.
--It will search the follwoing paths for the for copilot token:
--  - "$CODECOMPANION_TOKEN_PATH/github-copilot/hosts.json"
--  - "$CODECOMPANION_TOKEN_PATH/github-copilot/apps.json"
vim.env["CODECOMPANION_TOKEN_PATH"] = vim.fn.expand("~/.config")

vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

-- Your CodeCompanion setup
local plugins = {
  {
    "olimorris/codecompanion.nvim",
    dependencies = {
      { "nvim-treesitter/nvim-treesitter", build = ":TSUpdate" },
      { "nvim-lua/plenary.nvim" },
      -- Test with blink.cmp
      {
        "saghen/blink.cmp",
        lazy = false,
        version = "*",
        opts = {
          keymap = {
            preset = "enter",
            ["<S-Tab>"] = { "select_prev", "fallback" },
            ["<Tab>"] = { "select_next", "fallback" },
          },
          sources = {
            completion = {
              enabled_providers = { "lsp", "path", "buffer", "codecompanion" },
            },
            providers = {
              codecompanion = {
                name = "CodeCompanion",
                module = "codecompanion.providers.completion.blink",
                enabled = true,
              },
            },
          },
        },
      },
      -- Test with nvim-cmp
      -- { "hrsh7th/nvim-cmp" },
    },
    opts = {
      --Refer to: https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/config.lua
      strategies = {
        --NOTE: Change the adapter as required
        chat = { adapter = "gemini" },
        inline = { adapter = "gemini" },
      },
        adapters = {
            gemini = function()
              return require("codecompanion.adapters").extend("gemini", {
                env = {
                  api_key = "****",
                  -- model = "gemini-2.0-flash-exp"
                  model = "gemini-1.5-flash"
                },
              })
            end,
          },
      opts = {
        log_level = "DEBUG",
      },
    },
  },
}

require("lazy.minit").repro({ spec = plugins })

[...]

PR #568, opened 2024-12-23
fix(inline): fixes gemini-2.0-flash-exp code blocking

JonathanREmery:
## Description

I've been having issues with `gemini-2.0-flash-exp` returning code blocks inline, despite the prompt not to. As well as bad classifications... This is an issue with `gemini-2.0-flash-exp`, not CodeCompanion. However I saw other people were having the same issue so thought I'd put out what's worked for me.

This will manually strip out the code blocking if the model is set to `gemini-2.0-flash-exp`. I have also added one more example to the `PLACEMENT_PROMPT` that seems to help with classification a lot.

## Related Issue(s)

https://github.com/olimorris/codecompanion.nvim/issues/536

## Checklist

- [✓ ] I've read the [CONTRIBUTING](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md) guidelines and have adhered to them in this PR

### DIFF
--- a/lua/codecompanion/adapters/gemini.lua
+++ b/lua/codecompanion/adapters/gemini.lua
@@ -136,7 +136,17 @@ return {
           return
         end
 
-        return json.candidates[1].content.parts[1].text
+        local text = json.candidates[1].content.parts[1].text
+        local model = json.modelVersion
+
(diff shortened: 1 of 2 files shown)


olimorris:
Thanks so much for adding this.

PR #581, opened 2025-01-02
feat(copilot): add token count handler using openai adapter

GitMurf:
## Description

The copilot adapter has a similar response type as openai to allow for using the openai token counter handler.

## Checklist

- [x] I've updated the README

### DIFF
--- a/lua/codecompanion/adapters/copilot.lua
+++ b/lua/codecompanion/adapters/copilot.lua
@@ -110,7 +110,7 @@ return {
   },
   features = {
     text = true,
-    tokens = false,
+    tokens = true,
     vision = false,
   },
   url = "https://api.githubcopilot.com/chat/completions",
@@ -167,6 +167,9 @@ return {
     end,
     form_messages = function(self, messages)
       return openai.handlers.form_messages(self, messages)
+    end,
+    tokens = function(self, data)
+      return openai.handlers.tokens(self, data)
     end,
     chat_output = function(self, data)
       return openai.handlers.chat_output(self, data)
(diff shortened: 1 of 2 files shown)


olimorris:
Fantastic thanks. Have you tested this with Claude as well?

GitMurf:
> Fantastic thanks. Have you tested this with Claude as well?

If you mean using Claude 3.5 as the model (instead of 4o, o1 etc) via copilot, yes.

Claude is my default model I use, so actually now that I think of it, I should make sure and test the other copilot OpenAI models like 4o etc. I assume it is the same but will test tomorrow to make sure.

GitMurf:
Ok @olimorris I have tested the following models for copilot and they all successfully show token counts. I did have to add a few changes to the PR (please review... see additional commit above).

## Some notes

### Tested against following copilot models

- claude-3.5-sonnet
- o1-mini-2024-09-12
- o1-2024-12-17
- gpt-4o-2024-08-06
- gpt-4o-mini-2024-07-18

### It seems `o1-preview` no longer works and is now replaced by `o1`

I got an api error when trying to use `o1-preview-2024-09-12`. I looked on openai website and noticed in mid-december they released the `o1` model (`o1-2024-12-17`) and tested it (as seen in notes above) and it worked successfully with this PR.

### Added commit to this PR with a helper function

The update to the `o1` model seemed to come with a slight change in the response structure which is what I addressed in the most recent commit added to this PR. Instead of doing the `data:sub(7)` "hardcoded" parse of the response string before JSON parsing, I created a simple helper function that now looks for the first `{` curly brace and uses that in place of `7`.

Let me know if you have any questions.

olimorris:
Fantastic!! Thanks so much. I'll look into `o1` later today.

PR #582, opened 2025-01-02
Fix completions keymap to allow for multiple completion keywords in a single line

GitMurf:
@olimorris Note: this PR is on top of my other PR #581 so includes those changes (minimal). So if you reject that one or change it, just need to update in this PR accordingly.

## Description

Currently if you try to do two completions in the same line, it deletes everything after the first completion and just provides completions items for that first keyword. See the following as example:

(cursor at end of line below, after `#ls`)
```
I am writing a message and want to reference `#buffer` and include `#ls`
```

When I try to do a completion after `#ls` (for `#lsp`), it jumps back to the first found completion keyword in the line (`#buffer`) and removes everything after it... the following is removed: ` and include #ls` 

It is because the current code is just checking across the entire line for the first match of a trigger character. This PR will instead just look at the "current word" where the cursor is at and ignore the rest of the line (before and after current cursor/word).

## Checklist

- [x] I've updated the README

### DIFF
--- a/lua/codecompanion/strategies/chat/keymaps.lua
+++ b/lua/codecompanion/strategies/chat/keymaps.lua
@@ -206,7 +206,10 @@ M.completion = {
         return
       end
 
-      local prefix, start = unpack(vim.fn.matchstrpos(line:sub(1, col), [[\%(@\|/\|#\|\$\)\S*]]))
+      local before_cursor = line:sub(1, col)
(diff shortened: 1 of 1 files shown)


olimorris:
You happy with the PR? I can review later

GitMurf:
> You happy with the PR? I can review later

Yes I have been using it and so far no issues and doing exactly what I hoped.

olimorris:
Thanks for much for contributing this. Really appreciated.

PR #596, opened 2025-01-05
feat(completion): Boost score of blink completion suggestions

schilkp:
## Description

By default, blink's `path` provider has a score offset of 3, meaning that completions for slash commands in the buffer are deep down below possible path completions in the list of suggestions.

Since this suggestion only triggers in the chat buffer, it is reasonable to assume that slashcommands are much more likely to be the correct suggestions than folders at `/`.

## Screenshots

Before:
image

After:
image

## Checklist

- [X] I've read the [CONTRIBUTING](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md) guidelines and have adhered to them in this PR
- [X] I've updated the README
- [X] I've ran the `make docs` command

### DIFF
--- a/lua/codecompanion/init.lua
+++ b/lua/codecompanion/init.lua
@@ -338,6 +338,7 @@ M.setup = function(opts)
             name = "CodeCompanion",
             module = "codecompanion.providers.completion.blink",
             enabled = true,
+            score_offset = 10,
           })
         end)
       end,


olimorris:
Amaaaaazing! Such a good idea. Anything obvious I miss with blink, please do PR it.

PR #615, opened 2025-01-08
handle library promts in `:CodeCompanion<Cr>`

### LINKED ISSUE #614: [Bug]: `:CodeCompanion<Cr>` does not parse prompts from the prompt library
### Your `minimal.lua` config

n/a

### Error messages

_No response_

### Log output

n/a

### Health check output

_No response_

### Describe the bug

When running `:CodeCompanion /buffer write module-level documentation<Cr>` the response is as expected. When doing the same but using `:CodeCompanion<Cr>` and then typing `/buffer write module-level documentation` in the input prompt, a generic response is returned that is unrelated to the buffer.

The logs indicate that the first variant adds the buffer contents to the prompt as expected:

```
Prompt to classify: { {
    content = "Here is the content of a buffer, for context:\n\n ....."
```

But the second variant does not:

```
Prompt to classify: { {
    content = "<question>/buffer add module-level documentation</question>",
    role = "user"
  } }
```

### Reproduce the bug

_No response_

### Adapter and model

claude-3.5-sonnet

### Final checks

JeanMertz:
`:CodeCompanion<Cr>` now behaves the same as  `:CodeCompanion /buffer<Cr>` if you add `/buffer` as the text in the input field.

Fixes #614

### DIFF
--- a/lua/codecompanion/commands.lua
+++ b/lua/codecompanion/commands.lua
@@ -40,15 +40,7 @@ return {
   {
     cmd = "CodeCompanion",
     callback = function(opts)
-      if #vim.trim(opts.args or "") == 0 then
-        vim.ui.input({ prompt = config.display.action_palette.prompt }, function(input)
-          if #vim.trim(input or "") == 0 then
-            return
-          end
-          opts.args = input
-          codecompanion.inline(opts)
-        end)
-      else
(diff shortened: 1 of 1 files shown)


olimorris:
Thanks for the PR!

PR #623, opened 2025-01-09
fix(copilot): adapt to json usage token response change

GitMurf:
Seems that the Copilot Chat (and potentially openai) response for `usage` token information changed slightly. Adding more robust check as `total_tokens` can be 0 evidently, so if so, will instead add completions and prompt tokens.


olimorris:
Thanks for spotting this. I've moved this logic to the Copilot adapter as the openai one is consumed by a lot of other models and it appeared to still be functioning.

PR #643, opened 2025-01-13
Raising error when both blink and cmp are available

edmondop:
It happened to me that I had blink cmp declared as a dependency and never used (and I know it is a bad practice), and I couldn't understand why the slash commands weren't working in the chat. This should give an heads up to people who could face the same problem

### DIFF
--- a/lua/codecompanion/init.lua
+++ b/lua/codecompanion/init.lua
@@ -331,7 +331,7 @@ M.setup = function(opts)
   local has_cmp, cmp = pcall(require, "cmp")
   local has_blink, blink = pcall(require, "blink.cmp")
   if has_blink and has_cmp then
-    return log:error("Both cmp and blink available, please uninstall one")
+    log:warn("Both cmp and blink available, please uninstall one")
   end
   if has_blink then
     pcall(function()
--- a/lua/codecompanion/init.lua
+++ b/lua/codecompanion/init.lua
@@ -330,6 +330,9 @@ M.setup = function(opts)
   -- Setup completion for blink.cmp and cmp
   local has_cmp, cmp = pcall(require, "cmp")
   local has_blink, blink = pcall(require, "blink.cmp")
+  if has_blink and has_cmp then
+    return log:error("Both cmp and blink available, please uninstall one")
+  end
   if has_blink then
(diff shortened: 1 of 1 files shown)


edmondop:
@olimorris please consider this a draft until I add tests. I wanted to open it so you can tell me if you agree with the ideas

olimorris:
Hey @edmondop and thanks for raising

I thought I'd covered this scenario. The plugin should look to load blink first, and if not, check to see if nvim-cmp is available and load that. This isn't happening?

> (and I know it is a bad practice)

No way! You can use Neovim how you like. I too had been dual running blink and nvim-cmp as the former was a little unstable for my liking.

edmondop:
My use case is slightly different: I use nvim-cmp and I tried to move to blink but couldn't get it work so I reverted.

In the meantime blink remained among my plugins because I forgot to remove it.

When I installed codecompanion the chat slash actions weren't working for me, so I opened the source code and I found the check , opened my plugins and found blink declared as a dep, I removed it and everything works

I submitted this change so that if someone else is in the same situation it gets an error rather than unexpected behavior

olimorris:
Got it. Thank you for sharing

lua/codecompanion/init.lua
```diff
@@ -330,6 +330,9 @@ M.setup = function(opts)
   -- Setup completion for blink.cmp and cmp
   local has_cmp, cmp = pcall(require, "cmp")
   local has_blink, blink = pcall(require, "blink.cmp")
+  if has_blink and has_cmp then
+    return log:error("Both cmp and blink available, please uninstall one")
```
olimorris, line 334: I think it makes sense to turn this into a `log:warn` and to remove the return. I don't want to cripple someones workflow because of this, instead just notify them.

edmondop: Thanks, is this visible in checkhealth ?

olimorris: It's not, principally because it's not a requirement for the plugin.

olimorris:
Thanks for adding this. Very much appreciated

olimorris:
Head's up. I had to revert this because of #651

PR #661, opened 2025-01-17
feat: register source for filetype in blink.cmp

Saghen:
## Description

I've added support for plugins to add `per_filetype` sources dynamically. It will fail if the user has set the `per_filetype` to a function, which should be extremely rare for this case. Open to feedback on the API! It's currently only available on `main`, and will be included in the next release (0.11.0).

I've also switched the recommended setup for <= 0.10.0 to use `sources.per_filetype` to avoid conflicts with other sources.

P.S. apologies for all the blink.cmp issues you've had to deal with. If I had it my way, less people would be using my plugin while it's still in beta :smile: Love the new docs site!

## Related Issue(s)

N/A

## Screenshots

N/A

## Checklist

- [x] I've updated the README

### DIFF
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -59,11 +59,13 @@ EOF
 
 ## Completion
 
-Out of the box, the plugin supports completion with both [nvim-cmp](https://github.com/hrsh7th/nvim-cmp) and [blink.cmp](https://github.com/Saghen/blink.cmp). For the latter, ensure that you've added `codecompanion` as a source:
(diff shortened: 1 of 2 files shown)


olimorris:
Well firstly let me say thanks so much for _your_ awesome docs that were the inspiration. Embarrassingly, I haven't acknowledged you in the README for that (...now corrected!).

> P.S. apologies for all the blink.cmp issues you've had to deal with. If I had it my way, less people would be using my plugin while it's still in beta 😄

Not at all. You make it perfectly clear up front that stuff will break whilst it's in the beta phase and I remember those days in CodeCompanion rather well...think it took me 8 months before I took my big warning notice down from the README 😆. I just didn't want the pressure that came with maintaining popular plugin, so huge respect to you.

Keep up the amazing work!

Issue #696, opened 2025-01-25
[Bug]: DeepSeek reasoning doesn't work with system prompt


niba:
### Your `minimal.lua` config

Not required

### Error messages

```txt
Error: {"error":{"message":"The first message (except the system message) of deepseek-reasoner must be a user message, but an assistant message detected.","type":"invalid_request_error","param":null,"code":"invalid_request_error"}}
```

### Log output

.

### Describe the bug

The problem is that 

> model expects interleaved user/assistant messages in the message sequence.

Here is more about this problem:
https://github.com/deepseek-ai/DeepSeek-R1/issues/21#issuecomment-2604700253

so if you have first message as role = "system" and next as "user" then deepseek-reasoning doesn't work

Maybe we could disable role="system" when we detect deepseek-reasoning?

### Reproduce the bug

1. Add deepseek-reasoning
2. Try to use chat with two user or system / user prompts

### Adapter and model

DeepSeek R1

### Final checks

- [x] I have made sure this issue exists in the latest version of the plugin
- [x] I have tested with the `minimal.lua` file from above and have shared this
- [x] I have shared the contents of the log file

olimorris:
Hi @niba. As I don't support DeepSeek in the plugin, there's nothing for me to change. You would need to modify the [custom adapter](https://codecompanion.olimorris.dev/extending/adapters.html) you've built, to handle this. The `handlers.form_messages` function would be the one to look at.

niba:
Hi @olimorris. Thanks for the tip. I used the built-in one that was merged a few days ago. I managed to find the problem in the deepseek adapter and created a PR with the fix.

h0pes:
@niba Sorry to tag you, I was wondering If your could share your custom deepseek-r1 adapter configuration. Thanks a lot

niba:
@h0pes hey. Sure! I use the built-in one with small modification to also get the "reasoning" output on the chat buffer. 

```lua
        deepseek = function()
          return require("codecompanion.adapters").extend("deepseek", {
            env = {
              api_key = "DEEPSEEK_API_KEY",
            },
            schema = {
              model = {
                default = "deepseek-reasoner",
              },
              temperature = {
                default = 0,
              },
            },
            handlers = {
              chat_output = function(self, data)
                local output = {}

                if data and data ~= "" then
                  local data_mod = prepare_data_for_json(data)
                  local ok, json = pcall(vim.json.decode, data_mod, { luanil = { object = true } })

[...]

PR #700, opened 2025-01-25
fix(deepseek): messages order in deepseek adapter

niba:
## Description

Deepseek R1 doesn't work when you don't have a `user` message after a `system` message. The current code transforms all messages (except first) with `system` roles to `assistant` roles. This logic breaks the chat when you have two or more system prompts at the beginning because Deepseek R1 requires a `user` message after a `system` message.

For example, if you try to send the following messages:

```
[
   { role: "system",  message: CodeCompanionMessage },
   { role: "system",  message: CustomMessage},
   { role: "user", message: question },
]
```

After transforming messages, you end up with:

```
[
   { role: "system",  message: CodeCompanionMessage },
   { role: "assistant",  message: CustomMessage},
   { role: "user", message: question },
]
```

This is not accepted by Deepseek, resulting in:

> Error: {"error":{"message":"The first message (except the system message) of deepseek-reasoner must be a user message, but an assistant message detected.","type":"invalid_request_error","param":null,"code":"invalid_request_error"}}

In this PR, I moved the code that changes all `system` roles (except the first one) to `assistant` to be called after message concatenation with the same role.

As a result, you can have multiple system messages at the beginning, and they will be concatenated. All other system messages will be transformed to assistant messages (if such situations exist).

Ping @bashtoni 

## Related Issue(s)

https://github.com/olimorris/codecompanion.nvim/issues/696

## Checklist

- [x] I've updated the README

### DIFF
--- a/lua/codecompanion/adapters/deepseek.lua
+++ b/lua/codecompanion/adapters/deepseek.lua
@@ -71,27 +71,6 @@ return {
         model = model()
       end
 
-      ---System role is only allowed as the first message, after this we must label
-      ---all system messages as assistant
-      local has_system = false
-      messages = vim
-          .iter(messages)
-          :map(function(m)
-            local role = m.role
-            if role == self.roles.llm then
-              if not has_system then
-                has_system = true
-              else
-                role = "assistant"
-              end
-            end
-            return {
-              role = role,
-              content = m.content,
-            }
-          end)
-          :totable()
-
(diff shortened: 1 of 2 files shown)


olimorris:
Apologies, I forgot we added this earlier in the week. Can you amend the deepseek test as well? Currently the test shows as passing but we should add the case where it's currently failing

niba:
Done! 

No worries. I'm glad that I can help

PR #716, opened 2025-01-26
feat(deepseek): show reasoning output

olimorris:
Reasoning is now displayed under H3 headers called "Reasoning". The response is displayed under a H3 header called "Response":

2025-01-26 21_33_30 - WezTerm@2x

2025-01-26 21_34_10 - WezTerm@2x


olimorris:
FYI...the reasoning is excluded from the messages table in the chat buffer

PR #750, opened 2025-01-28
Put initialization logic in `plugin/` directory

brianhuster:
## Description

- I moved initialization code to `plugin/` directory, as that is the built-in way to initialize a plugin (see `:h global-plugin`)
=> Users no longer have to call `require('codecompanion').setup()` just to use the plugin with default config

- Less dependent on `plenary.nvim`: Neovim 0.8+ has `vim.fs.joinpath()`, I think the plugin should use it instead of `plenary.path`

## Checklist


brianhuster:
Oops, I think I need to fix the style first

olimorris:
Sorry to be picky, but you haven't adhered to the CONTRIBUTING guidelines. This is quite a substantial change to the plugin that would have benefitted from a discussion before a PR was created. That being said, I agree with the sentiment of the PR.

How does this affect start up time for users with a package manager, if at all? Could you share any profiling? My limited knowledge of the vim bootup sequence, makes it look like these changes would automatically load the plugin which would be unacceptable to most. I note that there isn't a check to prevent any double loading.

I also want to point out that the guide you referenced hasn't been accepted yet and is still in discussion. So right now, I'm leaning towards rejecting this PR, but open to moving the plugin to the agreed best practice in the future.

Btw, good spot on the logging; I'm happy to accept that as a separate PR.

brianhuster:
> Sorry to be picky, but you haven't adhered to the CONTRIBUTING guidelines. 

I'm sorry for that, I guess I misunderstood the CONTRIBUTING guide. I thought I just needed to open a discussion before opening the PR

> How does this affect start up time for users with a package manager, if at all? Could you share any profiling? My limited knowledge of the vim bootup sequence, makes it look like these changes would automatically load the plugin which would be unacceptable to most.

This is my screenshot when using this PR
image

> My limited knowledge of the vim bootup sequence, makes it look like these changes would automatically load the plugin which would be unacceptable to most.

Technically it will only load the necessary files for initializing the plugin. That includes:
- plugin/codecompanion.lua
- lua/codecompanion/config.lua (required by plugin/codecompanion.lua)
- lua/codecompanion/commands.lua (required by plugin/codecompanion.lua)
- lua/codecompanion/utils/log.lua (required by plugin/codecompanion.lua)

In this PR, I have also move `require('plenary.async')` to a function so that it won't be loaded on startuptime

For people who use `lazy-loading` feature of `lazy.nvim`, they will still be just fine. `lazy.nvim` will not add the plugin to `rtp` until the desired event happens (for example `cmd = "CodeCompanion"`), so no file will be loaded until then.

> I note that there isn't a check to prevent any double loading.

My bad! I will add that. Possibly I'll use Vimscript variable `vim.g.loaded_codecompanion` for that, as it seems to be the standard way, as mentioned in the [doc](https://neovim.io/doc/user/usr_41.html#write-plugin), it is also used in some popular plugins like [telescope.nvim](https://github.com/nvim-telescope/telescope.nvim/blob/master/plugin/telescope.lua), [fzf-lua](https://github.com/ibhagwan/fzf-lua/blob/main/plugin/fzf-lua.lua)

> I also want to point out that the guide you referenced hasn't been accepted yet and is still in discussion.

But it has been assigned to milestone `0.11`, so I think it will be accepted soon. Reading all discussions there, it seems noone in Neovim team object to the point that initialization code should live in `plugin/` or `ftplugin/` (otherwise, why would they add support for using Lua in those folders in Nvim 0.5, right?), and that initialization and configuration code should be seperated. I think there are only 2 things remaining unresolved there:

- Whether to include guide about `subcommand`, which is not relevant to this PR
- Tests not passed in some old platforms

olimorris:
> I'm sorry for that, I guess I misunderstood the CONTRIBUTING guide. I thought I just needed to open a discussion before opening the PR

No worries. It's just I once had someone spend about 2 days on a PR which I couldn't accept because it would have broken the whole API for the user base (!!!) so I hate for people to spend so much time on something I might not accept.

Thanks for your clear explanation throughout, very much appreciated. This is looking really good. There might be a few things that I'd keep in the setup function such as the creation of the namespaces (although I should probably move them to the chat buffer at some point)...

I'll give this a try tomorrow and report back.

Thanks for your efforts on this.

olimorris:
Somehow I've ended up pushing some changes to your branch which need to be merged into this. Could you action that and then I think we're good to merge this.

brianhuster:
Do you mean I need to update the branch? Yes, I just did it and tested it with a few tasks (file system operations, command line running) so I think it should be safe to merge

olimorris:
Awesome! Thanks again for your hard work and diligence on this.

PR #756, opened 2025-01-29
fix(fetch): cached url folder needed to be created first on windows

GitMurf:
## Description

Trying to use the fetch slash command I was getting the error in the screenshot below.

## Screenshots

image

## Checklist

- [x] I've read the [contributing](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md) guidelines and have adhered to them in this PR
- [x] I've updated the README and/or relevant docs pages

PR #761, opened 2025-01-29
fix(workspaces): workspace path constant cleared after first use

GitMurf:
## Description

This is a weird one. I could not pinpoint the culprit, but if I started a chat convo, used /workspace slash command (works fine the first time), but then create a new convo and try /workspace slash again it gives a plenary error (see screenshot below).

The path arg going into the function is blank and somehow the CONSTANT for the path in that file is also suddenly blank.

You can see what I mean by the change in my commit (a couple lines). This solves my immediate problem but I have no clue if there is something more / deeper that has to be addressed 🤷‍♂️ 

## Related Issue(s)

No idea but curious if this is somehow related to #756 ?

## Screenshots

TabTip - Shell_Handwriting_Canvas_B4bKogwdUp_2025-01-28_22-14-58

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/slash_commands/workspace.lua
+++ b/lua/codecompanion/strategies/chat/slash_commands/workspace.lua
@@ -88,6 +88,11 @@ function SlashCommand:read_workspace_file(path)
   if not path then
     path = CONSTANTS.WORKSPACE_FILE
   end
+  if not path then
+    path = vim.fs.joinpath(vim.fn.getcwd(), "codecompanion-workspace.json")
+    CONSTANTS.WORKSPACE_FILE = vim.fs.joinpath(vim.fn.getcwd(), "codecompanion-workspace.json")
+  end
+
   if not vim.uv.fs_stat(path) then
     return log:warn(fmt("Could not find a workspace file at `%s`", path))
   end


GitMurf:
Not sure if this is another Windows issue 🤷‍♂️

olimorris:
Really weird. I can confirm the same. Thanks for fixing either way.

GitMurf:
@olimorris fyi may be worth figuring out the "why" because it could be something that bites us somewhere else down the road too ;)

I tried to figure out where it could be going wrong but cannot and it doesn't make sense! The only thing I can think of is something mutates it "by ref" accidentally somewhere else up/down stream 🤷‍♂️ It certainly wasn't obvious though where / how that could have been based on the time I spent trying to trace it (but did not spend too much time).

GitMurf:
> The only thing I can think of is something mutates it "by ref" accidentally somewhere else up/down stream

Assuming lua works the same way as javascript, the path as a string should not mutate up/down stream so all I can think of is the parent constant object (table) `CONSTANTS.WORKSPACE_FILE` got reset or mutated somewhere else 🤷‍♂️ ??

olimorris:
I've put this on my "lower priority" list. Still a priority but I wanted to tweak my workspace file for the plugin and then use it for problems such as this.

I think realistically, the constant manipulated has to be the answer. Wonder if setting path to a deep copy might fix this.

PR #770, opened 2025-01-29
fix: handle multiline userdata in copilot adapter

rachartier:
## Description

Hi, thank you for the plugin.

I encountered a bug with the copilot provider: when opening the chat buffer and entering text, then pressing `Enter`, it crashes by outputting:

image

When looking at the source code, I found that you'll read only one line of the file:
```lua
      local userdata = vim.json.decode(vim.fn.readfile(file_path)[1])
```

`vim.fn.readfile` returns a list of lines.

So this PR will adapt the code to read the content and then concatenate all lines of this content to pass to `json.decode`.

Maybe `vim.json.decode(vim.fn.readfile(file_path)[1])` was intended to fix an edge case.

Do not hesitate to tell me if something needs to be changed or to explain why it was done that way.

Thanks!

## Checklist

### DIFF
--- a/lua/codecompanion/adapters/copilot.lua
+++ b/lua/codecompanion/adapters/copilot.lua
@@ -60,7 +60,13 @@ local function get_token()
 
   for _, file_path in ipairs(file_paths) do
     if vim.uv.fs_stat(file_path) then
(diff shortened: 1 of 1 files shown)


olimorris:
Amazing catch! Thank you

PR #798, opened 2025-01-31
docs: add all docs to panvimdoc generation

milisims:
## Description

Adds scripts necessary to use panvimdoc to generate all docs.

The added `scripts/panvimdoc-cleanup.lua` is a pandoc filter that makes a few changes when generating the docs:

- A few headers are changed (see the table) to fit better in the text width
- Emphasis (italics in the docs website) are converted to backtick wrapped code elements, for a similar effect
- Some extra handling to fix `_@this_`, because it’s recognized as a citation
- I try to fix all of the links, but a few aren’t there yet. 

As is, there are a few links that aren’t properly found as tags. When you have relative links in the docs, it’s not easy to track where they came from. So for example, [L438](https://github.com/olimorris/codecompanion.nvim/blob/146ce88607acc4d6f7f7c96a0352b0c90ad6a2fd/doc/codecompanion.txt#L438) has a link to `|codecompanion-prompt-library|` when it should be `|codecompanion-configuration-prompt-library| `.

I think the two easiest fixes, plus the hardest are:

- In make, copy to a temporary directory all of the doc files, editing the relative links to absolute ones
- Edit all of the links in the docs source to be absolute links, and add a check in the makefile that throws an error in `make docs` if any relative links are found, with their locations and suggested changes (similar to above)
- handle it all in the lua filter (probably the most complicated)

What do you think?

## Screenshots

Screenshot from 2025-01-30 18-31-22

## Checklist

- [x] Fix header width
- [x] Fix Emphasis
- [x] Fix 'citations'
- [ ] Fix links

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -13,8 +13,8 @@ docs: deps/panvimdoc
 		--metadata="ignorerawblocks:true" \
 		--metadata="docmapping:false" \
 		--metadata="docmappingproject:true" \
-		--lua-filter=deps/panvimdoc/scripts/include-files.lua \
-		--lua-filter=deps/panvimdoc/scripts/skip-blocks.lua \
+		--lua-filter deps/panvimdoc/scripts/include-files.lua \
+		--lua-filter deps/panvimdoc/scripts/skip-blocks.lua \
 		--lua-filter scripts/panvimdoc-cleanup.lua \
 		-t deps/panvimdoc/scripts/panvimdoc.lua \
 		scripts/vimdoc.md \
--- a/Makefile
+++ b/Makefile
@@ -1,21 +1,24 @@
 all: format docs test
 
 docs: deps/panvimdoc
-	@echo Generating Docs... && \
-	cd deps/panvimdoc && \
-	pandoc \
+	@echo Generating Docs...
+	@pandoc \
 		--metadata="project:codecompanion" \
 		--metadata="vimversion:NVIM v0.10.0" \
 		--metadata="titledatepattern:%Y %B %d" \
 		--metadata="toc:true" \
 		--metadata="incrementheadinglevelby:0" \
 		--metadata="treesitter:true" \
-		--lua-filter scripts/skip-blocks.lua \
-		--lua-filter scripts/include-files.lua \
-		--lua-filter scripts/remove-emojis.lua \
-		-t scripts/panvimdoc.lua \
-		../../README.md \
-		-o ../../doc/codecompanion.txt
+		--metadata="dedupsubheadings:true" \
+		--metadata="ignorerawblocks:true" \
+		--metadata="docmapping:false" \
+		--metadata="docmappingproject:true" \
+		--lua-filter=deps/panvimdoc/scripts/include-files.lua \
+		--lua-filter=deps/panvimdoc/scripts/skip-blocks.lua \
+		--lua-filter scripts/panvimdoc-cleanup.lua \
+		-t deps/panvimdoc/scripts/panvimdoc.lua \
+		scripts/vimdoc.md \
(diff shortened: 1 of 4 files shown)


Makefile
```diff
@@ -1,21 +1,24 @@
...
-		--lua-filter scripts/include-files.lua \
-		--lua-filter scripts/remove-emojis.lua \
-		-t scripts/panvimdoc.lua \
-		../../README.md \
-		-o ../../doc/codecompanion.txt
+		--metadata="dedupsubheadings:true" \
+		--metadata="ignorerawblocks:true" \
+		--metadata="docmapping:false" \
+		--metadata="docmappingproject:true" \
+		--lua-filter=deps/panvimdoc/scripts/include-files.lua \
```
olimorris, line 16: Can we run the `remove-emojis.lua` script to remove the GitHub emojis?

milisims: Demojify just removes actual emoji characters, I was leaning on looking them up somehow (like [here](https://github.com/github/gemoji/blob/master/db/emoji.json) or [here](https://github.com/muan/unicode-emoji-json/blob/main/data-by-emoji.json)) or just removing `:these_strings:`. What do you think?

olimorris: It definitely removes GitHub emojis. I asked kdheepak to [implement](https://github.com/kdheepak/panvimdoc/issues/26) it a couple of years ago. Should just be a `--lua-filter scripts/remove-emojis.lua` call. Lines 58 to 65 in CodeCompanion.txt is now showing them

milisims: Lol, I added it before the `include-files` script so it didn't do anything when I tried it :upside_down_face:.
That said, it's causing some issues:
image

Instead, I just added a simple filter in the pandoc filters to remove it, removing the emoji codes without any issues.

olimorris: Amazing. Notice you've still got the "Fix Links" in your checklist, is it worth merging this anyway?

milisims: In the PR desc:

As is, there are a few links that aren’t properly found as tags. When you have relative links in the docs, it’s not easy to track where they came from. So for example, [L438](https://github.com/olimorris/codecompanion.nvim/blob/146ce88607acc4d6f7f7c96a0352b0c90ad6a2fd/doc/codecompanion.txt#L438) has a link to `|codecompanion-prompt-library|` when it should be `|codecompanion-configuration-prompt-library|`.

I think the two easiest fixes, plus the hardest are:

- In make, copy to a temporary directory all of the doc files, editing the relative links to absolute ones
- Edit all of the links in the docs source to be absolute links, and add a check in the makefile that throws an error in make docs if any relative links are found, with their locations and suggested changes (similar to above)
- handle it all in the lua filter (probably the most complicated)
What do you think?

milisims: fwiw these are all relative links found:
```
configuration/adapters
configuration/prompt-library
configuration/adapters
configuration/prompt-library
configuration/prompt-library
prompt-library
chat-buffer#diff
#keymaps
```
And only the last 3 have any issue.

olimorris: I have to be honest and say I never actually use vim docs for plugins...only just the internal Neovim documentation. Do you think have a few incorrect links would impact users? My view is the work you've done so far is immense and gives any vim doc users a huge benefit. From what you shared above, it sounds like it could be a nuisance to keep track of the links and I don't want to sacrifice any of the source doc linkage

milisims: I use them a lot clearly :smile:, I'd say just merge it then.

olimorris:
Sounds good to me. 

Once again, thank you so much for this. A real QOL improvement and I know there are still a number of people who use the docs so this will be a huge help.

PR #812, opened 2025-02-01
feat(adapters): add reasoning effort to copilot and openai

olimorris:
## Description

Add the [reasoning_effort](https://platform.openai.com/docs/api-reference/chat/create#chat-create-reasoning_effort) parameter to both Copilot and OpenAI adapters.

This also updates the Copilot adapter to point to a versioned model of `o3-mini`.


olimorris:
cc: @bassamsdata

bassamsdata:
Thanks a bunch for adding it so quickly!

PR #819, opened 2025-02-01
Make ui chat title configurable

petobens:
## Description

We can now define both a custom title for the chat window and a higlight group by doing something like:
```lua
window = {
    title = { { 'CodeCompanion', 'TelescopeTitle' } },
```

## Related Issue(s)

https://github.com/olimorris/codecompanion.nvim/discussions/815

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/ui.lua
+++ b/lua/codecompanion/strategies/chat/ui.lua
@@ -86,7 +86,7 @@ function UI:open()
       row = window.row or math.floor((vim.o.lines - height) / 2),
       col = window.col or math.floor((vim.o.columns - width) / 2),
       border = window.border,
-      title = "CodeCompanion",
+      title = window.title or "CodeCompanion",
       title_pos = "center",
       zindex = 45,
     }


olimorris:
As I mentioned in the discussion, this will overwrite the default `Title` highlight. Also, what about users who don't use Telescope?

Can we set this to `CodeCompanionChatTitle` and have it link to `Title` like in [plugin/codecompanion.lua](https://github.com/olimorris/codecompanion.nvim/blob/main/plugin/codecompanion.lua)

petobens:
Sorry I might no follow but the Telescope highlight group is just an example from my own config:
```lua
    display = {
        chat = {
            intro_message = '',
            show_settings = false,
            window = {
                layout = 'float',
                border = 'rounded',
                title = { { 'CodeCompanion', 'TelescopeTitle' } },
                height = 0.945,
                width = 0.45,
                relative = 'editor',
                col = vim.o.columns, -- right position
                opts = {
                    relativenumber = false,
                    number = false,
                },
            },
        },
    },
```
but there is nothing related to telescope at all in the proposed changes. The only thing this PR actually does is that it makes such title window configurable. If you want I can do something like this in the actual change to the code:
```lua
title = window.title or { { 'CodeCompanion', 'FloatTitle' } },
```
but I don't believe that is necessary since `:h api.nvim_open_win` shows that:
```
                  • title: Title (optional) in window border, string or list.
                    List should consist of `[text, highlight]` tuples. If
                    string, or a tuple lacks a highlight, the default
                    highlight group is `FloatTitle`.
```

olimorris:
Ahhh I see. Sorry, totally misunderstood what you wanted.

PR #872, opened 2025-02-08
feat: blink completions on invoke, without trigger char

Saghen:
## Description

At the moment, triggering completions at `/buf|` won't show items because the `ctx.trigger.kind` is `Invoked` rather than `TriggerCharacter`. This changes the get completions logic to use either the `ctx.trigger.character` or the character before the bounds of the current keyword (keyword is `buf` in the previous example). A few other changes:

- Adds a `callback()` when the plugin decides not to show completion items, so that blink.cmp doesn't consider the source to be hanging
- Adds type checking when `lazydev.nvim` is installed
- Uses `item.data` for source specific data

## Related Issue(s)

N/A

## Screenshots

N/A

## Checklist

### DIFF
--- a/lua/codecompanion/providers/completion/blink/init.lua
+++ b/lua/codecompanion/providers/completion/blink/init.lua
@@ -1,9 +1,12 @@
+--- @module 'blink.cmp'
+
 local completion = require("codecompanion.completion")
 
 local slash_commands = completion.slash_commands()
 local tools = completion.tools()
 local vars = completion.variables()
 
+--- @class blink.cmp.Source
 local M = {}
 
 function M.new()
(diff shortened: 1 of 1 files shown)


Saghen:
Out of curiosity, would you be interested in an interface where `/buffer |` shows a list of available buffers which you can select via autocomplete, rather than using vim input?

olimorris:
Thanks for continuing to look after CodeCompanion! It means a lot and your productivity on GitHub is unrivaled 🫡.

> Out of curiosity, would you be interested in an interface where /buffer | shows a list of available buffers which you can select via autocomplete, rather than using vim input?

Oh absolutely. That would be incredibly great from a UI perspective.

Saghen:
Great! I'll take a look at implementing that in a separate PR since I image it would be a bigger change

Saghen:
Lmk if there's anything else you need from me. I'm working on the immediate closing issue in https://github.com/Saghen/blink.cmp/issues/1303, which I might need to make another PR for

Issue #92, opened 2024-08-10
[Bug]: `inline` strategy no longer inject the response into the current buffer


huantrinh1802:
### Your `minimal.lua` config

```
  {
    "olimorris/codecompanion.nvim",
    dependencies = {
      "nvim-lua/plenary.nvim",
      "nvim-treesitter/nvim-treesitter",
      "nvim-telescope/telescope.nvim", -- Optional
      {
        "stevearc/dressing.nvim", -- Optional: Improves the default Neovim UI
        opts = {},
      },
    },
    config = function()
      local session_key = vim.fn.system("cat $TMPDIR/bw_session")
      require("codecompanion").setup({
        adapters = {
          openai = require("codecompanion.adapters").use("openai", {
            env = {
              api_key = ""
            },
            schema = { model = { default = "gpt-4o" } },
          }),
          anthropic = require("codecompanion.adapters").use("anthropic", {
            env = {
              api_key = "",
            },
            schema = { model = { default = "claude-3-5-sonnet-20240620" } },
          }),
        },
        default_prompts = {
          ["Auto-generate git commit message"] = {
            strategy = "inline",
            description = "Generate git commit message for current staged changes",
            opts = {
              mapping = "<LocalLeader>aacm",
            },
            prompts = {
              {
                role = "user",
                contains_code = true,
                content = function()
                  return [[You are an expert at following the Conventional Commit specification                       based on the following diff:
]] .. vim.fn.system("git diff --cached") .. [[

Generate a commit message for me. Follow the below structure:

   [dev | modify | fix] {Summary}

   {List of details if necessary using bullets}

Return the code only and no markdown codeblocks.
                    ]]
                end,
              },
            },
          },
        },
      })
      vim.api.nvim_set_keymap("n", "<leader>aa", "<cmd>CodeCompanionActions<cr>", { noremap = true, silent = true })
      vim.api.nvim_set_keymap("v", "<leader>aa", "<cmd>CodeCompanionActions<cr>", { noremap = true, silent = true })
      vim.api.nvim_set_keymap("n", "<leader>at", "<cmd>CodeCompanionToggle<cr>", { noremap = true, silent = true })
      vim.api.nvim_set_keymap("v", "<leader>at", "<cmd>CodeCompanionToggle<cr>", { noremap = true, silent = true })
    end,
  },
```

### Health check output

codecompanion: require("codecompanion.health").check()

codecompanion.nvim report ~
[...]

PR #904, opened 2025-02-12
Inline Assistant v2

olimorris:
## Progress

- System prompt has been revamped - much more structured and logical
- LLMs now return XML:

```xml
<response>
  <code>print('Hello World')</code>
  <language>python</language>
  <placement>add</placement>
</response>
```

This makes it much easier to parse code and errors if they occur:

```xml
<response>
  <error>Unable to process the prompt as it appears to be random characters without any clear instruction or meaning.</error>
</response>
```

- If the LLM returns XML in a markdown code block then we use Tree-sitter to attempt to parse it (something which was common for small parameter LLMs and Gemini 2.0). This will make the Inline Assistant much more resilient, however, this means that streaming is no longer possible
- Improved test coverage
- Adapter tests now test for inline output
- Can now pass in the adapter as the first argument to the prompt
- Prompts can now make use of variables
- Added two new variables: `#chat` and `#buffer`


taketwo:
In one of my custom inline prompts I'm using `placement = 'before|false'` to place the response in the beginning of the buffer and auto-accept the diff (suggested [here](https://github.com/olimorris/codecompanion.nvim/discussions/694#discussioncomment-11970408)). Seems like the auto-acceptance does not work anymore. Is this accidental or was the feature purposefully removed?

nshen:
After the v2 update, my :CodeCompanion commands have almost stopped working. I've tried Copilot, DeepSeek, and Alibaba's qwen-omni-turbo-latest models.

olimorris:
@nshen if this is the case, then please raise a support ticket so I can reproduce the issue.

olimorris:
> In one of my custom inline prompts I'm using `placement = 'before|false'` to place the response in the beginning of the buffer and auto-accept the diff (suggested [here](https://github.com/olimorris/codecompanion.nvim/discussions/694#discussioncomment-11970408)). Seems like the auto-acceptance does not work anymore. Is this accidental or was the feature purposefully removed?

The `placement|boolean` approach was discarded in the plugin many months ago. Apologies if my docs have been slow to reflect that. Also, auto-acceptance has never been a feature so I'm not sure where that came from? Do you mean auto placement? Because that's still very much supported for prompt library items.

taketwo:
To be clear: I've never seen `placement|boolean` being documented anywhere. This syntax was suggested by a kind soul in the above-referenced comment. 

> The placement|boolean approach was discarded in the plugin many months ago.

Yet it worked well until this last release.

> Also, auto-acceptance has never been a feature so I'm not sure where that came from?

It was suggested here: https://github.com/olimorris/codecompanion.nvim/issues/92#issuecomment-2283843335.

PR #925, opened 2025-02-15
Add config option to show only custom/defined adapters with change ad…

petobens:
…apter keymap

## Description

I want to only display my defined  adapters when using the `change_adapter` keymap.

## Related Issue(s)

https://github.com/olimorris/codecompanion.nvim/discussions/873#discussioncomment-12107770 

## Screenshots

Rather than displaying every single adapter now the list shows only adapters defined in the config file.

image

## Checklist

### DIFF
--- a/doc/configuration/adapters.md
+++ b/doc/configuration/adapters.md
@@ -215,3 +215,20 @@ require("codecompanion").setup({
   },
 }),
 ```
+
+## Hiding Default Adapters
+
+By default, the plugin shows all available adapters, including the defaults. If you prefer to only display the adapters defined in your user configuration, you can set the `show_defaults` option to `false`:
+
+```lua
+require("codecompanion").setup({
+  adapters = {
+    opts = {
+      show_defaults = false,
+    },
+    -- Define your custom adapters here
+  },
+})
+```
+
+When `show_defaults` is set to `false`, only the adapters specified in your configuration will be used, hiding the default ones provided by the plugin.
--- a/lua/codecompanion/config.lua
+++ b/lua/codecompanion/config.lua
@@ -26,7 +26,7 @@ local defaults = {
     -- OPTIONS ----------------------------------------------------------------
     opts = {
       allow_insecure = false, -- Allow insecure connections?
-      only_custom = false, -- Only use/show custom adapters
+      show_defaults = true, -- Show default adapters
       proxy = nil, -- [protocol://]host[:port] e.g. socks5://127.0.0.1:9999
     },
   },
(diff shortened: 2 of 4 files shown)


petobens:
I do have several questions:
i) is checking for the `function` type the correct way of retrieving  user defined adapters?
ii) in my config I define:
```lua
    adapters = {
        openai_gpt_4o = function()
            return adapters.extend('openai', {
                env = { api_key = OPENAI_API_KEY },
                schema = {
                    model = { default = 'gpt-4o' },
                    max_tokens = { default = 2048 },
                    temperature = { default = 0.2 },
                    top_p = { default = 0.1 },
                },
            })
        end,
        openai_o3_mini = function()
            return adapters.extend('openai', {
                env = { api_key = OPENAI_API_KEY },
                schema = {
                    model = { default = 'o3-mini-2025-01-31' },
                },
            })
        end,
    },
```
But you can see in my screenshot above that the first listed adapters is `openai`  that's because 
```lua
local current_adapter = chat.adapter.name
```
return `openai` even for my custom adapters. Can that be avoided?
iii) I guess that if the config option is accepted I need to update help + some other config file. Please point me towards that
iv) As I mentioned in the discussion what I'm really trying to achieve is switching between openai models rather than adapters. Maybe this PR could be extended to do that?

olimorris:
This approach is going to be a maintenance headache because adapters crop up all over the plugin. It would be better to implement this change at source, i.e. `codecompanion.setup`.

Alternatively, users can set adapters they don't want to `{}`, then that could be picked up at setup and removed from the config adapters table.

petobens:
>  It would be better to implement this change at source, i.e. codecompanion.setup.

> Alternatively, users can set adapters they don't want to {}, then that could be picked up at setup and removed from the config adapters table.

Can you elaborate a bit further what do you mean by this? I can then give it a shot.

petobens:
I pushed a commit which I think respects your idea. Please let me know if that's  what you were actually hinting. 

Even with this approach I still see the plain `openai` adapter listed. I guess that is intended since my other  2 adapters actually extend the openain adapter. Dunno if it can be removed.

olimorris:
Looks good. Coming at it fresh this morning, I think we should rename it `show_defaults` as it just reads better in the context of `config.adapters.opts.show_defaults`. Can you also add an example at the bottom of `doc/configuration/adapters.md` keeping in line with the examples that I've created? Perhaps ask an LLM to write the example in the same format as the other examples.

petobens:
Please check my last commits. Happy to make further changes if needed.

olimorris:
Love it. Thanks for a solid PR and documenting it 🚀

Issue #905, opened 2025-02-12
[Bug]: can't set debug window size


petobens:
### Your `minimal.lua` config

```lua
local root = '/tmp/nvim-minimal'

-- Set stdpaths to use root dir
for _, name in ipairs({ 'config', 'data', 'state', 'cache' }) do
    vim.env[('XDG_%s_HOME'):format(name:upper())] = root .. '/' .. name
end

-- Bootstrap lazy
local lazypath = root .. '/plugins/lazy.nvim'
if not vim.loop.fs_stat(lazypath) then
    vim.fn.system({
        'git',
        'clone',
        '--filter=blob:none',
        '--single-branch',
        'https://github.com/folke/lazy.nvim.git',
        lazypath,
    })
end
vim.opt.runtimepath:prepend(lazypath)

-- Install plugins
local plugins = {
    {
        'olimorris/codecompanion.nvim',
        dependencies = {
            'nvim-lua/plenary.nvim',
            'nvim-treesitter/nvim-treesitter',
        },
        config = function()
            require('codecompanion').setup({
                adapters = {
                    openai_gpt_4o = function()
                        return require('codecompanion.adapters').extend('openai', {
                            env = { api_key = 'cmd:pass show openai/yahoomail/apikey' },
                            schema = {
                                model = { default = 'gpt-4o' },
                            },
                        })
                    end,
                },
                display = {
                    chat = {
                        window = {
                            layout = 'float',
                            height = vim.o.lines - 5,
                            width = 0.45,
                            relative = 'editor',
                            col = vim.o.columns,
                            row = 1,
                        },
                    },
                    debug_window = {
                        width = 10,
                        height = 5,
                    },
                },
                strategies = {
                    chat = {
                        adapter = 'openai_gpt_4o', -- default adapter
                    },
                },
            })
        end,
    },
}

require('lazy').setup(plugins, {
    root = root .. '/plugins',
})
```

### Error messages

None

### Log output

None

### Describe the bug

The debug window height and width settings doesn't seem to work.

### Reproduce the bug

As in the GIF:
1. Open nvim with the minimal init.lua above
2. Run `:CodeCompanionChat`
3. Press `gd` to open the debug window

After 3. you can see that the debug window width and height are not honored. 

![Image](https://github.com/user-attachments/assets/1d4ffd0c-60cb-4331-9a9e-2ee69d64e02b)

### Adapter and model

Openai - gpt-4o

### Final checks

[...]

PR #926, opened 2025-02-15
Fix setting  debug window height and width

petobens:
See #905


petobens:
Mmm i need to do some further tweaking. Because I call teh debug window and toggle the chat window; the  width seems to change

olimorris:
Awesome, this looks great.

PR #934, opened 2025-02-17
feat: add adapter for mistral ia

josealvaradoo:
## Description

Add an adapter for [Mistral IA](console.mistral.ai). Why? Because I've tried using the OpenAI compatible adapter but Mistral rejects the request because some parameters in the request payload.

## Screenshots

image

image

## Checklist

### DIFF
--- a/lua/codecompanion/adapters/mistral.lua
+++ b/lua/codecompanion/adapters/mistral.lua
@@ -194,28 +194,6 @@ return {
         "mistral-small",
       },
     },
-    reasoning_effort = {
-      order = 2,
-      mapping = "parameters",
-      type = "string",
-      optional = true,
-      condition = function(schema)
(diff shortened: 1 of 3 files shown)


olimorris:
I thought the Mistral API was compatible with OpenAI?

I would prefer for users to inherit from the `openai_compatible` adapter and have this in their own config, rather than have the plugin support it.

This will be the plugin's default position moving forward. I'm happy for this to be referenced in the docs, however.

S1M0N38:
If you are planning to use Codestral model,
`https://codestral.mistral.ai/v1/chat/completions`
it's the [suggested endpoint](https://docs.mistral.ai/capabilities/code_generation/) (and it's free).

Like `api.mistral.ai`, it's "openai-compatible", so no need to implement a custom adapter.

PR #943, opened 2025-02-18
make stderr capturing work for "streaming" job output

JeanMertz:
I noticed when asking the LLM to run a command, my `cargo build` errors weren't being logged and instead the response was:

> There was an error from the command `cargo build`:
> 
> ```txt
> Tool failed with code 101 and no output
> ```

With this change, the error is correctly captured and added to the prompt:

> There was an error from the command `cargo build`:
> 
> ```txt
> bunch of stderr output here...
> ```

This might not be the correct way to solve the issue, but in the logs I always see the `err` variable as `nil`, and Plenary docs aren't very helpful on understanding what the variable represents. Feel free to close this and solve it the proper way, but this _worked on my machine_.

### DIFF
--- a/lua/codecompanion/strategies/chat/tools/init.lua
+++ b/lua/codecompanion/strategies/chat/tools/init.lua
@@ -363,7 +363,11 @@ function Tools:run()
             table.insert(strip_ansi(stdout), data)
           end)
         end,
-        on_stderr = function(err, _)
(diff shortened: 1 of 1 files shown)


olimorris:
Great stuff thanks for this.

At some point this week I'm going to move tools from leveraging `Job.new():start()` as I want to be strict over the order of completion

PR #949, opened 2025-02-18
fix(inline): ensure placement is in lower case

olimorris:
## Description

Gemini is marvellous to talk to and a pain in the backside to program with. Instead of returning the placement in lower case like every LLM, it likes to capitalise it...


olimorris:
@towry it's working for me:

2025-02-22 16_46_47 - WezTerm

Perhaps your LLM is doing something funky.

olimorris:
The PR isn't the place to raise an issue. Raise a bug report or better still, investigate the issue and raise an appropriate PR.

PR #955, opened 2025-02-19
feat(chat): when `stderr` is not empty, do not discard `stdout` when exit code is 0

Davidyz:
## Description

Currently, the tool class discard `stdout` when `stderr` is not empty even if the exit code is 0. In the real world a lot of programs/libraries (in my case, [VectorCode](https://github.com/Davidyz/VectorCode/wiki/Neovim-Integrations#tool)) use `stderr` for warnings that don't actually terminate the command. The current way in which `stdout` and `stderr` are handled caused false negatives in some cases (outputs with useful information get thrown away).

> I'm not sure if it's appropriate to call both `output.error` and `output.success`. I simply copied the code under `elseif not vim.tbl_isempty(stdout) then` here.

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/tools/init.lua
+++ b/lua/codecompanion/strategies/chat/tools/init.lua
@@ -404,6 +404,11 @@ function Tools:run()
               output.error(cmd, strip_ansi(stderr))
               log:debug("[Tools] %s finished with stderr: %s", self.tool.name, stderr)
               stderr = {}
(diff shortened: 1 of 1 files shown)


olimorris:
Great inclusion, thanks.

I'm currently working on refactoring the tools class. Hopefully finished tomorrow. No changes to functionality but it needs splitting out into separate files, so hopefully easier to understand.

PR #965, opened 2025-02-20
refactor tools

olimorris:
## Description

This PR sets out to streamline the agents and tools implementation in the plugin, with a view to moving to native tool/function calling in the near future. Currently, all the logic sits within a tools/init.lua file. It's difficult to distinguish between function tools and command tools. And every time I have to get my head back into the code, I need to refresh myself on how I've built the solution.

This PR sets out a significant number of improvements.

### Test Coverage

There is now significant test coverage for the feature. Thanks to mini.test's child processes capability, it's now possible to run tool tests from within a separate Neovim process. This saves me from having to mock Plenary.job and to test the feature in the way users do in the real world.

### Queuing system (FIFO)

There were edge cases with the previous implementation whereby if a user ran a function tool, then a command tool, then another function tool, the two function tools would complete before the command tool. For some cases, this might not have been an issue, but in the example that I give in the README, i.e. asking the LLM to create Snake, this is a definite flaw. In fact, it turns out the only reason that video worked at all was because the human authorization delayed the function tools enough for the command tool to complete. But in this update, all of that goes away and the plugin will only execute the next tool if the previous one has completed.

### Refactor

There's also been a significant refactor in the plugin as well. Instead of everything being in a tool/init file, it is now structured as follows:

```
strategies/chat/agents
├── executor/
│   ├── cmd.lua
│   ├── func.lua
│   ├── init.lua
│   ├── queue.lua
├── tools/
│   ├── cmd_runner.lua
│   ├── editor.lua
│   ├── files.lua
│   ├── rag.lua
├── init.lua
```

My hope is that it's much less daunting for users to make contributions to this particular feature in the plugin, and that it also provides a scalable base for the next phase of enhancements.


doc/extending/tools.md
```diff
@@ -6,22 +6,22 @@ In CodeCompanion, tools offer pre-defined ways for LLMs to execute actions and a
 
 In the plugin, tools work by sharing a system prompt with an LLM. This instructs them how to produce an XML markdown code block which can, in turn, be interpreted by the plugin to execute a command or function.
```
lazymaniac, line 7: Some simple but complete example how to add a tool into the config would be awesome. Something similar to prompts library docs.

olimorris: I'll do this in an upcoming PR. I'm going to move away from this XML approach now a lot of LLMs have native function calling. This PR is to add a tonne of tests and refactor ahead of this.

PR #966, opened 2025-02-20
fix(chat): add missing write flag attribute when editing files

michaldziurowski:
## Description
Previously, the return values of `gsub` were passed to `p:write`. This is incorrect because `gsub` returns two values: the changed data and the number of substitutions made. Since the method signature for `write` is `write(txt, flag, mode)`, the number of substitutions returned by `gsub` was incorrectly passed as the flag attribute. This could lead to incorrect behavior.

Passing `"w"` explicitly means that the file will be opened for writing, and if it has any content, it will be truncated.

I also modified the test so that it would fail with the previous implementation. The reason for the failure is that when invoking `write` with the changed text and a flag of 1 (1 is the number of substitutions returned by gsub, and by accident it also corresponds to the O_WRONLY flag), the file is not truncated. Since the substitution text is shorter than the original file content, some leftovers remain in the file.

P.S. I've read the contributing guide but decided not to start a discussion since it seems like an obvious oversight. I hope you don't mind. :)

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/tools/files.lua
+++ b/lua/codecompanion/strategies/chat/tools/files.lua
@@ -83,11 +83,12 @@ local function edit(action)
     return util.notify(fmt("No data found in %s", action.path))
   end
 
-  if not content:find(vim.pesc(action.search)) then
+  local changed, substitutions_count = content:gsub(vim.pesc(action.search), vim.pesc(action.replace))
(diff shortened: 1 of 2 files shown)


michaldziurowski:
>Great catch

As often is such situations it was pure luck :)
Thanks for your hard work and this great plugin.

PR #980, opened 2025-02-21
fix(adapter): variable substitution in 'raw'.

pettni:
Pass adapter.raw through adapter:set_env_vars() to enable variable substitution.

Also appends adapter.raw to the static raw arguments defined in the http client.

## Description

Implements the change discussed in #974 that makes it easier to write an adapter for AWS Bedrock Anthropic models.

## Testing

Tested with my custom Bedrock adapter, as well as a built-in adapter (deepseek).

### DIFF
--- a/lua/codecompanion/http.lua
+++ b/lua/codecompanion/http.lua
@@ -101,12 +101,16 @@ function Client:request(payload, actions, opts)
     table.insert(raw, "--no-buffer")
   end
 
+  if adapter.raw then
+    vim.list_extend(raw, adapter:set_env_vars(adapter.raw))
+  end
+
   local request_opts = {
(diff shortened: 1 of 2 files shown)


olimorris:
Awesome, thank you for this. Do you think you could add a singular test case to the [adapter test file](https://github.com/olimorris/codecompanion.nvim/blob/main/tests/test_adapter.lua) too? I'm keen to make sure any features that people add, we support with tests so I don't accidentally break them in the future.

pettni:
Sure, always a good idea to have some test. I added one in the http test file that tests that variable substitutions have been done in the data that is going to `curl`

olimorris:
Fantastic. Thanks for such a great PR and would love to see your AWS Bedrock implementation too.

PR #1012, opened 2025-02-25
feat(copilot): get available models from copilot endpoint

S1M0N38:
## Related Issue(s)

From discussion #858
https://github.com/olimorris/codecompanion.nvim/discussions/858#discussioncomment-12311109

## Screenshots

Screenshot 2025-02-25 at 16 21 31

During the development of this feature I was able to access "claude-3.7". Now I can no longer access claude 3.7 in vscode. Maybe they have [restricted the access](https://github.blog/changelog/2025-02-24-claude-3-7-sonnet-is-now-available-in-github-copilot-in-public-preview/)? (im on student plan)

## Checklist

- [x] I've read the [contributing](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md) guidelines and have adhered to them in this PR
- [ ] I've updated the README and/or relevant docs pages
- [ ] I've run `make docs` to update the vimdoc pages

PR #1026, opened 2025-02-26
docs: improved CONTRIBUTING.md

S1M0N38:
The new CONTRIBUTING.md file was generated using claude 3.7 sonnet from the repo code imported in Anthropic project. The prompt was:
```
Rewrite the CONTRIBUTING.md including more information about the whole
project so that help new developer getting start contributing
(e.g. code structure, debugger, logging, linter,
```

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -118,7 +118,7 @@ A simple proxy server can be set up using [mitmproxy](https://mitmproxy.org/).
 From now on, all requests will be forwarded to the proxy server.
 <details>
 <summary>screenshot</summary>
-<img width="1506" alt="debug request with proxy screeshot" src="https://github.com/user-attachments/assets/60f31736-da83-4b80-bc61-341bb7fc82f7" />
+<img width="1506" alt="debug request with proxy screenshot" src="https://github.com/user-attachments/assets/60f31736-da83-4b80-bc61-341bb7fc82f7" />
 </details>
 
 With mitmproxy you can much more using custom scripts/hooks like simulating slower connections, patch requests, etc. Check out the [documentation](https://docs.mitmproxy.org/stable/addons-overview/) for more information.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -116,6 +116,10 @@ A simple proxy server can be set up using [mitmproxy](https://mitmproxy.org/).
 ```
 
 From now on, all requests will be forwarded to the proxy server.
+<details>
+<summary>screenshot</summary>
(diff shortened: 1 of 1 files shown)


S1M0N38:
I'm planning to explore codecompanion.nvim. I will check if the generated CONTRIBUTING.md sticks to the truth.

Let me know if there is some obvious error or hallucination.

olimorris:
This is a marked improvement already. Will be very happy to merge this when you're ready.

S1M0N38:
Are those contributing instructions reflecting your workflow?

olimorris:
There's some testing stuff that I haven't merged that would make the workflow much better but I can add that later.

S1M0N38:
It occurred to me that there is no documented way to reproduce issues/undesired behaviour caused by slow connection (e.g. interface responsiveness in https://github.com/olimorris/codecompanion.nvim/pull/1012#issuecomment-2686622910).

I came up with a relatively simple solution:

1. Configure a local proxy for codecompanion.nvim
```lua
opts = {
  adapters = {
    opts = {
      allow_insecure = true,
      proxy = "http://127.0.0.1:8080",
    },
  }
}
```

2. Start proxy on port 8080, which simulates a slower connection.

For example, using mitm with the following custom script, I can delay the requests and responses.

```python
from mitmproxy import http, ctx
import time


class ConnectionSlower:
    def __init__(self):
        self.delay_ms = 500  # milliseconds delay

    def request(self, flow: http.HTTPFlow) -> None:
        time.sleep(self.delay_ms / 1000)  # Convert ms to seconds

    def response(self, flow: http.HTTPFlow) -> None:
        time.sleep(self.delay_ms / 1000)  # Convert ms to seconds

    def load(self, loader):
        loader.add_option("delay_ms", int, 500, "Delay in milliseconds")

    def configure(self, updates):
        if "delay_ms" in updates:
            self.delay_ms = ctx.options.delay_ms


addons = [ConnectionSlower()]
```
```bash
mitmweb -s scripts/mitm-slower.py --set delay_ms=300
```

(Moreover, requests can be inspected in a web interface so it's easy to debug errors, wrong parameters, error codes, ...)

---

If you like this solution, I can add a `## Debugging Http Requests/Responses` section in CONTRIBUTING.md and maybe add the simple Python script to `scripts/`.

S1M0N38:
I didn't add the custom script for simulating slow connection. I keep this PR solely focused on CONTRIBUTING.md

olimorris:
Awesome stuff. Grazie, @S1M0N38.

PR #1092, opened 2025-03-09
feat: allow for full height chat windows when opening

SamSaffron:


olimorris:
Can you share some screenshots of what this looks like with and without this PR?

S1M0N38:
https://github.com/user-attachments/assets/10ae7be5-68f0-40a7-ae29-f502724c86d0

IMHO, `full_height=true` should be the default. In other editors (VSCode, Cursor, Zed), it usually considers a right panel with full height (like a file explorer on the left).

SamSaffron:
I agree with @S1M0N38 I think it is a far better default. 

Essentially the diff between `botright vsplit` and `vsplit` is: 


**Current**

image


**Proposed**

image


This feature is mainly important if you tend to use splits a lot, if you are a tab based user then this is a no-op.

olimorris:
Oh this looks great! Yeah let's absolutely set this as the default and then I can merge it. Thanks!

SamSaffron:
sure, no problem at all, thanks @olimorris made the change.

doc/configuration/chat-buffer.md
```diff
@@ -199,7 +199,7 @@ require("codecompanion").setup({
         height = 0.8,
         width = 0.45,
         relative = "editor",
-        full_height = false -- when set to true, code companion window will open using full editor height
+        full_height = true -- when set to false, vsplit will be used to open the chat buffer vs. botright/topleft vsplit
```
S1M0N38, line 202: missing comma

S1M0N38 (changes requested):

doc/configuration/chat-buffer.md
SamSaffron: thanks, sorted

S1M0N38 (approved):

Issue #1103, opened 2025-03-10
[Bug]: Type annotation conflicts in Executor


Davidyz:
### Did you check the docs and existing issues?

- [x] I have read all the docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched for existing issues/discussions
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.10.4

### Operating system/version

Arch Linux

### Describe the bug

[The documentation](https://github.com/olimorris/codecompanion.nvim/blob/7d6eb1279762032d3f638e7f72dd49bd76bfdf48/doc/extending/tools.md?plain=1#L174) states that a function-based tool should always return a lua table with 2 keys: `status` and `data`. This means the variable `output` [here](https://github.com/olimorris/codecompanion.nvim/blob/7d6eb1279762032d3f638e7f72dd49bd76bfdf48/lua/codecompanion/strategies/chat/agents/executor/func.lua#L85) would be a table of the same structure.

This introduced the diagnostics warning mentioned in #1101 because [`Executor:error`](https://github.com/olimorris/codecompanion.nvim/blob/7d6eb1279762032d3f638e7f72dd49bd76bfdf48/lua/codecompanion/strategies/chat/agents/executor/init.lua#L149) accepts the `output` argument as a string. 

There are multiple ways to fix this, and there seem to be other type annotation mismatch here (`data` can be any lua table, but in the type annotation of `Executor:success` and `Executor:error`, it's treated as `string?`), so I'm opening this issue to discuss how this should be fix (I'm getting a feeling that this might be beyond my knowledge of this project... so it'll be nice if @olimorris or some other collaborators can help).

### Steps To Reproduce

See #1101 

### Expected Behavior

Clear documentation on function-based tool and accurate type annotation of executors.

### init.lua file

```lua
--[[
NOTE: Set the config path to enable the copilot adapter to work.
It will search the following paths for a token:
  - "$CODECOMPANION_TOKEN_PATH/github-copilot/hosts.json"
  - "$CODECOMPANION_TOKEN_PATH/github-copilot/apps.json"
--]]
vim.env["CODECOMPANION_TOKEN_PATH"] = vim.fn.expand("~/.config")

vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

-- Your CodeCompanion setup
[...]

PR #1107, opened 2025-03-11
feat(tool): implement a callback for async function-based tool.

Davidyz:
## Description

This PR implements a `output_handler` parameter for the function tool, which can be called by an async callback used in the function tool body.

## Related Issue(s)

Discussed in #1103 

## Details

Previously a function-based tool has the following signature:
```
fun(agent, actions, input): {status: "success"|"error", data: any}
```
This PR introduces a 4th parameter, `output_handler`, which invokes the `executor:success` or `executor:error`. This allows users to build async function tools without blocking the UI. Example:
```lua
cmds = {
  function(agent, actions, input, output_handler)
    -- this is for demonstration only
    vim.lsp.client.request(lsp_method, lsp_param, function(err, result, _, _)
      output_handler({status = "success", data = result})
    end, buf_nr)
  end
}
```
I tested this with the calculator demo (as a synchronous function) and a WIP VectorCode tool that uses async LSP request. Both worked fine, and for the async tool, the virtual text `Tool processing ...` disappeared as usual. If this is a feasible implementation, I can update the documentation as well.

I'm a total noob at unit tests for neovim plugins, so that'll have to wait...

## Checklist

### DIFF
--- a/doc/extending/tools.md
+++ b/doc/extending/tools.md
@@ -164,16 +164,18 @@ end,
 
 **Function-based Tools**
 
-Function-based tools use the `cmds` table to define functions that will be executed one after another. Each function has three parameters, itself, the actions request by the LLM and any input from a previous function call. For the purpose of our calculator example:
+Function-based tools use the `cmds` table to define functions that will be executed one after another. Each function has three parameters, itself, the actions request by the LLM, any input from a previous function call and a `output_handler` callback for async execution. 
+For the purpose of our calculator example:
 
 ```lua
 cmds = {
   ---@param self CodeCompanion.Agent.Tool The Tools object
   ---@param actions table The action object
   ---@param input? any The output from the previous function call
+  ---@param output_handler fun(output: { status: "success"|"error", data: any })? callback for async tool
   ---@return nil|{ status: "success"|"error", data: any }
-  function(self, actions, input)
-    -- Get the numbers and operation requested by the LLM
+  function(self, actions, input, output_handler)
+-- Get the numbers and operation requested by the LLM
     local num1 = tonumber(actions.num1)
     local num2 = tonumber(actions.num2)
     local operation = actions.operation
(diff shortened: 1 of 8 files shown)


Davidyz:
we may also need to figure out some mechanism to limit the `output_handler` so that it can only be called once to avoid misuse/confusion.

olimorris:
Handling asynchronous functions is going to be quite complicated. In this [test](https://github.com/olimorris/codecompanion.nvim/blob/8823590f3cb251eb09315682bb3f40abf212969b/tests/strategies/chat/agents/executor/test_queue.lua#L35), you can see that I want to ensure that if multiple tools are called by an LLM, they are executed in the correct order. It's unlikely VectorCode is going to be called with multiple tools, but I want to stay true to my design principles here.

I think (without testing), that your PR would not adhere to them.

I was mulling over this scenario earlier today. I think immediately after calling an async function in your tool, a status of "async" should be returned which the executor picks up. It then sets a listener and stops, not pulling the next tool from the queue. Your async function could complete, fire an autocmd that the listener picks up and then starts the process of drawing from the queue again. This would enable all of the output handlers you've defined in your tool to work as expected.

Davidyz:
I might be wrong but I think `output_handler` delays the execution of `FuncExecutor:proceed_to_next()` to the end of the async function execution (the `callback` parameter is called at the end of `output_handler`), and this would also make sure `next_executor:orchestrate(output)` is only called when the output is ready. I'm not 100% sure on this, though, and I'll try to write some tests for this tmr to make sure of this.

olimorris:
That sounds great. We do have a small section on testing in the [CONTRIBUTING.md](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md) guide but give me a shout if you hit a block. Being able to mimic an asynchronous function in the test should be possible with how I've setup mini test in the example I provided above.

Davidyz:
@olimorris I've finished the tests, and they all passed. Could you review the tests and see if they've correctly addressed your concerns?

olimorris:
Amazing. Thanks for such a high quality PR.

Really appreciate all the effort that you've put into the documentation and also the testing. Just trying this out myself and everything works as expected. The only feedback is that the tests run a little slow. I wonder if we could decrease the times in the `defer_fn` without compromising the tests. Otherwise, this is absolutely amazing and we can merge asap.

Davidyz:
Hi, thanks for providing a high quality codebase for making this possible! I wanted the time in `defer_fn` to be long enough so that if the ordering is wrong it'll interfere with the `cmd` tool in the queue. The cmd tool runs `sleep 0.5` so there isn't much room to improve. It's possible though, to reduce the time for `vim.wait`. 3000 ms is definitely overkill for both test cases.

olimorris:
Thanks once again. I'm in the middle of the workspace PR which will be adding a prompt library item to generate a workspace file. I'll be conditionally loading the `@vectorcode` tool to help produce that. Very excited to see how that turns out.

Davidyz:
Now that the async function tool has been merged, it might be more performant to just create a built-in tool that can call LSP methods (go to definition, etc.). With this, we might be able to generate a basic "symbol relationship graph" that has the potential to destroy vectorcode lol

PR #1141, opened 2025-03-17
start refactoring

olimorris:
## Description

To be updated

### DIFF
--- a/lua/codecompanion/strategies/chat/init.lua
+++ b/lua/codecompanion/strategies/chat/init.lua
@@ -603,19 +603,10 @@ end
 
 ---Add the given tool to the chat buffer
 ---@param tool string The name of the tool
----@param tool_config table The tool from the config
 ---@return CodeCompanion.Chat
-function Chat:add_tool(tool, tool_config)
+function Chat:add_tool(tool)
   if self.tools_in_use[tool] then
     return self
-  end
-
-  -- Add the overarching agent system prompt first
-  if not self:has_tools() then
-    self:add_message({
-      role = config.constants.SYSTEM_ROLE,
-      content = config.strategies.chat.tools.opts.system_prompt,
-    }, { visible = false, reference = "tool_system_prompt", tag = "tool" })
   end
 
   local id = "<tool>" .. tool .. "</tool>"
@@ -626,14 +617,6 @@ function Chat:add_tool(tool, tool_config)
   })
 
   self.tools_in_use[tool] = true
-
-  local resolved = self.agents.resolve(tool_config)
-  if resolved then
-    self:add_message(
-      { role = config.constants.SYSTEM_ROLE, content = resolved.system_prompt(resolved.schema, xml2lua) },
-      { visible = false, tag = "tool", reference = id }
-    )
-  end
 
   util.fire("ChatToolAdded", { bufnr = self.bufnr, id = self.id, tool = tool })
 
--- a/lua/codecompanion/adapters/openai.lua
+++ b/lua/codecompanion/adapters/openai.lua
@@ -11,6 +11,7 @@ return {
   },
   opts = {
     stream = true,
+    tools = true,
   },
   features = {
     text = true,
@@ -75,6 +76,18 @@ return {
         :totable()
 
       return { messages = messages }
+    end,
+
+    ---Provides the schemas of the tools that are available to the LLM to call
+    ---@param self CodeCompanion.Adapter
+    ---@param tools table
+    ---@return table|nil
+    form_tools = function(self, tools)
+      if not self.opts.tools then
+        return
+      end
+
+      return { tools = tools }
     end,
 
     ---Returns the number of tokens generated from the LLM
--- a/lua/codecompanion/config.lua
+++ b/lua/codecompanion/config.lua
@@ -77,6 +77,10 @@ local defaults = {
             requires_approval = true,
           },
         },
+        ["weather"] = {
+          callback = "strategies.chat.agents.tools.weather",
+          description = "Get the latest weather",
+        },
         opts = {
           auto_submit_errors = false, -- Send any errors to the LLM automatically?
           auto_submit_success = false, -- Send any successful output to the LLM automatically?
--- a/lua/codecompanion/http.lua
+++ b/lua/codecompanion/http.lua
@@ -81,6 +81,7 @@ function Client:request(payload, actions, opts)
       handlers.form_parameters and handlers.form_parameters(adapter, adapter:set_env_vars(adapter.parameters), payload)
         or {},
       handlers.form_messages and handlers.form_messages(adapter, payload) or {},
+      handlers.form_tools and handlers.form_tools(adapter, payload) or {},
       adapter.body and adapter.body or {},
       handlers.set_body and handlers.set_body(adapter, payload) or {}
     )
--- a/lua/codecompanion/strategies/chat/agents/init.lua
+++ b/lua/codecompanion/strategies/chat/agents/init.lua
@@ -277,18 +277,7 @@ function Agent:parse(chat, message)
   if tools or groups then
     if tools and not vim.tbl_isempty(tools) then
       for _, tool in ipairs(tools) do
-        chat:add_tool(tool, self.tools_config[tool])
-      end
-    end
-
-    if groups and not vim.tbl_isempty(groups) then
-      for _, group in ipairs(groups) do
-        if self.tools_config.groups[group].system_prompt then
-          chat:add_message({
-            role = config.constants.SYSTEM_ROLE,
-            content = self.tools_config.groups[group].system_prompt,
-          }, { tag = "tool", visible = false })
-        end
+        chat:add_tool(tool)
       end
     end
     return true
--- a/lua/codecompanion/strategies/chat/agents/tools/weather.lua
+++ b/lua/codecompanion/strategies/chat/agents/tools/weather.lua
@@ -1,0 +1,49 @@
+return {
+  name = "weather",
+  cmds = {
+    function(city)
+      return {
+        status = "success",
+        data = "The weather in " .. city .. " is 75°F",
+      }
+    end,
+  },
+  schema = [[{
+    "type": "function",
+    "function": {
+        "name": "get_weather",
+        "description": "Retrieves current weather for the given location.",
+        "parameters": {
+            "type": "object",
+            "properties": {
+                "location": {
+                    "type": "string",
+                    "description": "City and country e.g. Bogotá, Colombia"
+                },
+                "units": {
+                    "type": "string",
+                    "enum": [
+                        "celsius",
+                        "fahrenheit"
+                    ],
+                    "description": "Units the temperature will be returned in."
+                }
+            },
+            "required": [
+                "location",
+                "units"
+            ],
+            "additionalProperties": false
+        },
+        "strict": true
+    }
+}]],
+  output = {
+    ---@param agent CodeCompanion.Agent
+    ---@param cmd table The command that was executed
+    ---@param stdout table
+    success = function(agent, cmd, stdout)
+      print("Weather: " .. stdout[1])
+    end,
+  },
+}
--- a/tests/adapters/test_openai.lua
+++ b/tests/adapters/test_openai.lua
@@ -21,6 +21,41 @@ T["OpenAI adapter"]["it can form messages to be sent to the API"] = function()
   h.eq({ messages = messages }, adapter.handlers.form_messages(adapter, messages))
 end
 
+T["OpenAI adapter"]["it can form tools to be sent to the API"] = function()
+  local tools = [[{
+    "type": "function",
+    "function": {
+        "name": "get_weather",
+        "description": "Retrieves current weather for the given location.",
+        "parameters": {
+            "type": "object",
+            "properties": {
+                "location": {
+                    "type": "string",
+                    "description": "City and country e.g. Bogotá, Colombia"
+                },
+                "units": {
+                    "type": "string",
+                    "enum": [
+                        "celsius",
+                        "fahrenheit"
+                    ],
+                    "description": "Units the temperature will be returned in."
+                }
+            },
+            "required": [
+                "location",
+                "units"
+            ],
+            "additionalProperties": false
+        },
+        "strict": true
+    }
+}]]
+
+  h.eq({ tools = tools }, adapter.handlers.form_tools(adapter, tools))
+end
+
 T["OpenAI adapter"]["Streaming"] = new_set()
 
 T["OpenAI adapter"]["Streaming"]["can output streamed data into the chat buffer"] = function()


ravitemer:
Hi @olimorris. Nice to see that you have started function calling. 

From the code, I see that tools no longer support system prompt. I understand that this keeps the tools super simple and easy to build. But will you consider keeping the system prompt? For e.g coming to mcphub.nvim's `@mcp` tool, whenever the user needs to use the tool we need to let the LLM know about the current active mcp servers to be called upon. Without the system prompt per tool if we need additional information specific for a tool we might need to add this additional information to the global system prompt. Currently we have to do some conditional logic inside the strategies.chat.system_prompt to check if the chat has mcp tool and if so add this additional information. Users may have their own custom prompts where they need to add this logic too in order for the mcp tool to work. If we can keep the system_prompt per tool as it is before, we can dynamically add additional info there. And also if we can evaluate the type = "system" messages from the messages array everytime we do chat:submit() this helps in for e.g in mcphub's case to let the user dynamically include or exclude necessary mcp servers and tools themselves so LLM will always have the currently active servers. Currently when the user does `@mcp do something` the servers active at the time are being stored in the system_prompt. 

Are there any issues with keeping the architecture this way? I know you have to consider wider use cases. Eager to hear your thoughts on this. 

Thank you

olimorris:
> Are there any issues with keeping the architecture this way? I know you have to consider wider use cases. Eager to hear your thoughts on this.

@ravitemer - none at all. I'll revert those changes back later today. For you and @Davidyz, I'm hoping the only changes will be on the schema side of things and I'm keen to keep it that way.

As I mentioned yesterday, this PR will not be merged until you both are happy with it.

ravitemer:
Thank you 😃.

Davidyz:
From my side, I'd also prefer to have a system prompt (whatever it's called, as long as there's a way for me to write instructions to teach the LLM how to call the tool). Otherwise, I'm quite happy with the existing implementation, so if there are no significant changes in the interface, it should be easy for me to keep up with.

To be completely honest, I think the VectorCode codecompanion tool should eventually be deprecated in favour of MCP. I've been receiving feedback from CopilotChat devs, and it looks like they're doing their own tool stuff. It'll be a lot easier if I could just maintain one tool implementation instead of multiple ones for each AI plugin. The MCP server is not quite ready though, so I'll still keep maintaining the tool for codecompanion and copilotchat for the time being.

lua/codecompanion/types.lua
```diff
@@ -108,7 +108,7 @@
 ---@field name string The name of the tool
 ---@field cmds table The commands to execute
 ---@field schema table The schema that the LLM must use in its response to execute a tool
----@field system_prompt fun(schema: table, xml2lua: table): string The system prompt to the LLM explaining the tool and the schema
+---@field system_prompt string The system prompt to the LLM explaining the tool and the schema
```
ravitemer, line 111: Can we keep this as a function? @olimorris

olimorris: Do you need it to be? The breaking change I am intending is for the schema to be JSON string, so not sure the system prompt would have any parameters therefore, doesn't need to be a function. But happy to change this.

ravitemer: My only reason is: when we declare like the following 

```lua
chat = {
  tools = {
    name = "mcp",
    --callback can also be a table, function, string that points to a lua file
    callback = {
       --other tool schema
      system_prompt = hub:get_active_servers_prompt() 
    }
  }
}
```
where tool.callback is a table, the system_prompt will be evaluated at the codecompanion.setup() by the time which the hub might not be ready leading to empty servers.  (same case when `callback = require("user.my_tool")`)

If the callback is a function or string,  both will be resolving into a table when `@mcp` tool is included in the prompt and resolved on chat:submit() at which point system_prompt is will be generated at that time. (This works fine but this might cause issues when we want to reevaluate role = system messages before every chat:submit() plus user may need to create another function somewhere else in the file inorder to first get the hub, check if it is not nil, then return prompt etc

With system_prompt being a function, the prompt will be generated dynamically, logic stays in the function.

S1M0N38:
A common pattern in LLM-related frameworks, when implementing function calling, is to leverage the type annotation to automatically parse the code of a function into the JSON representation of the function call.

For example:
- `@agent.tool` [pydantic-ai](https://ai.pydantic.dev/tools/)
- `@tool` in [smolagent](https://huggingface.co/docs/smolagents/tutorials/building_good_agents)
- `@mcp.tool` in [mcp python-sdk](https://github.com/modelcontextprotocol/python-sdk?tab=readme-ov-file#quickstart)

With the high level of code introspection offered by Python with the syntactic sugar offered by function decorators, this translates to the "@tool" decorator expected to be applied to a typed annotated function.

---

I wonder if something similar is achievable in Lua through Lua-language-server (which is already extensively used in this project judging by the Lua CATS scatter in the code base).

This would be considered simply a parser from Lua code to JSON/Lua-table LLM tool API.
```mermaid
flowchart LR
 subgraph s1["code2tool"]
    n1["lua function<br>with string annotation<br>(source code)"]
    n2["OpenAI tool repr. <br>(json or lua table)"] 
    n5["Anthropic tool repr. <br>(json or lua table)"] 
    n6["Other tool repr. <br>(json or lua table)"] 
    n1 --> n2
    n1 --> n5
    n1 --> n6
  end
    n2 -- opeai adapter --> n7
    n5 -- anthropic adapter --> n7
    n6 -- other adapter --> n7
    n7["codecompanion.nvim"]
```

olimorris:
@S1M0N38 thanks for sharing those links.

My plan will be to get OpenAI working first, executing the tools as expected, ensure i've got test coverage etc. Then I'll investigate the other adapters. I'm hoping that I can have a transformation class that can map between the adapters although Gemini's looks suitably different that this may be troublesome. I'll hopefully know early next week.

olimorris:
And it's rapid:

https://github.com/user-attachments/assets/3ff3b85a-23ac-4489-8a41-cf0f5d68a5f9

Note that I'm just returning a string in the tool, but not having to write XML in the buffer and parse it seems to make a huge difference.

olimorris:
✅ OpenAI is now complete

olimorris:
Function calling for all major adapters has now been added.

I'm now going to modify the existing tools and then I'll be running thorough integration tests across all of the adapters and their various models asking them to create the snake demo from the README as well as solve a coding challenge using an agentic workflow.

ravitemer:
@olimorris , Thank you!

I found this weird bug while updating mcp tool to use function calling. I replicated it with the weather tool.

```lua
  output = {
    ---@param agent CodeCompanion.Agent
    ---@param cmd table The command that was executed
    ---@param stdout table
    success = function(agent, cmd, stdout)
      return agent.chat:add_buf_message({
        role = "user",
        --INFO: This works
        -- content = string.format([[%s]], stdout[#stdout] .. ". "),

        --FIX: This renders the references after the message, and on next chat:submit() the tool seems to be gone
        --from it's memory as well as the system prompt of this tool
        content = string.format(
          [[
        %s
]],
          stdout[#stdout] .. ". "
        ),
      })
    end,
  },

```

For some reason, for multiline content, the references are being added after the content. And on next message the weather tool's system prompt and the weather tool itself seems to be gone from the chat history. 

image

olimorris:
@ravitemer you're further ahead with testing than I am so thanks for flagging this.

I'll try and recreate in my upcoming tests.

Davidyz:
Putting this here in case anyone's interested: https://github.com/Davidyz/VectorCode/pull/59

btw, I noticed that the LLM (deepseek-v3-0324) fails to produce the correct tool call schema until I added examples in the system prompt. This could be unique to VectorCode but might be worth mentioning in the docs.

ravitemer:
It is working great now. Although `gpt-4o-2024-11-20` seems to get confused with `uri` vs `url` and keeps on passing `url`  like this for a simple fetch tool. For others servers it is working fine.
```txt
[DEBUG] 2025-04-04 16:38:07
Args: {
  action = "use_mcp_tool",
  server_name = "fetch",
  tool_name = "fetch",
  url = "example.com"
}
```

<details>
<summary>function schema</summary>

```lua
    schema = {
        type = "function",
        ["function"] = {
            name = "mcp",
            description = "The Model Context Protocol (MCP) enables communication with locally running MCP servers that provide additional tools and resources to extend your capabilities. This tool calls mcp tools and resources on the mcp servers using `use_mcp_tool` and `access_mcp_resource` actions respectively.",
            parameters = {
                type = "object",
                properties = {
                    action = {
                        description = "Action to perform: one of `access_mcp_resource` or `use_mcp_tool`. Must be provided always.",
                        type = "string",
                        enum = {
                            "use_mcp_tool",
                            "access_mcp_resource",
                        },
                    },
                    server_name = {
                        description = "Name of the on the available MCP servers. Must be provided always.",
                        type = "string",
                    },
                    tool_name = {
                        description = "Name of the tool from server_name's available tools to call while using `use_mcp_tool` action. Must be provided when action is `use_mcp_tool`",
                        type = "string",
                    },
                    arguments = {
                        description = "Arguments for the `use_mcp_tool` action based on tool_name's inputSchema. Must be provided when action is `use_mcp_tool`.",
                        type = "object",
                    },
                    uri = {
                        description = "URI of the resource or resourceTemplate to access when using `access_mcp_resource` action. Must be provided when action is `access_mcp_resource`.",
                        type = "string",
                    },
                },
                required = {
                    "action",
                    "server_name",
                },
                additionalProperties = false,
            },
            strict = true,
        },
    },
```

</details>

Given this schema, we can split this into two separate tools `use_mcp_tool` and `access_mcp_resource` rather than being greedy and combining them into a single function. This might solve issues with some models unable to generate proper schema. 
Thanks to tool groups we can combine these two into one `@mcp`

@olimorris what do you think?

ravitemer:
I know this is still in the works, but I found this bug:

When  `stream = false`  in the adapter (I'm using `copilot`) and using claude models, the text before the tool call is in `choices[1]` and the actual tool call is in `choices[2]` as given below which leads to the tool not being called. 


<details>

<summary>Formatted Response Log </summary>

```json
{
  "body": {
    "choices": [
      {
        "finish_reason": "tool_calls",
        "message": {
          "content": "I'll fetch example.com using the fetch tool:",
          "role": "assistant"
        }
      },
      {
        "finish_reason": "tool_calls",
        "message": {
          "role": "assistant",
          "tool_calls": [
            {
              "function": {
                "arguments": "{\"params\":{\"url\":\"https://example.com\"},\"server_name\":\"fetch\",\"tool_name\":\"fetch\"}",
                "name": "use_mcp_tool"
              },
              "id": "tooluse_Te5_iqfmTSOeq28W2eNTUg",
              "type": "function"
            }
          ]
        }
      }
    ],
    "created": 1743830102,
    "id": "0615dccb-24bc-474d-b333-cd26692103c0",
    "model": "Claude 3.5 Sonnet",
    "usage": {
      "completion_tokens": 115,
      "prompt_tokens": 3843,
      "total_tokens": 3958
    }
  },
  "status": 200
}
```

</details>

olimorris:
Thanks for sharing that. I'll add some specific Copilot tests.

With streaming on, it's working for you? Copilot with Claude is my daily driver and testing so far has been okay.

As an aside, for tool runs I need to return some visual indication that the tool is being generated. Previously the user could see the XML being generated but I intend to keep the tool JSON out of the buffer with this PR.

ravitemer:
Thank you! Yes, with streaming on, claude is working great. 

> I need to return some visual indication that the tool is being generated

This is great the way it is now. Waiting to see how you implement showing the tool call results in the buffer.

olimorris:
Fixed Copilot + function calls with streaming turned off and added full test coverage.

ravitemer:
> Fixed Copilot + function calls with streaming turned off and added full test coverage.

Thanks you @olimorris , there is a small bug:

Copilot is throwing this when used with stream = false when it calls a tool:

`Couldn't decode the tool arguments: bad argument #1 to '?' (string expected, got table)`

in `adapters.openai.lua` we need to remove the 

```lua
arguments = vim.json.decode(tool["function"]["arguments"]),
```

This works fine!

```lua
      ---Process a tool call in non-streaming mode
      local function process_nonstreamed_tool_call(tools_table, tool, index)
        local id = (tool.id and tool.id ~= "") and tostring(tool.id) or tostring(index)
        tools_table[id] = {
          name = tool["function"]["name"],
          arguments = tool["function"]["arguments"],
        }
      end
```

jacobdaddario:
Super excited to see the editor change land. It's shocking to realize just how much of the prompt was teaching the LLM how to use the home cooked tool usage. The prompt is so much shorter now!

Do you anticipate that this will increase accuracy since less input about topics unrelated to coding/the supplied buffers will be present now? Do you reckon it needs any other information on how to use the tool for code modifications, or are the LLMs just _that_ good at synthesizing their coding capabilities with the too usage?

Thanks for your amazing work on this stuff. I use it every day 🫡

olimorris:
> > Fixed Copilot + function calls with streaming turned off and added full test coverage.
> 
> Thanks you @olimorris , there is a small bug:
> 
> Copilot is throwing this when used with stream = false when it calls a tool:
> 
> `Couldn't decode the tool arguments: bad argument #1 to '?' (string expected, got table)`
> 

Can you share the log trace so I can add a test for this? Also what model are you using?

olimorris:
> Super excited to see the editor change land. It's shocking to realize just how much of the prompt was teaching the LLM how to use the home cooked tool usage. The prompt is so much shorter now!

The history here was that I implemented tools before function calling was really universal so it was the only way to ensure users had a semi unified way of using CodeCompanion + tools across various models. 

> Do you anticipate that this will increase accuracy since less input about topics unrelated to coding/the supplied buffers will be present now? Do you reckon it needs any other information on how to use the tool for code modifications, or are the LLMs just that good at synthesizing their coding capabilities with the too usage?

For lower parameter models that support function calling I expect it will be much more reliable. The only issues I ever had with the XML tools implementation were with the self-hosted models. I'm aware that some users massively pushed the limits of LLMs with the current implementation. So for those users, I expect they'll see much more reliable output.

You'll notice the speed improvements are tremendous though.

jacobdaddario:
> The history here was that I implemented tools before function calling was really universal so it was the only way to ensure users had a semi unified way of using CodeCompanion + tools across various models.

For sure, totally get why you moved to tools with the ubiquity of stuff like MCP emerging. I'm really glad you did it with XML though, at least for a time. It was cool to see that only prompt engineering was necessary to get LLMs to combine their coding ability with XML markup creation.

ravitemer:
I am using `claude-3.5-sonnet` but it gives the same result for any model. It occurs with the default weather tool as well

```md
##   Ubuntu

> Context:
> - <tool>weather</tool>

@weather in delhi?

##   Copilot (claude-3.5-sonnet)

I'll help you check the weather in Delhi. I'll use the weather function with Delhi as the location. Since no specific unit preference was mentioned, I'll use Celsius which is the standard unit used in India.

> Called the `weather` tool

##   Ubuntu

> Context:
> - <tool>weather</tool>
```

The error is at `agents/init.lua` : `function Agent:execute`: For non streamed tools we are passing a table unlike streamed ones. 

```lua
    self.tool.name = name
    if tool.arguments then
      self.tool.args = tool.arguments
      -- For some Copilot models, the args are a JSON string
      local ok, args = pcall(vim.json.decode, tool.arguments)
      if not ok then
        return log:error("Couldn't decode the tool arguments: %s", args)
      end
      self.tool.args = args
    end
```

This is my extended copilot adapter

```lua
local constants = require("user.codecompanion.constants")

return {
	opts = {
		stream = false,
	},
	schema = {
		model = {
			default = vim.g[constants.CC_MODEL] or "claude-3.7-sonnet",
		},
	},
}
```

<details>
<summary> Logs ( just changed sensitive info in curl request log)</summary>

```txt
[TRACE] 2025-04-09 03:27:00
Chat created with ID 7357391
[TRACE] 2025-04-09 03:27:00
Chat opened with ID 7357391
[TRACE] 2025-04-09 03:27:00
Setting the header for the chat buffer
[DEBUG] 2025-04-09 03:27:12
[Tools] strategies.chat.agents.tools.weather identified
[TRACE] 2025-04-09 03:27:12
Settings:
{
  env = {
    api_key = <function 1>
  },
  features = {
    text = true,
    tokens = true,
    vision = false
  },
  formatted_name = "Copilot",
  handlers = {
    chat_output = <function 2>,
    form_messages = <function 3>,
    form_parameters = <function 4>,
    form_tools = <function 5>,
    inline_output = <function 6>,
    on_exit = <function 7>,
    setup = <function 8>,
    tokens = <function 9>
  },
  headers = {
    Authorization = "Bearer ${api_key}",
    ["Content-Type"] = "application/json",
    ["Copilot-Integration-Id"] = "vscode-chat",
    ["Editor-Version"] = "Neovim/0.10.4"
  },
  name = "copilot",
  opts = {
    stream = false,
    tools = true
  },
  parameters = {
    max_tokens = 15000,
    model = "claude-3.5-sonnet",
    n = 1,
    reasoning_effort = "medium",
    temperature = 0,
    top_p = 1
  },
  roles = {
    llm = "assistant",
    user = "user"
  },
  schema = {
    max_tokens = {
      default = 15000,
      desc = "The maximum number of tokens to generate in the chat completion. The total length of input tokens and generated tokens is limited by the model's context length.",
      mapping = "parameters",
      order = 4,
      type = "integer"
    },
    model = {
      choices = <function 10>,
      default = "claude-3.5-sonnet",
      desc = "ID of the model to use. See the model endpoint compatibility table for details on which models work with the Chat API.",
      mapping = "parameters",
      order = 1,
      type = "enum"
    },
    n = {
      condition = <function 11>,
      default = 1,
      desc = "How many chat completions to generate for each prompt.",
      mapping = "parameters",
      order = 6,
      type = "number"
    },
    reasoning_effort = {
      choices = { "high", "medium", "low" },
      default = "medium",
      desc = "Constrains effort on reasoning for reasoning models. Reducing reasoning effort can result in faster responses and fewer tokens used on reasoning in a response.",
      mapping = "parameters",
      optional = true,
      order = 2,
      type = "string"
    },
    temperature = {
      condition = <function 12>,
      default = 0,
      desc = "What sampling temperature to use, between 0 and 2. Higher values like 0.8 will make the output more random, while lower values like 0.2 will make it more focused and deterministic. We generally recommend altering this or top_p but not both.",
      mapping = "parameters",
      order = 3,
      type = "number"
    },
    top_p = {
      condition = <function 13>,
      default = 1,
      desc = "An alternative to sampling with temperature, called nucleus sampling, where the model considers the results of the tokens with top_p probability mass. So 0.1 means only the tokens comprising the top 10% probability mass are considered. We generally recommend altering this or temperature but not both.",
      mapping = "parameters",
      order = 5,
      type = "number"
    }
  },
  url = "https://api.githubcopilot.com/chat/completions",
  <metatable> = {
    __index = {
      extend = <function 14>,
      get_env_vars = <function 15>,
      make_from_schema = <function 16>,
      make_safe = <function 17>,
      map_roles = <function 18>,
      map_schema_to_params = <function 19>,
      new = <function 20>,
      resolve = <function 21>,
      set_env_vars = <function 22>
    }
  }
}
[TRACE] 2025-04-09 03:27:12
Messages:
{ {
    content = "You are an AI programming assistant named \"CodeCompanion\". You are currently plugged into the Neovim text editor on a user's machine.\n\nYour core tasks include:\n- Answering general programming questions.\n- Explaining how the code in a Neovim buffer works.\n- Reviewing the selected code in a Neovim buffer.\n- Generating unit tests for the selected code.\n- Proposing fixes for problems in the selected code.\n- Scaffolding code for a new workspace.\n- Finding relevant code to the user's query.\n- Proposing fixes for test failures.\n- Answering questions about Neovim.\n- Running tools.\n\nYou must:\n- Follow the user's requirements carefully and to the letter.\n- Keep your answers short and impersonal, especially if the user's context is outside your core tasks.\n- Minimize additional prose unless clarification is needed.\n- Use Markdown formatting in your answers.\n- Include the programming language name at the start of each Markdown code block.\n- Avoid including line numbers in code blocks.\n- Avoid wrapping the whole response in triple backticks.\n- Only return code that's directly relevant to the task at hand. You may omit code that isn’t necessary for the solution.\n- Avoid using H1 and H2 headers in your responses.\n- Use actual line breaks in your responses; only use \"\\n\" when you want a literal backslash followed by 'n'.\n- All non-code text responses must be written in the English language indicated.\n\nWhen given a task:\n1. Think step-by-step and, unless the user requests otherwise or the task is very simple, describe your plan in detailed pseudocode.\n2. Output the final code in a single code block, ensuring that only relevant code is included.\n3. End your response with a short suggestion for the next user turn that directly supports continuing the conversation.\n4. Provide exactly one complete reply per conversation turn.",
    cycle = 1,
    id = 1999425959,
    opts = {
      tag = "from_config",
      visible = false
    },
    role = "system"
  }, {
    content = "weather in delhi?",
    cycle = 1,
    id = 1084998870,
    opts = {
      visible = true
    },
    role = "user"
  }, {
    content = "The weather tool must only be called once for each location",
    cycle = 1,
    id = 1784457729,
    opts = {
      reference = "<tool>weather</tool>",
      tag = "tool",
      visible = false
    },
    role = "system"
  } }
[INFO] 2025-04-09 03:27:12
Chat request started
[DEBUG] 2025-04-09 03:27:12
Authorizing GitHub Copilot token
[TRACE] 2025-04-09 03:27:14
Reusing GitHub Copilot token
[TRACE] 2025-04-09 03:27:14
Reusing GitHub Copilot token
[INFO] 2025-04-09 03:27:14
Request body file: /tmp/lvim.ubuntu/p1vnuI/2.json
[DEBUG] 2025-04-09 03:27:14
Request:
{ "-sSL", "-D", "/run/user/1000//plenary_curl_097f4542.headers", "--compressed", "-X", "POST", "-H", "Copilot-Integration-Id: vscode-chat", "-H", "Authorization: Bearer tid=xxxx;exp=1744171032;sku=monthly_subscriber;proxy-ep=proxy.individual.githubcopilot.com;st=dotcom;chat=1;cit=1;malfil=1;editor_preview_features=1;ccr=1;rt=1;8kp=1;ip=xxxx;asn=xxxx", "-H", "Editor-Version: Neovim/0.10.4", "-H", "Content-Type: application/json", "-d", "@/tmp/lvim.ubuntu/p1vnuI/2.json", "--retry", "3", "--retry-delay", "1", "--keepalive-time", "60", "--connect-timeout", "10", "https://api.githubcopilot.com/chat/completions" }
[TRACE] 2025-04-09 03:27:19
Output data:
{
  body = "{\"choices\":[{\"finish_reason\":\"tool_calls\",\"message\":{\"content\":\"I'll help you check the weather in Delhi. I'll use the weather function with Delhi as the location. Since no specific unit preference was mentioned, I'll use Celsius which is the standard unit used in India.\",\"role\":\"assistant\"}},{\"finish_reason\":\"tool_calls\",\"message\":{\"role\":\"assistant\",\"tool_calls\":[{\"function\":{\"arguments\":\"{\\\"location\\\":\\\"Delhi, India\\\",\\\"units\\\":\\\"celsius\\\"}\",\"name\":\"weather\"},\"id\":\"tooluse_urgEDBnoQyeJnMgn3lAt5w\",\"type\":\"function\"}]}}],\"created\":1744169239,\"id\":\"1c7b8d8c-28e4-44d7-8795-2251be5328cb\",\"model\":\"Claude 3.5 Sonnet\",\"usage\":{\"completion_tokens\":115,\"prompt_tokens\":894,\"total_tokens\":1009}}",
  exit = 0,
  headers = { "content-security-policy: default-src 'none'; sandbox", "content-type: application/json", "strict-transport-security: max-age=31536000", "x-request-id: 00000-3c9e966a-a41e-4f10-9419-9934a8056186", "date: Wed, 09 Apr 2025 03:27:19 GMT", "content-length: 696", "x-github-backend: Kubernetes", "x-github-request-id: C2BC:22125E:63C497:868DFE:67F5E914" },
  status = 200
}
[TRACE] 2025-04-09 03:27:19
Tokens: 1009
[DEBUG] 2025-04-09 03:27:19
[Tools] strategies.chat.agents.tools.weather identified
[ERROR] 2025-04-09 03:27:21
Couldn't decode the tool arguments: bad argument #1 to '?' (string expected, got table)
[DEBUG] 2025-04-09 03:27:21
Executor:execute - Queue empty
[INFO] 2025-04-09 03:27:21
Chat request finished

```
</details>

ujisati:
Codecompanion is beautiful and I'm so thankful you created this because I almost started using VSCode for LLM extensions.

Just want to ask about this:

> btw, I noticed that the LLM (deepseek-v3-0324) fails to produce the correct tool call schema until I added examples in the system prompt. This could be unique to VectorCode but might be worth mentioning in the docs.

I haven't tried adding examples to the system prompt, but I'm noticing deepseek will not attempt using tool calls at all. E.g.

```
{
    content = "weather LA",
    cycle = 1,
    id = 1016520905,
    opts = {
      visible = true
    },
    role = "user"
  }, {
    content = "The weather tool must only be called once for each location",
    cycle = 1,
    id = 1784457729,
    opts = {
      reference = "<tool>weather</tool>",
      tag = "tool",
      visible = false
    },
    role = "system"
  }, {
    content = "I don't have direct access to real-time weather data or external APIs. \n\nTo get weather data in Neovim, you could:\n1. Use a weather API wrapper in your preferred language\n2. Create a simple script that fetches data from a weather service\n3. Use an existing plugin like `voldikss/vim-weather`\n\nWould you like me to suggest a specific weather API integration approach?",
    cycle = 1,
    id = 141190074,
    opts = {
      visible = true
    },
    role = "llm"
  } }
```
It sounds like its just a prompt issue, but I'm wondering if the direction for CodeCompanion is to have a minimal prompt, and for users to customize as needed? It would be awesome if most SOTA models that support tools worked out-of-the-box, with community contributing best prompts per model. But again, thank you for this plugin.

Davidyz:
> I haven't tried adding examples to the system prompt, but I'm noticing deepseek will not attempt using tool calls at all. E.g.

Were you able to reproduce this on other tools? The weather tool looks like a placeholder for the development of this PR and doesn't do much. Seeing the changes to the system prompts for the `cmd_runner` and `editor` tools, I think it's fair to say that the system prompt should still contain detailed instructions on how the tool should be used.

olimorris:
For those that follow this PR, one of the nuisances in recent days has been feeding the tool call and the tool output back into the message history. Finding a clean approach to do that - given OpenAI and Anthropic do this differently - has required some thinking as this needs to be suitable for other LLMs that may do it even more differently.

olimorris:
...there's light at the end of the tunnel...

Latest PR adds the ability (in less than 100 LOC) to test the JSON payload of all of the adapters after they've sent a function call request, received a response back and added the output from the function call back into the chat buffer.

olimorris:
@ravitemer @Davidyz - There's now a helper function, `codecompanion.has`, that will enable you to determine if a user's codecompanion version has the old xml tools or function calling:

```lua
local codecompanion = require("codecompanion")

codecompanion.has("xml-tools") -- true|false
codecompanion.has("function-calling") -- true|false
codecompanion.has() -- returns a table of supported features
```

For some users, I'm expecting they may want to stay on an older version of the plugin for a bit. I'm having some _very_ mixed results with function calling with Llama 3.1.

This has been added to the `main` branch as well.

ravitemer:
Awesome! Thank you. I was about to ask this sort of feature so that I can provide xml and function versions for mcp tool.

> For some users, I'm expecting they may want to stay on an older version of the plugin for a bit. I'm having some very mixed results with function calling with Llama 3.1.

Yes, I feel many users who doesn't use 3.5, 3.7, gtp 4o etc will stick to xml version. As far as I know codecompanion is the only one that supports xml based tools which just works with any kind of model although results may vary. Function calling will ensure that the tool calls will be reliable with supported models. 

Just a thought, that might help very small number of users and not that urgent, so don't take this seriously:
I understand that this might be too much and too costly to maintain, but I feel like an ideal case would be having a global variable or adapter specific option to determine which kind of tool to use might be super useful 🤭 just like how currently we can tweak the system prompt based on model or adapter. This can be toggled with some custom keymap instead of having to switch between two different branches , uninstalling and installing the plugin.

olimorris:
> but I feel like an ideal case would be having a global variable or adapter specific option to determine which kind of tool to use might be super useful

Love that idea and I would love the community to pick something like that up. I plan on doing a couple of videos using VectorCode and MCPHub to show people how to contribute to the plugin or squash some bugs.

 In an ideal world, I would have supported both XML tools and function calling.

ravitemer:
In case anyone wants to test function tools with mcphub, I have opened at Draft PR at https://github.com/ravitemer/mcphub.nvim/pull/94 . It supports both types of tools using `codecompanion.has()`. See there for more information.

olimorris:
I shall be moving on to testing now Anthropic is [closed](https://github.com/olimorris/codecompanion.nvim/pull/1141/commits/6a7db129fe053ae5056b5372e8ae9fdcf4212e25). That was a challenge...Claude 3.7 choked. Gemini 2.5 Pro struggled so I had to do 90% of the Anthropic function calling implementation myself...o4-mini came through massively towards the end.

All adapters are covered by integration tests (as close as you can _really_ get in Neovim) that test submitting content in the chat buffer through to checking that the JSON payload that would be sent to the LLM is correct.

There's some weird issue with the Ollama adapter that I can't workout...it doesn't seem to recognise the tool output...I'll sort that later on.

But...would very much welcome everyone playing around with the `@weather` tool to check everything is working with their setups.

olimorris:
So it [appears](https://github.com/ollama/ollama/issues/9632) function calling in Ollama is pretty broken. I'd tried using function calling with the non-OpenAI endpoint and Ollama just continuously called the same tool in every response no matter what I tried.

So I moved to the OpenAI endpoint (I want to make sure all adapters, where possible, adopt this) and function calling is just plain broken.

For this PR, Ollama will not be supported 😢.

olimorris:
2025-04-21 22_17_52 - WezTerm@2x

Anyone watching this PR, would appreciate you testing the editor tool out. There's some quirks with the diff occasionally - some changes aren't captured but the screenshot shows how it should look.

olimorris:
Agentic workflows are working pretty nicely with function calls. GPT-4.1 solving a coding challenge in < 15 seconds with me pressing submit once:

https://github.com/user-attachments/assets/2144b254-0135-4a6b-a5c1-927d672c110c

Work has been super busy recently which is why this PR has taken so long. But agentic workflows are an awesome way of testing all of the adapter tool implementations so I shall persevere with these before fixing the files tool and running some Snake demos.

jacobdaddario:
Swapped to this branch, and o3-mini was reporting schema issues for the tools. I think that they reduced the cost of o3-mini and made 4.1 cost roughly the same as o3-mini did. I swapped to 4.1 to see if it fixed my issues. Holy cow, it's so fast. Not sure if it's the change of model, use of native tools, or both.

On top of that, outputs seem to be more reliable too. I probably don't need to use a reasoning model anymore since I don't need it to ensure it gets the XML formatting right (4o would bomb sometimes).

All in all, seems like between the enhanced reliability from native tools and the smaller context being sent to the LLM, things are coming out a lost faster.

olimorris:
> Swapped to this branch, and o3-mini was reporting schema issues for the tools

Thanks for sharing. I'll look into that. I've tested with GPT-4o, GPT-4.1 and o4-mini and all working well although the latter doesn't seem to like running tools in parallel via Copilot.

There's a bug in the Anthropic adapter I need to fix as well. I need to ensure messages with the same `tool_use_id` are merged.

> Holy cow, it's so fast. Not sure if it's the change of model, use of native tools, or both.

A combination of both but function calling is much faster than the XML version I was using previously.

amadeus:
I have noticed a bug, not sure if it's related to this branch or not, but I've been on this branch so figured I'd report it, where if I close out the chat buffer with a specialized command I have, I can't re-summon it. Attached a video of the issue, and also the code I have for the wipeout comment:

https://github.com/user-attachments/assets/3ff72b05-05f2-424d-9662-7663c43270f6

```lua
local function wipeout_all_buffers()
  vim.cmd("%bwipeout")
end

local function wipeout_buffers()
  local current_buf = vim.api.nvim_get_current_buf()
  local all_bufs = vim.api.nvim_list_bufs()
  for _, buf in ipairs(all_bufs) do
    -- Skip current buffer only if it's not startify
    if buf ~= current_buf or vim.bo[current_buf].filetype == "startify" then
      -- Check if the buffer is valid before attempting to wipe it
      if vim.api.nvim_buf_is_valid(buf) then
        -- Use pcall to catch any errors that might occur when trying to wipe a buffer
        pcall(function()
          vim.cmd("bwipeout! " .. buf)
        end)
      end
    end
  end
end

vim.api.nvim_create_user_command("Wipeout", wipeout_buffers, {})
vim.api.nvim_create_user_command("WipeoutAll", wipeout_all_buffers, {})

return {}
```

olimorris:
@amadeus lets keep comments related to the purpose of the PR please. This should be in the discussions not in a PR related to function calling.

olimorris:
## Model Testing

I've ran the `Edit<->Test` workflow where we ask an LLM to solve a coding challenge using the `@editor` and `@cmd_runner` tools. I've ran this across every model for the supported adapters, resolving a few issues as I've gone.

The results below indicate which models are working with this PR, but not necessarily solving the challenge. I'm testing functionality rather than testing the model's capability.

## Working Models

### Anthropic

- [x] claude-3-7-sonnet-20250219
- [x] claude-3-5-sonnet-20241022
- [x] claude-3-5-haiku-20241022
- [x] claude-3-opus-20240229 (this model is still awesome!)

### Copilot

- [x] gpt-4o
- [x] gpt-4.1
- [x] o1
- [x] o3-mini
- [x] o4-mini
- [x] claude-3-5-sonnet
- [x] claude-3.7-sonnet

### DeepSeek

- [x] DeepSeek Chat

### Google

- [x] Gemini-2.0-flash
- [x] Gemini-2.5-pro-exp-03-25

### OpenAI

- [x] gpt-3.5-turbo
- [x] gpt-4
- [x] gpt-4o
- [x] gpt-4o-mini
- [x] o1-2024-12-17
- [x] o3-mini-2025-01-31

## Non-Working Models

### Copilot

* claude-3.7-sonnet-thought (doesn't support tools)
* gemini-2.5-pro
* gemini-2.0-flash-001

### OpenAI

* o1-mini-2024-09-12 (doesn't support tools)

## Next Steps

@ravitemer and @Davidyz - With these tests I'm happy with where this PR is at. I just need to write/update the docs, convert the `@files` tool and do the final tests with that. I don't expect any more changes to the wider plugin either.

When you both give me the green light, I'll merge this PR.

olimorris:
@jacobdaddario - The o3-mini bugs have been fixed. Weird quirk in the OpenAI API whereby tools were working fine for all other models except o3-mini

Davidyz:
Exciting progress here! I've been trying this with [VectorCode](https://github.com/Davidyz/VectorCode/pull/59) on Gemini (2.0 pro) and OpenRouter (openai compatible backend, deepseek-v3-0324:free). Gemini 2.0 pro works great, but the free deepseek model won't call the tools (either reply with no tool call, or the tool call is just a json code block in the answer, unlike Gemini where there's actually a key in the message (in the debug chat window) called `tool_calls` and the values are the call arguments. This could be just an issue with the model (or provider, because openai-compatible API is probably as fragmented as Linux distros), and I'll keep experimenting.

olimorris:
If that's all you're seeing in the chat history then openrouter/DeepSeek is the problem. In the logs you should be seeing something akin to:

```
[{"index": 0, "id": "call_DdmO9pD3xa9XTPNJ32zg2hcA", "function": {"arguments": "", "name": "get_weather"}, "type": "function"}]
[{"index": 0, "id": null, "function": {"arguments": "{\"", "name": null}, "type": null}]
[{"index": 0, "id": null, "function": {"arguments": "location", "name": null}, "type": null}]
[{"index": 0, "id": null, "function": {"arguments": "\":\"", "name": null}, "type": null}]
[{"index": 0, "id": null, "function": {"arguments": "London", "name": null}, "type": null}]
[{"index": 0, "id": null, "function": {"arguments": ",", "name": null}, "type": null}]
[{"index": 0, "id": null, "function": {"arguments": " UK", "name": null}, "type": null}]
[{"index": 0, "id": null, "function": {"arguments": "\"}", "name": null}, "type": null}]
```

DeepSeek-Reasoner didn't work with function calling last time I checked their docs.

Davidyz:
I see. I'll try some other providers just to be safe, but since Gemini's working I'd say the migration is mostly done. I'll just have to copy-paste the old implementation for compatibility (with the `has` function).

Davidyz:
Tried gemini 2.0 flash [on openrouter](https://openrouter.ai/google/gemini-2.0-flash-exp:free) and it fails to give actual tool calls (gemini 2.0 flash works when using the Gemini adapter, but not when using the openai-compatible adapter through openrouter). Also tried official Qwen openai-compatible API and it doesn't work either (doesn't trigger the tool call).

Also one annoying thing with the official qwen API: THEY DON'T SUPPORT `/v1/models`

olimorris:
I suspect they're not passing on the tools parameter block. Copilot and Gemini isn't working either although that's for other reasons I'm yet to figure out.

Davidyz:
Yes it's weird because initially I noticed the free deepseek models on oepnrouter don't have the `tool` tag, so I thought that's the reason. But then I tried Gemini through oepnrouter, which DOES have the `tool` tag, and it still doesn't work.

Davidyz:
Would it work if we extract JSON codeblocks from the response for providers that fails to reply with `tool_calls`? It'll be similar to the old xml-based approach, but _maybe_ will work for providers that don't properly handle tools via the api?

ujisati:
> Would it work if we extract JSON codeblocks from the response for providers that fails to reply with `tool_calls`? It'll be similar to the old xml-based approach, but _maybe_ will work for providers that don't properly handle tools via the api?

@Davidyz It sounds like gemini on openrouter is working now, when I tested a few days ago with gemini 2.5 on openrouter it didn't work, but as far as falling back to a different tool use strategy, I was looking at how https://github.com/RooVetGit/Roo-Code handles tools for openrouter, and it looks like all of their tool requests use custom XML prompts and parsing. I'm not if sure its technically anthropic format I'm not familiar with it, but their openrouter provider doesn't use OpenAI tool calls at all. e.g.

src/core/prompts/tools/write-to-file.ts
src/api/providers/openrouter.ts
src/core/Cline.ts

I think @olimorris said that support for XML handling will come from the community now on a different branch. I'm not sure the reason to uses JSON instead. If Gemini 2.5 is working through the Gemini provider then I may just use that but I haven't tested it.

olimorris:
Just the creating tools guide to update now.

As a heads up, Copilot and Anthropic are super stubborn when it comes to running parallel tools. I'll be opening this up to the community for help.

olimorris:
Docs updated. Video re-recorded and edit. This PR is complete and once #1290 is ready then we'll merge.

@Davidyz - You happy from a VectorCode perspective?

Davidyz:
Yep and I've merged the `has` check for compatibility. Still need to move some things around for the tool calling branch, but nothing major (I mean the issue with some providers not doing tool use properly persists but that's their fault). I'm happy where we are about this PR.

Davidyz:
I've done the PR for the function-calling VectorCode tool and I'm looking into openai-compatible adapters, and it looks like there's no `tools` field in the request body. The tool is correctly picked up in the chat:
```markdown
## Me

> Context:
> - <tool>vectorcode</tool>

@vectorcode reranker implementation in this project?
```
but in the request body there's no `tools` field
```json
{
	"model":"Qwen/Qwen2.5-72B-Instruct-128K",
	"stream_options":{"include_usage":true},
	"stream":true,
	"messages":[
		// some messages here
	]
}
```

olimorris:
> I've done the PR for the function-calling VectorCode tool and I'm looking into openai-compatible adapters, and it looks like there's no `tools` field in the request body. The tool is correctly picked up in the chat:

Should be good now

Davidyz:
Sadly no, but the following patch from gemini did the trick:
```
diff --git a/lua/codecompanion/adapters/openai_compatible.lua b/lua/codecompanion/adapters/openai_compatible.lua
index ce57f2f..9a12bb1 100644
--- a/lua/codecompanion/adapters/openai_compatible.lua
+++ b/lua/codecompanion/adapters/openai_compatible.lua
@@ -123,6 +123,9 @@ return {
     form_messages = function(self, messages)
       return openai.handlers.form_messages(self, messages)
     end,
+    form_tools = function(self, tools)
+      return openai.handlers.form_tools(self, tools)
+    end,
     chat_output = function(self, data, tools)
       return openai.handlers.chat_output(self, data, tools)
     end,
```
with this patch the gemini 2.0 flash exp on openrouter works flawlessly.
I also tried with qwen2.5 72b from a different openai-compatible provider, and the wrong tool call format froze the chat (`Couldn't decode the tool arguments` error).

olimorris:
Ah so dumb of me. Included that in the latest commit.

> I also tried with qwen2.5 72b from a different openai-compatible provider, and the wrong tool call format froze the chat

So the provider isn't OpenAI-compatible when it comes to tools? Or is there an issue in how we're putting the tool calls back in the messages request?

Davidyz:
> So the provider isn't OpenAI-compatible when it comes to tools? Or is there an issue in how we're putting the tool calls back in the messages request?

```
[ERROR] 2025-04-28 17:16:24
Couldn't decode the tool arguments:  {"command": "query", "options": {"query": ["vectorcode", "reranker", "implementation"], "count": 15, "project_root": "/home/davidyz/git/VectorCode"}}}
```
There was an extra `}` in the response, which broke the json syntax. I remember in the xml-based tools when there's a decoding error the error message will just be added to the chat (prob. invisible?) so that the model knows it's wrong.

olimorris:
> There was an extra } in the response, which broke the json syntax. I remember in the xml-based tools when there's a decoding error the error message will just be added to the chat (prob. invisible?) so that the model knows it's wrong.

Okay and when you say "froze the chat", I assume the chat buffer wasn't unlocked, ready for you to add another user prompt?

I must have run 100+ tests now and I haven't had a model produce incorrect JSON (I thought they were fine-tuned to be able to call functions ffs 😆). So, I'll make sure we let the LLM know of the error in the chat.

lua/codecompanion/strategies/chat/init.lua
```diff
@@ -1096,9 +1138,40 @@ function Chat:add_buf_message(data, opts)
...
+
+  local existing = find_tool_call(tool_call.id, self.messages)
+  if existing then
+    existing.content = existing.content .. "\n\n" .. output.content
+  else
+    table.insert(self.messages, output)
+  end
+
+  -- Add a notification to the UI
+  for_user = for_user or for_llm
```
Davidyz, line 1162: Can we have a mechanism for disabling the `for_user` output? Like if VectorCode is returning 30 files, I wouldn't want all of them to be shown in the buffer. I'm hoping to (optionally) have a single `for_user` message that says `VectorCode added 30 files to the chat`. I tried passing an empty string for `for_user` and it inserted a bunch of blank lines. maybe just skip the `add_buf_message` call if `for_user` is empty string (`""`)?

Or rather, should I just concat all files into one single string and call `add_tool_output` once?

olimorris: This makes a lot of sense, I'll add in later.

You could always, update the chat buffer messages in the `on_exit` handler. I added in an option in this PR for `use_handlers_once` (see the editor tool). This peeks at the next item in the queue to see if it's the same tool. If it is, then it won't call `handlers.setup` and will only call `handlers.on_exit` if it's the last item in the queue or the items is a different tool altogether.

olimorris: `for_user` can now be `""`. Also updated the docs for this.

olimorris:
> > Okay and when you say "froze the chat", I assume the chat buffer wasn't unlocked, ready for you to add another user prompt?
> 
> It's stuck here:
> 
> ```
> ## Me
> 
> > Context:
> > - <tool>vectorcode</tool>
> 
> @vectorcode reranker implementation?
> 
> ## CodeCompanion (OpenAI Compatible)
> 
> To find the implementation of the reranker in the VectorCode project, I will perform a semantic search with the following keywords: "reranker implementation". This should retrieve files that are relevant to the reranker functionality.
> 
> <tool_call>
> ```
> 
> with the buffer lock and no `## Me` header for me to type under. Funny enough, I was able to consistently reproduce this with qwen2.5-72b from this provider (siliconflow). I'll try this with some other providers/models and make some final adjustments to the system prompt, and then I'll merge the updated vectorcode tool.

I've addressed this now. If the JSON can't be decoded then an error's thrown, the chat buffer's updated and the lock is released.

jacobdaddario:
🚀

PR #1145, opened 2025-03-18
Add caching to adapters that make http calls to model endpoints

olimorris:
## Description

We have a number of adapters that make http calls to endpoints to fetch the available models. Currently, to reduce latency, this is done once per session. However, providers may update the model list during a user's session so it makes sense to enable the user to fetch these. The compromise is a 30 min cache.

### DIFF
--- a/lua/codecompanion/adapters/copilot.lua
+++ b/lua/codecompanion/adapters/copilot.lua
@@ -188,7 +188,7 @@ local function get_models(self, opts)
   end
 
   _cached_models = models
-  _cache_expires = utils.refresh_cache(_cache_file, config.adapters.opts.cache_for)
+  _cache_expires = utils.refresh_cache(_cache_file, config.adapters.opts.cache_models_for)
 
   return models
 end
--- a/lua/codecompanion/config.lua
+++ b/lua/codecompanion/config.lua
@@ -27,9 +27,9 @@ local defaults = {
     -- OPTIONS ----------------------------------------------------------------
     opts = {
       allow_insecure = false, -- Allow insecure connections?
-      cache_for = 1800, -- Cache adapter models for this long (seconds)
-      show_defaults = true, -- Show default adapters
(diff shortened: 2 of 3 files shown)


lua/codecompanion/adapters/copilot.lua
```diff
@@ -136,8 +138,8 @@ end
 ---@params opts? table
 ---@return table
 local function get_models(self, opts)
-  if _cached_available_models then
-    return _cached_available_models
+  if _cached_models and _cache_expires > os.time() then
```
dkimot, line 141: I'm getting an error on this line: `attempt to compare number with nil`. This should probably be something like

```lua
if _cached_models and _cache_expires and _cache_expires > os.time() then
```

dkimot: sorry to resurrect a closed PR, but I figured this would be the most reasonable approach

olimorris: Great spot thank you!

lua/codecompanion/config.lua
```diff
@@ -27,8 +27,9 @@ local defaults = {
     -- OPTIONS ----------------------------------------------------------------
     opts = {
       allow_insecure = false, -- Allow insecure connections?
-      show_defaults = true, -- Show default adapters
+      cache_models_for = 1800, -- Cache adapter models for this long (seconds)
```
dkimot, line 30: sorry to continue this, but i then got an error on `cache_models_for` not being set. it was failing on line 10 of `utils/adapters.lua` which i traced back to `adapters/copilot.lua:191`

here's an excerpt of my config which I changed to fix this:

```lua
    codecompanion.setup({
      adapters = {
        opts = {
          cache_models_for = 1800, -- added by me, did not exist before
          show_defaults = false,
        },
    })
```

this is probably down to me not understanding lua very well. not sure if i should have been using an `extend`? or where i went wrong. but, adding `cache_models_for` fixes it.

PR #1165, opened 2025-03-23
Update custom slash_command example

jinzhongjia:
The original example is no longer available in the current version

Note that I directly use lgithub's edit function, so the code format may be incorrect

### DIFF
--- a/doc/configuration/chat-buffer.md
+++ b/doc/configuration/chat-buffer.md
@@ -90,23 +90,29 @@ require("codecompanion").setup({
     chat = {
       slash_commands = {
         ["git_files"] = {
-          description = "List git files",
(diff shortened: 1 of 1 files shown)


jinzhongjia:
I suggest you format it locally

olimorris:
This is just an example in the docs. Are you saying it doesn't work?

jinzhongjia:
Yes, the main reason is the lack of role="user"

jinzhongjia:
And I replaced it with a system using vim, synchronous processing would be better

olimorris:
Thanks for the catch. I've added in the user role. This is meant to serve as an example of how a user can add a slash command rather than being an example they _should_ add.

PR #1167, opened 2025-03-24
Allow Empty Tables to Be Encoded as Array or Objects

atomdmac:
## Description

This may be too niche for anyone else to find useful.  However, it works well for me and may come in handy for others.

### TL;DR
Modify the utility that parses Lua tables to JSON so that it's possible to dictate whether an empty table should be encoded as an empty _array_ (current behavior) or an empty _object_.  To mark a table as an empty _object_, it needs to (ironically) include a single entry of `"__empty_object__"`.

### Full Explanation
I really like the UX of interacting with LLMs via Neovim.  However, my most common use-case so far is to use LLMs to summarize internet search results.  None of the current adapters support this out of the box, though it's easy enough to extend the `body` of any adapter to include the necessary inputs.

However, Google Gemini is my provider of choice and as luck would have it, the payload looks like this:
```json
...
"tools": [
  { "google_search": {} }
]
...
```

Unfortunately, the built-in `vim.json.encode` function will always parse the empty object as an array.  Since Google's API does not currently accept _any_ options for the `google_search` tool, it _has_ to be passed as an empty object.

## Related Issue(s)
N/A

## Screenshots
N/A

## Checklist

### DIFF
--- a/lua/codecompanion/http.lua
+++ b/lua/codecompanion/http.lua
@@ -3,6 +3,7 @@ local Path = require("plenary.path")
 local config = require("codecompanion.config")
 local log = require("codecompanion.utils.log")
 local util = require("codecompanion.utils")
+local json = require("codecompanion.utils.json")
 
 ---@class CodeCompanion.Client
 ---@field adapter CodeCompanion.Adapter
@@ -16,7 +17,7 @@ Client.static = {}
 Client.static.opts = {
   post = { default = Curl.post },
   get = { default = Curl.get },
-  encode = { default = vim.json.encode },
+  encode = { default = json.encode },
   schedule = { default = vim.schedule_wrap },
 }
 
--- a/lua/codecompanion/utils/json.lua
+++ b/lua/codecompanion/utils/json.lua
@@ -3,7 +3,11 @@ local M = {}
 ---@param data any
 ---@return string
 M.encode = function(data)
-  return vim.json.encode(data)
+  local output = vim.json.encode(data)
+  -- Replace instances where an empty array should actually be an empty object.
+  -- Specify an empty object by adding a single entry of "empty_object" to
+  -- the table.
+  return output:gsub('%["__empty_object__"%]', "{}")
 end
 
 ---Decode a yaml node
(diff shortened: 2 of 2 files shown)


atomdmac:
@olimorris Yes, it is a Gemini requirement, specifically if you want to enable the model to have access to internet search results.  [Here is the Gemini documentation outlining the payload I'm trying to accommodate with this change](https://ai.google.dev/gemini-api/docs/grounding?lang=rest#configure-search).  To be precise, it is to accommodate the empty object assigned to the `google_search` key.

The error that bounces back from Google's API occurs because the JSON encoder that CodeCompanion uses (i.e. `vim.json.encode` encodes an empty table as an array but Google's API only accepts an empty object as a value for `google_search`.  In fact, the object _has_ to be empty because there aren't any valid options to attach to it.

The `__empty_object__` "notation" is simply a way to tell the parser shim that "this should be encoded as an empty object, NOT as as array."

Does that answer your question?

To be honest, I'm not _thrilled_ with this solution.  I'm very open to alternatives.  But this is working for me right now.

olimorris:
@atomdmac I can't see anything referencing the `__empty_object__` on that page. But given this is Gemini specific, we'd need to find a way to support this in the adapter itself. With your proposed implementation, it would require some comprehensive testing to make sure it doesn't break interactions with other LLMs.

I am adding support for native tool calling in #1141. My proposal is to wait for this to be merged. It will make for calling LLM provider tools much, much easier.

atomdmac:
@olimorris Sorry, I'm not explaining myself very well :P 

The `"__empty_object__"` notation isn't a Gemini requirement per se.  The Gemini requirement is that an empty object be given as a value for the `"google_search"` key in the REST call to the Google API (as mentioned in the Gemini docs that I linked).

The `"__empty_object__"` notation is a construction that I made up.  The goal is simply to allow the user a way to tell the Lua -> JSON parser to convert a table to an empty _object_ instead of an empty _array_ (which is how it will always treat empty tables).  As far as I can tel, there is no other way to tell `vim.json.encode` to do this.

In any case, it sounds like you're already actively working on a more comprehensive implementation for tool calling so I think it makes sense to disregard my changes in favor of yours.

Thanks for taking a look (and for making this plugin!)

Issue #1175, opened 2025-03-27
[Bug]: Inline variables throw error if callback is a function


ravitemer:
### Did you check the docs and existing issues?

- [x] I have read all the docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched for existing issues/discussions
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

NVIM v0.10.4

### Operating system/version

Linux 

### Describe the bug

With the following,
```lua
	inline = {
		adapter = adapter,
		variables = {
			test = {
				description = "Test variable",
				callback = function(self)
					vim.notify("Test variable callback")
				end,
			},
		},
	},
```
I get 

```txt
   Error  04:33:05 msg_show.lua_error   CodeCompanion #test Error executing Lua callback: ...m/lua/codecompanion/strategies/inline/variables/init.lua:71: attempt to concatenate local 'callback' (a function value)
stack traceback:
	...m/lua/codecompanion/strategies/inline/variables/init.lua:71: in function 'output'
	...panion.nvim/lua/codecompanion/strategies/inline/init.lua:284: in function 'inline'
	...ubuntu/codecompanion.nvim/lua/codecompanion/commands.lua:67: in function <...ubuntu/codecompanion.nvim/lua/codecompanion/commands.lua:43>
```

because it doesn't support callbacks being functions. 

### Steps To Reproduce

Add the test variable to strategies.inline.variables

### Expected Behavior

It should resolve the variable successfully

### init.lua file

```lua
--[[
NOTE: Set the config path to enable the copilot adapter to work.
It will search the following paths for a token:
  - "$CODECOMPANION_TOKEN_PATH/github-copilot/hosts.json"
  - "$CODECOMPANION_TOKEN_PATH/github-copilot/apps.json"
--]]
vim.env["CODECOMPANION_TOKEN_PATH"] = vim.fn.expand("~/.config")

vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

-- Your CodeCompanion setup
local plugins = {
  {
    "olimorris/codecompanion.nvim",
    dependencies = {
      { "nvim-treesitter/nvim-treesitter", build = ":TSUpdate" },
      { "nvim-lua/plenary.nvim" },
      -- Test with blink.cmp (delete if not required)
      {
[...]

PR #1176, opened 2025-03-27
fix: (#1175) 

ravitemer:
inline variables can be functions, chat variable function callbacks gets self

CodeCompanion.Inline.Variable config.callback can be a function. This function will have a single argument i.e CodeCompanion.Inline.Variable

CodeCompanion.Variable.User config.function callback will receive a single argument i.e CodeCompanion.Variable.User

## Checklist

- [x] I've run `make docs` to update the vimdoc pages


olimorris:
Appreciate the PR! This was something I didn't add on account for seeing if users wanted it so thank you for taking the time to PR it.

PR #1182, opened 2025-03-28
feat(symbols): add symbols for Java

Chainso:
## Description

Adds the Treesitter symbol queries for Java from [Aerial](https://github.com/stevearc/aerial.nvim/blob/master/queries/java/aerial.scm).

## Related Issue(s)

N/A

## Screenshots

N/A

## Checklist

### DIFF
--- a/queries/java/symbols.scm
+++ b/queries/java/symbols.scm
@@ -1,0 +1,40 @@
+; Expanded from Ref: https://github.com/stevearc/aerial.nvim/blob/master/queries/java/aerial.scm
+; MIT License
+
+(package_declaration
+  (scoped_identifier) @name
+  (#set! "kind" "Import")) @symbol
+
+(import_declaration
+  (scoped_identifier) @name
+  (#set! "kind" "Import")) @symbol
+
+(interface_declaration
+  name: (identifier) @name
+  (#set! "kind" "Interface")) @symbol
+
+(method_declaration
+  name: (identifier) @name @start
+  (#set! "kind" "Method")) @symbol
+
+(constructor_declaration
+  name: (identifier) @name
+  (#set! "kind" "Constructor")) @symbol
+
+(class_declaration
+  name: (identifier) @name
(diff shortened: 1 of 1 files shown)


queries/java/symbols.scm
```diff
@@ -0,0 +1,40 @@
...
+  name: (identifier) @name
+  (#set! "kind" "Interface")) @symbol
+
+(method_declaration
+  name: (identifier) @name @start
+  (#set! "kind" "Method")) @symbol
+
+(constructor_declaration
+  name: (identifier) @name
+  (#set! "kind" "Constructor")) @symbol
```
olimorris, line 22: Constructor isn't a `kind` in the symbols command ([source](https://github.com/olimorris/codecompanion.nvim/blob/51fe5a782dbbd5cad8189420cb8d38fd7c245684/lua/codecompanion/strategies/chat/slash_commands/symbols.lua#L238))

olimorris:
Thanks for this PR, really good to add Java. I've added a couple of comments as constructor and field aren't listed as a kind and I'm not entirely sure they're really need for a symbol.

Chainso:
Would agree here. Since we aren't gleaning the full type signature information from methods having the constructor doesn't help here, will remove.

On another note, is the design to keep symbols across all languages lightweight intentional? The context would greatly benefit from having parameter and return type information from symbols without needing to pass the buffer/buffer scope into the context as well.

queries/java/symbols.scm
Chainso: Removing this declaration.

olimorris:
> On another note, is the design to keep symbols across all languages lightweight intentional

Yes, exactly. Designed to be a symbolic outline which then the LLM can probe further if required.

Also you've left `Field` in as well which isn't a kind that's supported. If you can remove that I'll be happy to merge.

Chainso:
Removed field, should be alright now.

PR #1189, opened 2025-04-01
Improve diff opt default

leiserfg:
## Checklist

### DIFF
--- a/lua/codecompanion/config.lua
+++ b/lua/codecompanion/config.lua
@@ -963,7 +963,15 @@ You must create or modify a workspace file through a series of prompts over mult
       enabled = true,
       close_chat_at = 240, -- Close an open chat buffer if the total columns of your display are less than...
       layout = "vertical", -- vertical|horizontal split for default provider
-      opts = { "internal", "filler", "closeoff", "algorithm:patience", "followwrap", "linematch:120" },
+      opts = {
+        "internal",
+        "filler",
+        "closeoff",
+        "algorithm:histogram", -- https://adamj.eu/tech/2024/01/18/git-improve-diff-histogram/
+        "indent-heuristic", -- https://blog.k-nut.eu/better-git-diffs
+        "followwrap",
+        "linematch:120",
+      },
       provider = "default", -- default|mini_diff
     },
     inline = {


olimorris:
Can you share what this actually does? Before and after screenshots?

leiserfg:
Check the links in the comments of the changes.

olimorris:
Firstly, you can apply this yourself in your config. If i merge this, I'm changing everyone's settings by default. You've failed to explain what this does and why it matters to CodeCompanion...

leiserfg:
Well, in that sense codecompanion is already changing users defaults by setting `diffopt` and  not reverting the change to whatever the user had before. Neither patient nor histogram are the default diffing algorithm (it's myers).

Regarding why the addition, `histogram` gives the same output as `patience` most of the time, but as per https://link.springer.com/article/10.1007/s10664-019-09772-z  it's better for code. 
And the indent heuristic uses indention to tie break between potential candidates (which also gives a better diffs in code but does and does not affect much in regular text).

But I understand if you don't like them, or you don't wanna change the defaults of your plugin. In that case will you accept a PR to :
 - keep the algorithm from the user defined `diffopts` untouched? (like, set `patience` only if another one is not explicitly set)
 - keep any other extra option there that the  settings don't care about (like override vertical but don't touch `indent-heuristic`)
 - rollback `diffopts` to whatever the user had configured before these usage of diff
 
 ?

olimorris:
You're far more knowledgeable on Neovim's diff options than I am or intend to be so appreciate your input. I think most of the users will care more about how the diff looks than how it functions. Can you share some screenshots showing how it's improved?

leiserfg:
diffing these two files
[original.ts.txt](https://github.com/user-attachments/files/19583244/original.ts.txt)
[copy.ts.txt](https://github.com/user-attachments/files/19583245/copy.ts.txt)

with patience
patience

with histogram + indent-heuristic
histogram

leiserfg:
I'm using `vim.opt.fillchars = { eob = " ", diff = "╱" }`

leiserfg:
:heavy_check_mark:

PR #1195, opened 2025-04-02
Improve Telescope action implementation

petobens:
This improves the CodeCompanion Telescope  provider action palette implementation by:

1. Adding a preview buffer that either shows the entry description or prompt content (useful to display custom user prompts)
2. Reworking the entry display to show both the entry name and strategy properly ordered in columns and with adequate highlights.

Before:
before

After:
after


olimorris:
Appreciate the PR, thanks

PR #1204, opened 2025-04-04
fix(ui): respect opts.visible for non-"system" roles

### LINKED ISSUE #1203: [Bug]: opts.visible is not respected for non-"system" roles in chat
### Did you check the docs and existing issues?

- [x] I have read all the docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched for existing issues/discussions
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.11.0

### Operating system/version

macOS 15.4

### Describe the bug

The "user" role `content` is added to the chat buffer even if `opts = { visible = false }` is set.

See "Steps to Reproduce" below and the [fix commit in my fork](https://github.com/stormc/codecompanion.nvim/commit/29ca54cc94eda74cbf15ec2df51986c67d56f46d).

### Steps To Reproduce

1. Create a custom prompt
2. Set `opts = { visible = false }` in the "user" prompt
3. Start a chat and the "user" prompt `content` is added to the chat buffer regardless of setting in 2.

### Expected Behavior

The "user" prompt `content` is not added to the chat buffer if `opts = { visible = false }` is set.

### init.lua file

```lua
Skipped, as there is no relevant log output in the affected function `add_messages_to_buf(msgs)`.
```

### Log output

There is no relevant log output in `lua/codecompanion/strategies/chat/ui.lua` for the function `add_messages_to_buf(msgs)`.

### Have you provided and tested with a repro.lua file?

- [x] Yes, I have tested and provided a repro.lua file

stormc:
## Description

opts.visible is not respected for non-"system" roles in chat.

## Related Issue(s)

  - Fixes #1203

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/ui.lua
+++ b/lua/codecompanion/strategies/chat/ui.lua
@@ -235,7 +235,7 @@ function UI:render(context, messages, opts)
 
   local function add_messages_to_buf(msgs)
     for i, msg in ipairs(msgs) do
-      if msg.role ~= config.constants.SYSTEM_ROLE or (msg.opts and msg.opts.visible ~= false) then
+      if msg.role ~= config.constants.SYSTEM_ROLE and (msg.opts and msg.opts.visible ~= false) then
         -- For workflow prompts: Ensure main user role doesn't get spaced
         if i > 1 and self.last_role ~= msg.role and msg.role ~= config.constants.USER_ROLE then
           spacer()


olimorris:
Brilliant spot! Thanks for investigating and solving.

petobens:
Hi! Sorry but I don't quite get this change. Before the change if write a prompt like
```lua
 [' Bash Developer'] = {
            strategy = 'chat',
            description = 'Act as an expert Bash developer.',
            opts = {
                short_name = 'bash_role',
                is_slash_cmd = true,
                auto_submit = false,
                ignore_system_prompt = true,
            },
            prompts = {
                {
                    role = 'system',
                    content = [[
You are an expert Bash developer.
When giving code examples show the generated output.
]],
                    opts = { visible = true },
                },
                {
                    role = 'user',
                    content = [[]],
                },
            },
        }
```
and then  as in the GIF invoke the action palette and run the prompt then i) the system prompt would be visible and ii) the `opts = { visible = false }` flag would make it invisible. However with the change in this commit the content is always invisible and the opts.visible flag doesn't seem to be honored.

Peek 2025-04-05 16-50

olimorris:
Yes I've noticed this has caused intended consequences with `CodeCompanionChat`. I've reverted this for now with #1218.

Issue #959, opened 2025-02-19
[Bug]: ignore_system_prompt gives a nil error


petobens:
### Did you check the docs and existing issues?

- [x] I have read all the docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched for existing issues/discussions
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

v0.11.0-dev-1780+gf3ce67549c

### Operating system/version

Arch Linux

### Describe the bug

When using the `ignore_system_prompt = true,` option for prompts I get the following error:
```
Error executing Lua callback: ...on.nvim/lua/codecompanion/strategies/chat/tools/init.lua:451: attempt to index local 'message' (a nil value)
stack traceback:
        ...on.nvim/lua/codecompanion/strategies/chat/tools/init.lua:451: in function 'find'
        ...on.nvim/lua/codecompanion/strategies/chat/tools/init.lua:503: in function 'parse'
        ...ompanion.nvim/lua/codecompanion/strategies/chat/init.lua:675: in function 'apply_tools_and_variables'
        ...ompanion.nvim/lua/codecompanion/strategies/chat/init.lua:730: in function 'submit'
        ...ompanion.nvim/lua/codecompanion/strategies/chat/init.lua:316: in function 'create_chat'
        ...codecompanion.nvim/lua/codecompanion/strategies/init.lua:137: in function 'on_choice'
        /usr/share/nvim/runtime/lua/vim/ui.lua:55: in function 'select'
        ...lugins/codecompanion.nvim/lua/codecompanion/utils/ui.lua:224: in function 'action_palette_selector'
        ...ion.nvim/lua/codecompanion/providers/actions/default.lua:25: in function 'actions'
        ...lugins/codecompanion.nvim/lua/codecompanion/commands.lua:126: in function <...lugins/codecompanion.nvim/lua/codecompanion/commands.lua:125>
```

### Steps To Reproduce

1. Open nvim with the init file below
2. Run `:CodeCompanionActions`
3. Select 2 (Bash Developer) prompt

### Expected Behavior

No error after 3.

![Image](https://github.com/user-attachments/assets/f113b5ed-adab-4ad3-88aa-5f29be0370c0)

### File to Reproduce

```lua
local root = '/tmp/nvim-minimal'

-- Set stdpaths to use root dir
for _, name in ipairs({ 'config', 'data', 'state', 'cache' }) do
[...]

PR #1225, opened 2025-04-06
Add blank line before user role

petobens:
This addresses my second comment https://github.com/olimorris/codecompanion.nvim/issues/959#issuecomment-2694619521 by adding a blank line between the system role content and the user role heading.

spacer

### DIFF
--- a/lua/codecompanion/strategies/chat/ui.lua
+++ b/lua/codecompanion/strategies/chat/ui.lua
@@ -242,6 +242,9 @@ function UI:render(context, messages, opts)
         end
 
         if msg.role == config.constants.USER_ROLE and last_set_role ~= config.constants.USER_ROLE then
+          if last_set_role ~= nil then
+            spacer()
+          end
           self:set_header(lines, self.roles.user)
         end
         if msg.role == config.constants.LLM_ROLE and last_set_role ~= config.constants.LLM_ROLE then


olimorris:
I notice that this causes a blank line to be inserted if there's only a user role in a chat buffer. If we can add a check for that this will be good to merge.

petobens:
Hi! Sorry for the late reply. Added a guard for your request (assuming I understood correctly)

petobens:
@olimorris I used a simple user role only prompt like the following to test this out:
```lua
['Test'] = {
            strategy = 'chat',
            description = 'Act as an test developer.',
            opts = {
                short_name = 'test_role',
                is_slash_cmd = true,
                auto_submit = false,
                ignore_system_prompt = true,
            },
            prompts = {
                {
                    role = 'user',
                    content = [[Foo]],
                },
            },
        }
```

petobens:
Hi @olimorris. Sorry for the ping ,but do you want me to make further changes to this PR?

olimorris:
Hey thanks for this. I'm fully tied up with another PR so will try and take a look in the week. There were a few additional scenarios I wanted to test.

petobens:
Makes sense. Thanks for the reply

olimorris:
Thanks for this @petobens! Fixes a stupid UI annoyance really nicely.

petobens:
Thanks for an awesome plugin and kudos on the function tools work :)

PR #1245, opened 2025-04-13
feat: allow toggling the visibility of individual references.

Davidyz:
## Description

This PR allows certain references to be hidden from the chat buffer. This is useful for tools/slash commands that add _a lot_ of references to the chat buffer.

Reference providers (tool, slash command, variable, etc.) can use the `visible` option to choose whether it should be shown in the chat UI:
```lua
agent.chat.references:add({
  source = "tool",
  id = "ref_id",
  opts = { visible = false },
})
```

The value for the `visible` key falls back to `config.display.chat.show_references`, so it should respect existing configurations.

## Related Issue(s)

See [here](https://github.com/Davidyz/VectorCode/pull/72#issuecomment-2799884579).

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/references.lua
+++ b/lua/codecompanion/strategies/chat/references.lua
@@ -78,6 +78,9 @@ end
 ---@param ref CodeCompanion.Chat.Ref
 ---@param row integer
 local function add(chat, ref, row)
+  if not ref.opts.visible then
+    return
+  end
   local lines = {}
 
   table.insert(lines, string.format("> - %s", ref.id))
@@ -121,6 +124,11 @@ function References:add(ref)
     -- Ensure both properties exist with defaults
     ref.opts.pinned = ref.opts.pinned or false
     ref.opts.watched = ref.opts.watched or false
+    ref.opts.visible = ref.opts.visible
+
+    if ref.opts.visible == nil then
+      ref.opts.visible = config.display.chat.show_references
+    end
     table.insert(self.Chat.refs, ref)
     -- If it's a buffer reference and it's being watched, start watching
     if ref.bufnr and ref.opts.watched then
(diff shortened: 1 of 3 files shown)


olimorris:
Firstly, great PR and appreciate the contribution. I think this is a valid PR and potentially allows for a UI annoyance of mine which is references that are 10+ lines long.

olimorris:
Be interested to get your views...If I'm a token conscious user (which some might be given Copilot has a limit of 90k tokens) is displaying all of the references in the debug window, via `gd`, acceptable?

This PR raised a question in my mind which is some users may wish to edit references still and remove them manually (adding by mistake for instance).

Also...is adding an [event](https://codecompanion.olimorris.dev/usage/events.html) for a reference being added useful for you? I might add it anyway.

Davidyz:
> is displaying all of the references in the debug window, via `gd`, acceptable?

If it's a debug window, I think it's ok (or even expected) to display all refs. Awkwardly, I didn't actually check whether my PR breaks this bc I use `gd` for LSP go to definition (it's not broken btw).

> This PR raised a question in my mind which is some users may wish to edit references still and remove them manually (adding by mistake for instance).

I think this is a fair feature. For removing references, it could be just a picker (telescope, fzf-lua, etc.), which should be quite straightforward to implement.

olimorris:
> If it's a debug window, I think it's ok (or even expected) to display all refs. Awkwardly, I didn't actually check whether my PR breaks this bc I use gd for LSP go to definition (it's not broken btw).

All keymaps are scoped to a chat buffer when it's created which take precedence over any existing keymaps so would always be okay.

Davidyz:
> All keymaps are scoped to a chat buffer when it's created which take precedence over any existing keymaps so would always be okay.

Not when I'm using LSP backend for vectorcode and it registers the chat buffer to the Vectorcode LSP, which fires the `LspAttach` autocmd, which then adds the LSP keymaps 🤣 this may cause the LSP `gd` being registered after the codecompanion keymaps. It's more of an issue with my config.

PR #1269, opened 2025-04-19
Show open chat preview in telescope

petobens:
As it can be seen in the GIF we now get a chat preview in the Open Chats picker.

Some questions/comments:
1. I could modify the static.lua file and change the strategy field (from say `chat` to `open_chat`) or add a new field an check for that rather than
 `[No Messages]`.  
2. I fixed a minor width thingy along the way.

Peek 2025-04-19 19-27

This PR and [this mapping](https://github.com/petobens/dotfiles/blob/7587041a1e97230b21b463b8fb40b2f7ef2e6647/nvim/lua/plugin-config/codecompanion_config.lua#L462-L471) basically fixes https://github.com/olimorris/codecompanion.nvim/discussions/1243 for Telescope.


chadmckenna:
This had been annoying me and was about to jump in and see if I could fix this, then I came across this PR. Beat me to it! Thanks @petobens

PR #1331, opened 2025-05-01
feat/prompt-decorator

olimorris:
## Description

Small QOL update to allow user's to decorate their prompts prior to sending to an LLM.

### DIFF
--- a/.vectorcode/vectorcode.include
+++ b/.vectorcode/vectorcode.include
@@ -1,0 +1,4 @@
+doc/**
+lua/**
+queries/**
+tests/**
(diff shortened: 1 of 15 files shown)


petobens:
Hi @olimorris great feature! I've tried and have a couple of questions (I can move this to the discussions if needed):

1. Maybe this is intended but I noticed the `adapter.schema` field doesn't have the `model` field. Is it possible to add that? My use case is to add some conditional logic as in:
```lua
                prompt_decorator = function(message, adapter, _)
                    if adapter.name == 'ollama'  and adapter.schema.model.default == 'some_model' then
                        return string.format([[/no_think %s]], message)
                    else
                        return message
                    end
                end,
```
2. I was wondering whether you might consider adding a post process config decorator option now that we have a preprocess (prompt) one.  For instance I would like to remove the `-----------------` lines present in some messages.

Thanks in advance!

Issue #1311, opened 2025-04-30
[Bug]: Error executing lua: ...wrapped-0.11.0/share/nvim/runtime/lua/vim/treesitter.lua:131: str: expected string, got nil


blindFS:
### Did you check the docs and existing issues?

- [x] I have read all the docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched for existing issues/discussions
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.11.0

### Operating system/version

MacOS 15.4.1

### Describe the bug

codecompanion.nvim upgraded to the latest release 15.0.0. 

Got the following errors after fetching file content by GitHub MCP function call, then pressing enter without any text:

```
E5108: Error executing lua: ...wrapped-0.11.0/share/nvim/runtime/lua/vim/treesitter.lua:131: str: expected string, got nil
stack traceback:
	[C]: in function 'error'
	vim/shared.lua: in function 'validate'
	...wrapped-0.11.0/share/nvim/runtime/lua/vim/treesitter.lua:131: in function 'get_string_parser'
	...on.nvim/lua/codecompanion/strategies/chat/references.lua:161: in function 'clear'
	...ompanion.nvim/lua/codecompanion/strategies/chat/init.lua:694: in function 'submit'
	...anion.nvim/lua/codecompanion/strategies/chat/keymaps.lua:216: in function 'rhs'
	...y/codecompanion.nvim/lua/codecompanion/utils/keymaps.lua:80: in function <...y/codecompanion.nvim/lua/codecompanion/utils/keymaps.lua:77>
```

### Steps To Reproduce

Here's the conversation

<details>
## Me

> Context:
> - <tool>use_mcp_tool</tool>
> - <tool>access_mcp_resource</tool>
> - <tool>editor</tool>
> - <tool>files</tool>

@mcp @editor @files https://github.com/folke/snacks.nvim/issues/1808 analyse the related code of this issue.
tell me how to fix it

## CodeCompanion (Gemini)

Okay, I can help with that. To analyze the issue and the related code, I first need to fetch the details of the issue from GitHub.

### Tool Output

The `use_mcp_tool` call returned the following text:
{
  ...
[...]

PR #1336, opened 2025-05-01
Feat/better chat initialization

ravitemer:
## Description

@olimorris I have come across some bugs while implementing history. I have provided detailed description in each commit.
 
I have not opened a PR for each of these commits to keep it simple for you while testing. I am happy to open separate PRs if you  need them. 

## Checklist

- [x] I've added

### DIFF
--- a/doc/codecompanion.txt
+++ b/doc/codecompanion.txt
@@ -1,4 +1,4 @@
-*codecompanion.txt*          For NVIM v0.10.0         Last change: 2025 May 03
+*codecompanion.txt*          For NVIM v0.10.0         Last change: 2025 May 04
 
 ==============================================================================
 Table of Contents                            *codecompanion-table-of-contents*
@@ -1657,7 +1657,7 @@ The |codecompanion-usage-chat-buffer-agents-cmd-runner| tool enables an LLM to
 execute commands on your machine. This can be useful if you wish the LLM to run
 a test suite on your behalf and give insight on failing cases. Simply tag the
 `@cmd_runner` in the chat buffer and ask it run your tests with a suitable
-command e.g.� `pytest`.
+command e.g. `pytest`.
 
 
 NAVIGATING BETWEEN RESPONSES IN THE CHAT BUFFER ~
@@ -2216,7 +2216,9 @@ The events that you can access are:
 - `CodeCompanionChatOpened` - Fired after a chat has been opened
 - `CodeCompanionChatHidden` - Fired after a chat has been hidden
 - `CodeCompanionChatClosed` - Fired after a chat has been permanently closed
+- `CodeCompanionChatSubmitted` - Fired after a chat has been submitted
 - `CodeCompanionChatStopped` - Fired after a chat has been stopped
+- `CodeCompanionChatCleared` - Fired after a chat has been cleared
 - `CodeCompanionChatAdapter` - Fired after the adapter has been set in the chat
 - `CodeCompanionChatModel` - Fired after the model has been set in the chat
 - `CodeCompanionChatPin` - Fired after a pinned reference has been updated in the messages table
@@ -2657,7 +2659,7 @@ OpenAI adapter.
   as a great reference to understand how they’re working with the output of the
   API
 
-OPENAI�S API OUTPUT
+OPENAI’S API OUTPUT
 
 If we reference the OpenAI documentation
 <https://platform.openai.com/docs/guides/text-generation/chat-completions-api>
--- a/doc/usage/events.md
+++ b/doc/usage/events.md
@@ -10,6 +10,7 @@ The events that you can access are:
 - `CodeCompanionChatOpened` - Fired after a chat has been opened
 - `CodeCompanionChatHidden` - Fired after a chat has been hidden
 - `CodeCompanionChatClosed` - Fired after a chat has been permanently closed
+- `CodeCompanionChatSubmitted` - Fired after a chat has been submitted
 - `CodeCompanionChatStopped` - Fired after a chat has been stopped
 - `CodeCompanionChatCleared` - Fired after a chat has been cleared
 - `CodeCompanionChatAdapter` - Fired after the adapter has been set in the chat
--- a/doc/usage/events.md
+++ b/doc/usage/events.md
@@ -11,6 +11,7 @@ The events that you can access are:
 - `CodeCompanionChatHidden` - Fired after a chat has been hidden
 - `CodeCompanionChatClosed` - Fired after a chat has been permanently closed
 - `CodeCompanionChatStopped` - Fired after a chat has been stopped
+- `CodeCompanionChatCleared` - Fired after a chat has been cleared
 - `CodeCompanionChatAdapter` - Fired after the adapter has been set in the chat
 - `CodeCompanionChatModel` - Fired after the model has been set in the chat
 - `CodeCompanionChatPin` - Fired after a pinned reference has been updated in the messages table
--- a/lua/codecompanion/strategies/chat/agents/executor/init.lua
+++ b/lua/codecompanion/strategies/chat/agents/executor/init.lua
@@ -180,7 +180,6 @@ function Executor:close()
   log:debug("Executor:close")
   self.handlers.on_exit()
   util.fire("ToolFinished", { id = self.id, name = self.tool.name, bufnr = self.agent.bufnr })
-  chat.subscribers:process(chat)
   _G.codecompanion_current_tool = nil
 end
 
--- a/lua/codecompanion/strategies/chat/init.lua
+++ b/lua/codecompanion/strategies/chat/init.lua
@@ -842,9 +842,10 @@ local function ready_chat_buffer(chat)
   chat.references:render()
 
   -- If we're running any tooling, let them handle the subscriptions instead
-  if not chat.tools:loaded() then
-    chat.subscribers:process(chat)
-  end
+  -- if not chat.tools:loaded() then
+  --   chat.subscribers:process(chat)
+  -- end
+  chat.subscribers:process(chat)
 
   log:info("Chat request finished")
   chat:reset()
--- a/lua/codecompanion/strategies/chat/init.lua
+++ b/lua/codecompanion/strategies/chat/init.lua
@@ -888,6 +888,9 @@ function Chat:done(output, tools)
     self:add_message({
       role = config.constants.LLM_ROLE,
       tool_calls = tools,
+      opts = {
+        visible = false,
+      },
     })
     return self.agents:execute(self, tools)
   end
@@ -1183,6 +1186,10 @@ function Chat:add_tool_output(tool, for_llm, for_user)
   local output = self.adapter.handlers.tools.output_response(self.adapter, tool_call, for_llm)
   output.cycle = self.cycle
   output.id = make_id({ role = output.role, content = output.content })
+  output.opts = vim.tbl_extend("force", output.opts or {}, {
+    tag = "tool_output",
+    visible = true,
+  })
 
   local existing = find_tool_call(tool_call.id, self.messages)
   if existing then
--- a/lua/codecompanion/strategies/chat/init.lua
+++ b/lua/codecompanion/strategies/chat/init.lua
@@ -947,17 +947,12 @@ function Chat:check_references()
     end)
     :totable()
 
-  -- Clear any tool's schemas
-  self.tools.schemas = vim
-    .iter(self.tools.schemas)
-    :filter(function(tool_schemas)
-      if vim.tbl_contains(to_remove, tool_schemas) then
-        -- TODO: Remove from tools_in_use
-        return false
-      end
-      return true
-    end)
-    :totable()
+  for id, _ in pairs(self.tools.schemas) do
+    if vim.tbl_contains(to_remove, id) then
+      self.tools.schemas[id] = nil
+      self.tools.in_use[id] = nil
+    end
+  end
 end
 
 ---Add updated content from the pins to the chat buffer
--- a/lua/codecompanion/strategies/chat/init.lua
+++ b/lua/codecompanion/strategies/chat/init.lua
@@ -787,7 +787,9 @@ function Chat:submit(opts)
             end
             table.insert(output, result.output.content)
             self:add_buf_message(result.output)
-            self._tool_output_has_llm_response = true
+            if result.output.content ~= "" and not self._tool_output_has_llm_response then
+              self._tool_output_has_llm_response = true
+            end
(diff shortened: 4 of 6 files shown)


ravitemer:
I leave the double references being rendered to your expertise and will open an issue.

olimorris:
Firstly, thanks for the PRs and the efforts you've gone to here. I'm drowning in notifications atm so appreciate you an action first approach.

The damn double references are a real annoyance! I am planning on tackling that very soon.

Are you working on the `Check references error` bug?

ravitemer:
No, it is fixed. 

Only the double references error to be fixed currently.

ravitemer:
No rush at all here! These are not that commonly occurring bugs for users. Anyway they are fixed. Test them when you have some time. 

Always appreciate your efforts.

olimorris:
> When two tools are called at a time, each executor calls Chat:tools_done which call ready_chat_buffer where the references are being rendered two times.

I _think_ this was fixed on `main` btw...I had the agent's autocmds set in a loop. So two tools = two autocmds being created and two `Agent:reset()` and `Chat:tools_done()` calls.

olimorris:
> Check references error
> 
> Happens when more than one tool in the references and deleting one tool causes it. Having single tool and deleting it makes it an empty object instead.

This seems like this was because you had no prompt itself...some of the changes I pushed earlier might stop this from happening without being a direct fix

ravitemer:
> > Check references error
> > Happens when more than one tool in the references and deleting one tool causes it. Having single tool and deleting it makes it an empty object instead.
> 
> This seems like this was because you had no prompt itself...some of the changes I pushed earlier might stop this from happening without being a direct fix

I tried with the latest main branch and double-ref branch. We might need a direct fix for this. 

`Chat:check_references` is changing `self.tools.schemas` which should be a dictionary to a table. 

```lua
  -- Clear any tool's schemas
  self.tools.schemas = vim
    .iter(self.tools.schemas)
    :filter(function(tool_schemas)
      if vim.tbl_contains(to_remove, tool_schemas) then
        -- TODO: Remove from tools_in_use
        return false
      end
      return true
    end)
    :totable()
```

When there is one tool reference like 

```md
> Context:
> - <tool>files</tool>
```

and we delete the tool ref line, `Chat:check_references()` sets self.tools.schemas to empty `{}` which seems to work fine as it is inferred as a dictionary.

When we have two tool references like

```md
> Context:
> - <tool>use_mcp_tool</tool>
> - <tool>access_mcp_resource</tool>
```

and we delete any tool ref line, `Check:check_references()` sets self.tools.schemas to an array.

```lua
{
"<tool>use_mcp_tool</tool>",
{
--schema
}
}
```

In the below video if I send some text with it, it throws `need dictionary, received array` error. If I don't send any text at all, we don't get any error but the references are not removed from `self.refs` leading two both tool lines being rendered. 



https://github.com/user-attachments/assets/9d61e168-2be8-4b2e-b2ab-ccd6ce69c292

ravitemer:
> > Check references error
> > Happens when more than one tool in the references and deleting one tool causes it. Having single tool and deleting it makes it an empty object instead.
> 
> This seems like this was because you had no prompt itself...some of the changes I pushed earlier might stop this from happening without being a direct fix

The double references being rendered is fixed. But we got a new side effect where the result of the first run tool is shown for other tools as well. 


https://github.com/user-attachments/assets/0e0bf055-c14f-4701-bf2d-7a3daa44b3d0

olimorris:
> The double references being rendered is fixed. But we got a new side effect where the result of the first run tool is shown for other tools as well

I haven't merged the double references fix into main yet. There's still occasions when it will occur. But, off the top of my head, nothing has changed that would cause the duplicate tool output to be shown.

The agents/tools implementation leaves it up to the tool itself regarding the output shown in the buffer. So, if a Lua function tool runs successfully:

https://github.com/olimorris/codecompanion.nvim/blob/99b2c1a8e63f0d7f6cd8d015eda65fe449073f5f/lua/codecompanion/strategies/chat/agents/executor/func.lua#L67-L101

Or a cmd tool runs successfully:

https://github.com/olimorris/codecompanion.nvim/blob/99b2c1a8e63f0d7f6cd8d015eda65fe449073f5f/lua/codecompanion/strategies/chat/agents/executor/cmd.lua#L79-L87

Then `self.executor:success` is called:

https://github.com/olimorris/codecompanion.nvim/blob/99b2c1a8e63f0d7f6cd8d015eda65fe449073f5f/lua/codecompanion/strategies/chat/agents/executor/init.lua#L167-L174

which ultimately leads to the tool's `output.success` handler being called:

https://github.com/olimorris/codecompanion.nvim/blob/99b2c1a8e63f0d7f6cd8d015eda65fe449073f5f/lua/codecompanion/strategies/chat/agents/executor/init.lua#L67-L71

In MCPHub's `success` handler, I see:

https://github.com/ravitemer/mcphub.nvim/blob/81394b54e292a50361aba11198e5fa4ebb97d304/lua/mcphub/extensions/codecompanion/utils.lua#L127-L145

I'm wondering if changing:

```lua
local result = has_function_calling and stdout[1] or cmd[1]
```

to:

```lua
local result = has_function_calling and stdout[#stdout] or cmd[#cmd]
```

would fix this...

(Sorry a lot of this post is me talking out loud 😆)

ravitemer:
Sorry, that was it! 🤐 

```lua
            callback = {
                name = action_name,
                cmds = { utils.create_handler(action_name, has_function_calling, opts) },
```

I thought the stdout table corresponds to the cmds table in the tool schema. As we only have one function cmd inside cmds I didn't bother to change it to stdout[#stdout] like your builtin tools implementation. 

Sorry and thank you for wasting your time. 

And can you replicate this https://github.com/olimorris/codecompanion.nvim/pull/1336#issuecomment-2848443593 

I have made sure to not test it with `@mcp` tool 😅

olimorris:
> Sorry and thank you for wasting your time.

Never! Don't apologise. It's good to test the plugin logic and my expectations versus reality 😄 .

> And can you replicate this https://github.com/olimorris/codecompanion.nvim/pull/1336#issuecomment-2848443593

I haven't tried so I'll take a look over the weekend. I'll make sure we add some test coverage for this too

olimorris:
I can recreate the references error when tools are deleted. Seems like some weird junk is getting added to the tools table.

olimorris:
I've also noticed that if my prompt is:

`I want to test the @mcp tools. What should we test first?`

It actually gets sent over as:

`I want to test the  tools. What should we test first?`

I'm thinking this should really be:

`I want to test the access_mcp_resource, use_mcp_tool tools. What should we test first?`

olimorris:
#1357 should now fix the tools and references error. Can you try it out and then I can merge

ravitemer:
Thank you! it is working now. But a little bug when user doesn't send a message. As we are currently checking for a message to do 


https://github.com/user-attachments/assets/682af7bf-05a9-4bb1-85c9-6edb5e5e6cce


```lua
  -- NOTE: Sometimes the last message is a tool call so we need to account for that
  if message then
    message = self.references:clear(self.messages[#self.messages])
    self:replace_vars_and_tools(message)
    self:check_references()
    self:add_pins()
  end
```

As `check_references()` doesn't actually depend on the user's message we can safely take it out of the condition.

```lua
-- NOTE: Sometimes the last message is a tool call so we need to account for that
  if message then
    message = self.references:clear(self.messages[#self.messages])
    self:replace_vars_and_tools(message)
    self:add_pins()
  end
  self:check_references()
```
---

And as you can see from the video and the image below, I have the history extension ready. 

image


Before I make it public the extension needs the following commits to be merged or you can update along with the delete-references-fix branch. Should I consider opening a separate PR as merging this PR is dirty?

1. fix: subscribers not called when some tool is loaded in the chat.  https://github.com/olimorris/codecompanion.nvim/pull/1336/commits/a370dbb7fed6e61463edf2eda3efd8f47868ccb5

```lua
    -- If we're running any tooling, let them handle the subscriptions instead
    if not chat.tools:loaded() then
      chat.subscribers:process(chat)
    end
  end
```
2. Add `ChatCleared` and `ChatSubmitted` events 
3. Some other minor rendering fixes.

* [fix: user role messages with visible=false rendering in the chat](https://github.com/olimorris/codecompanion.nvim/pull/1336/commits/c8c9038ca9787d7ee1226451b8b1b34ec5cffe2d) 
[c8c9038](https://github.com/olimorris/codecompanion.nvim/pull/1336/commits/c8c9038ca9787d7ee1226451b8b1b34ec5cffe2d)

* [fix: empty space after llm header when just a tool is called without any text](https://github.com/olimorris/codecompanion.nvim/pull/1336/commits/2c8d76ef5f56a4315bea3bd37ced541d4d15c81d) 
[2c8d76e](https://github.com/olimorris/codecompanion.nvim/pull/1336/commits/2c8d76ef5f56a4315bea3bd37ced541d4d15c81d)


* [feat: allow tool_output messages to be rendered in the chat buffer](https://github.com/olimorris/codecompanion.nvim/pull/1336/commits/325cd6d61a5365460bbf9f5e4bbf0b780ff82d4d) 
[325cd6d](https://github.com/olimorris/codecompanion.nvim/pull/1336/commits/325cd6d61a5365460bbf9f5e4bbf0b780ff82d4d)

* [fix: ChatCreated should fire after all the initialization work.](https://github.com/olimorris/codecompanion.nvim/pull/1336/commits/ee56d2d8363ba6b24c8c955482d056cec3c3a868) 
[ee56d2d](https://github.com/olimorris/codecompanion.nvim/pull/1336/commits/ee56d2d8363ba6b24c8c955482d056cec3c3a868)

* [refactor: adds tags and explicit visible opts](https://github.com/olimorris/codecompanion.nvim/pull/1336/commits/189da313ed7a1c511e9a86e8fd86a1dbe32c2b4f)
[189da31](https://github.com/olimorris/codecompanion.nvim/pull/1336/commits/189da313ed7a1c511e9a86e8fd86a1dbe32c2b4f)

* [fix: update the chat.settings property when submitting the chat](https://github.com/olimorris/codecompanion.nvim/pull/1336/commits/99a03dcaa09a9001ad99781c90ca87adac5f0c01)

Sorry if this doesn't make any sense

olimorris:
> As check_references() doesn't actually depend on the user's message we can safely take it out of the condition.

The reason for this check was because of #1311. Basically, there are occurrences when the last message in a chat buffer isn't a user message, it might be a `tool`. So the conditional ensures that we don't check for references in a message that can't have them. Poor comment block on my part

olimorris:
> Before I make it public the extension needs the following commits to be merged or you can update along with the delete-references-fix branch. Should I consider opening a separate PR as merging this PR is dirty?

I'm happy to merge this as one. A lot of the commit's are very intertwined and come under the general theme of fixing the chat buffer. You've noticed and fixed _a lot_ of idiosyncrasies in the chat buffer so huge thanks! And people are going to lose their minds with the chat extension!

One thought I've had...most of the methods on the chat buffer are public right now. With the growing number of extensions, this means I'm at risk of breaking stuff for _a lot_ of people. I'm going to do a sweep and convert some to _private_ in a separate PR. Let me know if that would break anything in your extension. Then, I can safely work on the assumption that any method that is public, shouldn't be changed.

ravitemer:
> I don't think there will be many times when a user wants to remove a reference and send a blank message

Yeah. It's rare. Sorry just out of curiosity, 

> So the conditional ensures that we don't check for references in a message that can't have them.

`check_references` is to filter out any references user might have removed from the `> Content` block rather than checking from the actual message right. we are not using message that was parsed from the buffer (ts_parse_message) in check_references function at all right? Irrespective of whether the last message is user role or tool role, we are comparing the  `> Context` references block with that of self.refs and removing them. 

```lua
---Reconcile the references table to the references in the chat buffer
---@return nil
function Chat:check_references()
  local refs = self.references:get_from_chat()
  if vim.tbl_isempty(refs) and vim.tbl_isempty(self.refs) then
    return
  end

  -- Fetch references that exist on the chat object but not in the buffer
  local to_remove = vim
    .iter(self.refs)
    :filter(function(ref)
      return not vim.tbl_contains(refs, ref.id)
    end)
    :map(function(ref)
      return ref.id
    end)
    :totable()

  if vim.tbl_isempty(to_remove) then
    return
  end

  -- Remove them from the messages table
  self.messages = vim
    .iter(self.messages)
    :filter(function(msg)
      if msg.opts and msg.opts.reference and vim.tbl_contains(to_remove, msg.opts.reference) then
        return false
      end
      return true
    end)
    :totable()

  -- And from the refs table
  self.refs = vim
    .iter(self.refs)
    :filter(function(ref)
      return not vim.tbl_contains(to_remove, ref.id)
    end)
    :totable()

  for id, _ in pairs(self.tools.schemas) do
    if vim.tbl_contains(to_remove, id) then
      self.tools.schemas[id] = nil
      self.tools.in_use[id] = nil
    end
  end
end
```

ravitemer:
> One thought I've had...most of the methods on the chat buffer are public right now. With the growing number of extensions, this means I'm at risk of breaking stuff for a lot of people. I'm going to do a sweep and convert some to private in a separate PR. Let me know if that would break anything in your extension. Then, I can safely work on the assumption that any method that is public, shouldn't be changed.

This would be great!
 
 In history extension these are the main methods I am using these methods:
```lua
 Chat.new()
 local chat = chat_module.buf_get_chat(bufnr)
chat:close()
client.new({ adapter = settings }):request(payload, {
ui:hide()
ui:open() 
ui:is_active()
```

It is almost complete and have published it. Need to add some more tests. Can you please use it and suggest any ideas as you go along. Please note that until these commits are merged you might not be able see the full potential of it 😅 

https://github.com/ravitemer/codecompanion-history.nvim

olimorris:
Can you resolve the merge conflicts on this PR then I can merge 👍🏼

ravitemer:
Done! 

I haven't kept my change in ready_chat_buffer that relates to this commit: 

fix: subscribers not called when some tool is loaded in the chat. https://github.com/olimorris/codecompanion.nvim/commit/a370dbb7fed6e61463edf2eda3efd8f47868ccb5

This is my change.

```lua
  -- If we're running any tooling, let them handle the subscriptions instead
  -- if not chat.tools:loaded() then
  --   chat.subscribers:process(chat)
  -- end
  chat.subscribers:process(chat)
```

Can you please take a look at this, as far as I checked this worked fine.

ravitemer:
@olimorris Nothing more in this PR from me. 

Regarding (fix: subscribers not called when some tool is loaded in the chat. https://github.com/olimorris/codecompanion.nvim/commit/a370dbb7fed6e61463edf2eda3efd8f47868ccb5), you can test this with the history extension once merged as it uses the chat.subscribers:subscribe method. 

Will open a PR adding it to community extensions docs once I feel it's ready!

PR #1339, opened 2025-05-01
Fix/multiple tool fixes

olimorris:
## Description

A variety of fixes for a variety of issues

### DIFF
--- a/lua/codecompanion/adapters/openai.lua
+++ b/lua/codecompanion/adapters/openai.lua
@@ -168,10 +168,9 @@ return {
               local tool_index = tool.index and tonumber(tool.index) or i
 
               -- Some endpoints like Gemini do not set this (why?!)
-              -- We need this to ensure the #tool_calls = #tool_responses
               local id = tool.id
               if not id or id == "" then
-                id = string.format("call_%d", i)
+                id = string.format("call_%s_%d", json.created, i)
               end
 
               if self.opts.stream then
--- a/lua/codecompanion/adapters/openai.lua
+++ b/lua/codecompanion/adapters/openai.lua
@@ -170,7 +170,7 @@ return {
               -- Some endpoints like Gemini do not set this (why?!)
               local id = tool.id
               if not id or id == "" then
-                id = string.format("call_%s_%d", json.created, i)
+                id = string.format("call_%s_%s", json.created, i)
               end
 
               if self.opts.stream then
--- a/lua/codecompanion/adapters/openai.lua
+++ b/lua/codecompanion/adapters/openai.lua
@@ -171,7 +171,7 @@ return {
               -- We need this to ensure the #tool_calls = #tool_responses
               local id = tool.id
               if not id or id == "" then
-                id = string.format("call_%d", tool_index)
+                id = string.format("call_%d", i)
               end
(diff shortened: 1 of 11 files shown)


lua/codecompanion/adapters/openai.lua
```diff
@@ -171,7 +171,7 @@ return {
               -- We need this to ensure the #tool_calls = #tool_responses
               local id = tool.id
               if not id or id == "" then
-                id = string.format("call_%d", tool_index)
+                id = string.format("call_%d", i)
```
blindFS, line 174: I don't think this is good enough to fix #1311 (to be confirmed). As noted [in this comment](https://github.com/olimorris/codecompanion.nvim/issues/1311#issuecomment-2844635011), the collision happens across different messages within the same conversation.

olimorris: > I don't think this is good enough to fix https://github.com/olimorris/codecompanion.nvim/issues/1311 (to be confirmed)

Sorry - But chiming in on a PR without testing and saying it's not good enough is ridiculous. You may be surprised to learn that I have "confirmed" this as working for me.

I'm also not sure you've looked at Gemini's streaming output:

```
data: {"choices":[{"delta":{"role":"assistant","tool_calls":[{"function":{"arguments":"{\"units\":\"celsius\",\"location\":\"London\"}","name":"weather"},"id":"","type":"function"},{"function":{"arguments":"{\"units\":\"celsius\",\"location\":\"Paris\"}","name":"weather"},"id":"","type":"function"}]},"finish_reason":"tool_calls","index":0}],"created":1743628522,"model":"gemini-2.5-pro-exp-03-25","object":"chat.completion.chunk","usage":{"completion_tokens":38,"prompt_tokens":504,"total_tokens":542}}
```

blindFS: @olimorris I apologize if my wording sounds offensive and I don't mean it, just want the issue to get fixed, no matter in which way. And according to my test, this PR doesn't fix the id collision issue, actually I was pretty confident about that so I commented without testing.

blindFS: Sometimes I get too blunt, my bad. As you can see from my comments elsewhere, I have nothing but respect for your hard work and passion for this plugin. I'm all here to help, so don't get it personal.

Let me elaborate the problem I observed:

I have 2 cycles in the conversation. At first, Gemini fetches the details of an issue from GitHub. Its output has empty id, so that piece of code set an id of `call_1` for the call. 

With the new id, 

https://github.com/olimorris/codecompanion.nvim/blob/29ded250b186c550222aa6a4a648e52ad49dd85f/lua/codecompanion/strategies/chat/init.lua#L1190-L1195

inserts a new message for cycle 2, so far so good.
Then in the second round, Gemini responds with another function call to fetch the content of a file of interest. Since the it comes from another response, the id is still `call_1` after function `chatgpt_output`.

This time however, the first branch of line 1191 is taken and the tool output is then appended to the staled message of last cycle, thus causing the empty message content issue in the first place.

blindFS: It works, thanks!

PR #1340, opened 2025-05-02
fix(tools): `tool_call_id` collisions of gemini model

blindFS:
## Description

Use `json.created` as a workaround complement for gemini models when the call_id is missing.

## Related Issue(s)

Fixes the id collusion part of #1311 

## Checklist

- [ ] I've added

### DIFF
--- a/lua/codecompanion/adapters/openai.lua
+++ b/lua/codecompanion/adapters/openai.lua
@@ -171,7 +171,8 @@ return {
               -- We need this to ensure the #tool_calls = #tool_responses
               local id = tool.id
               if not id or id == "" then
-                id = string.format("call_%d", tool_index)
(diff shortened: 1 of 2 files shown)


olimorris:
Thanks for this, I'm addressing this in #1339.

If I recall, `json.created` remains static throughout  the stream. So if you have multiple tool calls then they'll end up with the same ID

blindFS:
> Thanks for this, I'm addressing this in #1339.
> 
> 
> 
> If I recall, `json.created` remains static throughout  the stream. So if you have multiple tool calls then they'll end up with the same ID

That's why toolidx is still used in the id format. Anyways, happy to see a better solution, will try #1339 later.

PR #1342, opened 2025-05-02
Fix: Set filetype to codecompanion for chat debug buffer

azzurro:
The debug chat buffer filetype was incorrectly set to lua.
This commit corrects the filetype to codecompanion, while preserving lua syntax highlighting.
This ensures proper plugin behavior based on the filetype.

### DIFF
--- a/lua/codecompanion/strategies/chat/debug.lua
+++ b/lua/codecompanion/strategies/chat/debug.lua
@@ -226,9 +226,10 @@ function Debug:render()
 
   ui.create_float(lines, {
     bufnr = self.bufnr,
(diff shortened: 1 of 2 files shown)


olimorris:
> The debug chat buffer filetype was incorrectly set to lua.

No, it was explictly set to be lua, because everything in it is in Lua.

You've not followed the PR template. What issues are you facing that warrant this change?

azzurro:
Disabling lcheylus/overlength.nvim based on filetype.

require('overlength').setup({
  disable_ft = {'codecompanion'},
})

PR #1347, opened 2025-05-02
feat(gemini): add gemini-2.5-flash

olimorris:
## Description

Added Gemini 2.5 Flash


jinzhongjia:
Geat, does the new tool call support support support support support do not support this model

PR #1349, opened 2025-05-03
Refactor @files tool to use diffs for update

pratyushmittal:
Hey @olimorris, thanks a lot for CodeCompanion. It is the reason I can still use `neovim`, my favourite editor.

I see a few issues in the current `@files` tool:

- It uses line numbers for read lines: LLMs are very bad at numbers. These are almost always wrong.
- The `start_line` and `end_line` parameters are often being used with other actions too. This too causes bad JSONs and outputs.
- The schema has too many options and parameters.
- The parameters are not well defined: the format for the search string or replacement was not well defined

[The `edit` rarely works for me](https://github.com/olimorris/codecompanion.nvim/discussions/1329) when handling multiple long files due to the above issues. Even the expensive models including `gpt-4.1` and `o4-mini` struggle.

GPT-4.1's recently published a prompt guide. It recommends a specific prompt technique to handle edits using diffs.
https://cookbook.openai.com/examples/gpt4-1_prompting_guide

Aider also uses the same technique for edits. They have documented various similar techniques here: https://aider.chat/docs/more/edit-formats.html

I have implemented the GPT-4.1's recommended technique in this PR. This `V4A` diff technique handles complex cases, makes the model *unlazy*, and gives better edits.

## Related Issue(s)

https://github.com/olimorris/codecompanion.nvim/discussions/1329

## WIP and Required Inputs

- There are a few edge cases. I am working on fixing them and adding more tests.
- I require your go-ahead about whether you are okay with this direction?

**Also, how do we see `print`s when using `child.lua` in tests?** I am having a hard time seeing errors and doing debugs.


olimorris:
Great stuff, thanks @pratyushmittal. Very much appreciate all of the thought and solid grounding in knowledge you've put into this.

My only concerns, which we'll know from testing, are will the lesser/older models be able to work with this new tool. If not, perhaps we have a `@files_legacy` tool...

But defo proceed with this and share your results. Great job on the tests too.

Minor point, I've used snake casing throughout CodeCompanion.

olimorris:
@pratyushmittal this looks incredible! Thanks for your hard work on this. I'm going to give this a try first thing my time tomorrow. Super excited for this

lua/codecompanion/strategies/chat/agents/tools/files.lua
```diff
@@ -326,28 +419,10 @@ return {
...
-          args.path,
-          file and file.filetype,
-          file and file.content
-        )
-      else
-        llm_output = output
-      end
-
-      chat:add_tool_output(self, llm_output, output)
+      local llm_output = vim.iter(stdout):flatten():join("\n")
```
olimorris, line 422: This is a minor point. If I get the `@files` tool to edit a file, in the buffer I see:

```
**Files Tool**: The UPDATE action for `edit.py` was successful
```

But if I press `gd` I see:

```
content = "Successfully updated edit.py",
```

So it's quite hard for me to visually reconcile what I see in the buffer with what I see in the message history. Could we make the output to the LLM `The UPDATE action for edit.py was successful` with the LLM too?

jinzhongjia:
Great, I'm looking forward to this pr merger

lua/codecompanion/strategies/chat/agents/tools/files.lua
```diff
@@ -4,154 +4,228 @@ This tool can be used make edits to files on disk.
...
-  local extracted = {}
-  local current_line = 0
+```%s
+%s
+```]], vim.fn.fnamemodify(p.filename, ":e"), p:read())
+  return output
+end
 
-  local lines = p:iter()
+local function get_new_change(focus, pre)
```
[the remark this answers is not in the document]
pratyushmittal: Done 👍🏽

lua/codecompanion/strategies/chat/agents/tools/files.lua
pratyushmittal: Done.

pratyushmittal:
Thanks @olimorris for testing it out. I have made the changes you suggested. I have also made few more small fixes.

meijieru:
Would this improve the experience with editor tool as well?

pratyushmittal:
> Would this improve the experience with editor tool as well?

Currently no. This PR is only for the `@files` tool. I can update `@editor` tool too on similar logic if this works well.

@olimorris should I create a similar PR for `@editor` tool too?

olimorris (approved):

ravitemer:
Does it make sense to merge both editor and files tools into one? To the current files tool we can add logic to open files in the editor while creating them, showing diff while editing them etc. Like how MCPHub's native neovim server's write_file, replace_file tools works?

olimorris:
@ravitemer put that in a new discussion. I'd like the community to be able to respond to that as well.

PR #1352, opened 2025-05-03
fix(reference): make sure ref with `opts.visible = false` are hidden.

Davidyz:
## Description

I was unable to hide references even if I added them with `opts = {visible = false}`.

Specifically, the extra test I added in this PR fails without this patch.

## Screenshots

image

## Checklist

- [x] I've added


olimorris:
Interesting you raise this. Something I spotted last night and I wasn't sure if these should be visible or if if they should be in `<file></file>` tags.

So thanks for the PR and the tests!

PR #1367, opened 2025-05-04
Fix/adapter resolution

ravitemer:
## Description

This is a very rare bug which only happens when `Chat.new()` is passed a adapter name e.g while restoring chats in the history extension.

When we create chat with `Chat.new({adapter = "adapter_name"})`, the
adapter is resolved with require which when failed used to be returned
as `config.adapters[name]` . When user defines the adapter as a function
this throws an error.

## Checklist

### DIFF
--- a/lua/codecompanion/http.lua
+++ b/lua/codecompanion/http.lua
@@ -150,17 +150,23 @@ function Client:request(payload, actions, opts)
           opts.status = "error"
         end
 
-        util.fire("RequestFinished", opts)
+        if not opts.silent then
+          util.fire("RequestFinished", opts)
+        end
         cleanup(opts.status)
         if self.user_args.event then
-          util.fire("RequestFinished" .. (self.user_args.event or ""), opts)
+          if not opts.silent then
+            util.fire("RequestFinished" .. (self.user_args.event or ""), opts)
+          end
         end
       end)
     end,
     on_error = function(err)
       vim.schedule(function()
         actions.callback(err, nil)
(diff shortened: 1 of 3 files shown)


ravitemer:
Olimorris, please don't mind me pushing some changes without discussing with you. Although I'm trying to push what's necessary, sometimes we might get carried away 😅 . Feel free to discard any of these.

olimorris:
@ravitemer you know the codebase exceptionally well so you don't need to discuss prior to a PR.

lua/codecompanion/http.lua
```diff
@@ -150,17 +150,23 @@ function Client:request(payload, actions, opts)
           opts.status = "error"
         end
 
-        util.fire("RequestFinished", opts)
+        if not opts.silent then
```
olimorris, line 153: Out of curiosity, what's your use case for adding in a silent option? I ask, because then I'll remember not to remove it in the future 😆.

ravitemer: Sorry mentioned it in the commit description. 


Useful for when user might want to send a custom request using
client.new().request() method but doesn't want the "Request*" events to
interfere with the core logic of codecompanion

In case of history extension when user submits his first message, we
send another request to generate title. For users who depend on
"Request*" events to show statusline progress the title request's
RequestFinished might cause progress status to stop before the actual
request is completed

olimorris: Oh that is a genius move. ❤️ that as an idea.

PR #1368, opened 2025-05-04
fix(diff): use mini_diff default if mini.diff is available

deas:
## Description

Disclaimer: Lua newbie here

I think we want diff to default to `mini_diff`/`mini.diff` when the latter  is available on the path. As is, this fails for me, and the attached patch fixes it.
 
## Checklist

### DIFF
--- a/lua/codecompanion/providers/init.lua
+++ b/lua/codecompanion/providers/init.lua
@@ -13,7 +13,7 @@ end
 ---Get the default Diff provider
 ---@return string
 local function diff_providers()
-  if pcall(require, "mini_diff") then
+  if pcall(require, "mini.diff") then
     return "mini_diff"
   else
     return "default"


lua/codecompanion/providers/slash_commands/init.lua
```diff
@@ -43,7 +43,7 @@ end
 local default_action_palette_provider = get_default_action_palette_provider()
 
 local function get_default_diff_provider()
-  if pcall(require, "mini_diff") then
+  if pcall(require, "codecompanion.providers.diff.mini_diff") then
```
olimorris, line 46: This will always resolve to true because CodeCompanion will always contain this module. This should be `mini.diff` as per [source](https://github.com/echasnovski/mini.diff?tab=readme-ov-file#installation)

olimorris:
Appreciate the PR and great spot!

lua/codecompanion/providers/slash_commands/init.lua
deas: Seen that bit. Actually wanted to trim [this similar looking condition](https://github.com/olimorris/codecompanion.nvim/blob/90e82abf4d65b64b0986a5be0981ba13e84eee8b/lua/codecompanion/providers/diff/mini_diff.lua#L7-L10) to use `require` directly, so we get expected behavior outside. Got that we may still want the latter when the user configures `mini_diff` directly.

Long story short: Using `mini.diff` in default condition now. It should be fine. ;)

olimorris:
I took a closer look at the providers logic. I missed that a PR put all of the provider resolving logic in `providers.slash_commands` when it should have been in `providers`. I've fixed that in ed54d02 now.

Could you resolve the conflicts in this PR and then I can merge.

You deserve the credit for spotting this and didn't want to implement your changes in ed54d02.

deas:
Feeling honored being new to Neovim and Lua - even for contributing only a single character change. ;)

olimorris:
You say "character change" I say "the determination to investigate a codebase and identify a bug" 😉.

Issue #1355, opened 2025-05-03
[Bug]: can't load references from custom prompt slash command


petobens:
### Did you check the docs and existing issues?

- [x] I have read all the docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched for existing issues/discussions
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

NVIM v0.12.0-dev-288+g6b233cd1a1

### Operating system/version

Arch Linux

### Describe the bug

If is set `is_slash_cmd = true` in a custom prompt that loads references and then call that prompt from a slash cmd the references are not loaded

### Steps To Reproduce

As in the GIF:
1. Open nvim with minimal init file below. The load reference `foo.py` file is simply something like
```python
"""Foo."""


def sum(x: float, y: float, z: float) -> float:
    """Sum."""
    return x + y + z


if __name__ == '__main__':
    print(sum(1, 2, 3))
```
2. Run `:CodeCompanionActions` and select the `Test References` custom prompt
3. Press `gd` to check that references were correctly loaded
4. Exit nvim and reopen it with the same init file
5. Run `:CodeCompanionChat` 
6. Insert `test_references` and press `<CR>` to execute the slash cmd
7. Press `gd` and check that references are not loaded now.
![Image](https://github.com/user-attachments/assets/4172374e-47a1-42a4-ad21-280f6abe2583)

### Expected Behavior

I expect references to load with the slash_cmd since the custom prompt defines those references

### init.lua file

```lua
local root = '/tmp/nvim-minimal'

-- Set stdpaths to use root dir
for _, name in ipairs({ 'config', 'data', 'state', 'cache' }) do
    vim.env[('XDG_%s_HOME'):format(name:upper())] = root .. '/' .. name
end

-- Bootstrap lazy
local lazypath = root .. '/plugins/lazy.nvim'
if not vim.loop.fs_stat(lazypath) then
    vim.fn.system({
        'git',
        'clone',
        '--filter=blob:none',
        '--single-branch',
        'https://github.com/folke/lazy.nvim.git',
        lazypath,
    })
end
vim.opt.runtimepath:prepend(lazypath)

-- Install plugins
local plugins = {
    {
        'hrsh7th/nvim-cmp',
        config = function()
            local cmp = require('cmp')
            cmp.setup({
                mapping = {
                    ['<C-n>'] = cmp.mapping(cmp.mapping.select_next_item()),
[...]

PR #1384, opened 2025-05-06
custom prompts slash commands can add references

petobens:
Hi @olimorris! This seems to (partially) address https://github.com/olimorris/codecompanion.nvim/issues/1355

There is one thing missing though: If I run two successive slash commands that both add references then the second slash cmd should clear the previous slash cmd reference before adding it's own right? Is there a simply way of removing references? (I left a TODO comment in the code)


petobens:
I removed the TODO. I guess we can manually remove the references when triggering a new slash command rather than adding some convoluted logic to do that automatically.

olimorris:
Give me some time to look into this.

Issue #552, opened 2024-12-20
[Bug]: Potential memory leak on longer conversations


GitMurf:
### Your `minimal.lua` config

Ran with nvim --clean but overall this is more of a FYI and curious whether you have experienced similar (or heard from others).

### Error messages

N/A

### Log output

N/A

### Health check output

N/A (everything is fine)

### Describe the bug

I have noticed the longer a conversation goes, the slower the code companion text input gets (choppy) and the memory consumption of nvim.exe rises continually with each message.

I am not talking humungous conversation, but also not small. Typically maybe 10-15 messages from me, but also pretty code heavy so responses are not short from LLM. The Code Companion buffer is around 700-1,000 lines (for the conversation) when it seems to start to feel more sluggish. And then it gets worse and worse from there.

In my most recent conversation this morning, nvim.exe balooned to 4GB of RAM! And the conversation was at 2,300 lines in the buffer FYI. When I quit/closed the conversation, the memory seemed to hang around even after opening a new chat buffer. It wasn't until I sent a new message in a new conversation that it paused (felt like it froze for a bit) in the middle of the response but then after waiting a bit it got responsive again and my nvim RAM consumption dropped back down to a normal level around 300MB. So it seems whatever "memory leak", it got cleared up after sending a new message to a new conversation after "deleting" / stopping / quitting the long/slow conversation buffer.

Let me know if you have any questions, ideas or thoughts on how I can provide more information / debug?

### Final checks

- [X] I have made sure this issue exists in the latest version of the plugin
- [X] I have tested with the `minimal.lua` file from above and have shared this
- [X] I have shared the contents of the log file

GitMurf:
Btw this has been happening for a while... basically as long as I can remember when doing longer conversations. Just more recently with the features to add references etc. it has been more valuable to have longer conversations so I have used it more.

Point is, I have no reason to believe this is anything recent that has caused a regression or something. I have noticed it at least for several weeks but I believe longer as I did not utilize longer conversations as much prior to the last month or so.

olimorris:
Can you share the adapter that you're using and any config changes you've made to it? Are you using a rendering plugin for any markdown files?

We could use an Agentic Workflow in the prompt library to easily replicate multiple successive prompts.

Off the top of my head, I can't fathom why after processing there would be so much memory usage. Unless something is getting attached to the event loop with every request.

olimorris:
I've created the following workflow that prompts the LLM 10 times for:

[...]

Issue #913, opened 2025-02-13
[Bug]: Incorrect Markdown breaks the chat window


f-leeske:
### Your `minimal.lua` config

```lua
---@diagnostic disable: missing-fields

--[[
NOTE: Set the config path to enable the copilot adapter to work.
It will search the following paths for a token:
  - "$CODECOMPANION_TOKEN_PATH/github-copilot/hosts.json"
  - "$CODECOMPANION_TOKEN_PATH/github-copilot/apps.json"
--]]
vim.env["CODECOMPANION_TOKEN_PATH"] = vim.fn.expand("~/.config")

vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

-- Your CodeCompanion setup
local plugins = {
  {
    "olimorris/codecompanion.nvim",
    dependencies = {
      { "nvim-treesitter/nvim-treesitter", build = ":TSUpdate", },
      { "nvim-lua/plenary.nvim", },
      -- Test with blink.cmp
      {
        "saghen/blink.cmp",
        lazy = false,
        version = "*",
        opts = {
          keymap = {
            preset = "enter",
            ["<S-Tab>"] = { "select_prev", "fallback", },
            ["<Tab>"] = { "select_next", "fallback", },
          },
          sources = {
            default = { "lsp", "path", "buffer", "codecompanion", },
            cmdline = {}, -- Disable sources for command-line mode
          },
        },
      },
      -- Test with nvim-cmp
      -- { "hrsh7th/nvim-cmp" },
    },
    opts = {
      --Refer to: https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/config.lua
      adapters = {
        openai = function()
          return require("codecompanion.adapters").extend("openai", {
            schema = {
              model = {
                default = "gpt-4o-mini",
              },
            },
          })
        end,
      },
      strategies = {
        --NOTE: Change the adapter as required
        chat = { adapter = "openai", },
        inline = { adapter = "copilot", },
      },
      opts = {
        log_level = "DEBUG",
      },
    },
  },
}

require("lazy.minit").repro({ spec = plugins, })

-- Setup Tree-sitter
local ts_status, treesitter = pcall(require, "nvim-treesitter.configs")
if ts_status then
  treesitter.setup({
    ensure_installed = { "lua", "markdown", "markdown_inline", "yaml", },
[...]

PR #1385, opened 2025-05-06
Broken markdown syntax should send proper user message; Allow dynamic user role

ravitemer:
## Description

The reason for this PR is to address a bug relating to treesitter parsing in the buffer. 

The issue was previously opened and was closed as not planned. https://github.com/olimorris/codecompanion.nvim/issues/913

While I was testing MCPHub's write tool, when user makes some changes to the LLM output we send a diff block of what's  changed to the LLM. As we enclose the result of the MCP tool inside triple ticks and the diff is also in the triple ticks this broke the markdown syntax of the buffer. (Although I fixed it in mcphub to enclose mcp tool response in 4 backticks, this issue will still happen in other situations)

image

When I tried to send something when the buffer is broken, instead of sending my message, it sent the initial prompt.

I thought this was fixed, and when searched through the issues I found this, 

image

I realized the issue was due to `ts_parse_messages` unable to form a proper tree thereby not recognizing the correct last user header even though we give it the header_line. 

On some thought, I felt we can do this without the ts parsing. I removed the `ts_parse_messages` and `ts_parse_headers` and replaced them with some simple functions. 

There are two scenarios where we are using ts to identify the user header_line. 

1. When `Chat.new()` is called with some messages, we are using `ts_parse_headers` to set `self.header_line`. When I loaded this broken conversation using history extension, the same issue happened. It set the header_line to 1. 
2. When we submit the message, as described above

For 1, get all the lines, from bottom up find the header line and set `self.header_line`

For 2, whenever we get a response from the llm even though the syntax is broken, the header_line is always set correctly using `set_text_editing_area(chat, modifier)`. As we know the last header line we just have to get all the lines after that line, remove empty lines and references and we get the user message. 

This seems to work very well  unless there is something I'm missing.

While working on this, I felt that resolving dynamic roles and setting the headers is related to the parsing of messages and headers. The current logic for resolving roles is confusing and it doesn't allow user role to be a function. The first commit allows user and llm roles to be dynamic. It resolves the roles when UI.new() is called and whenever a new_role() is rendered to the buffer while chatting. The resolved_roles property of UI class is used by everyone that needs the resolved roles which allows something like this

```lua
user = function(adapter)
    local time_elapsed = os.difftime(os.time(), time)
    local minutes = math.floor(time_elapsed / 60)
    local seconds = time_elapsed % 60
    return string.format("  Ravitemer " .. string.format("(%02d:%02d)", minutes, seconds))
end,
```
image

Also some minor bugs that I found along the way.

## Checklist

### DIFF
--- a/doc/codecompanion.txt
+++ b/doc/codecompanion.txt
@@ -522,7 +522,7 @@ LAYOUT ~
   [!NOTE] The Action Palette also supports Telescope.nvim
   <https://github.com/nvim-telescope/telescope.nvim>, mini.pick
   <https://github.com/echasnovski/mini.pick> and snacks.nvim
-  <https://github.com/folke/snacks.nvi>
+  <https://github.com/folke/snacks.nvim>
 You can change the appearance of the chat buffer by changing the
 `display.action_palette` table in your configuration:
 
@@ -1245,8 +1245,10 @@ customized in the configuration:
             end,
     
             ---The header name for your messages
-            ---@type string
-            user = "Me",
+            ---@type string|fun(adapter: CodeCompanion.Adapter): string
+            user = function(adapter)
+    			return vim.env.USER:gsub("^%l", string.upper)
+            end,
           }
         }
       }
@@ -1258,8 +1260,6 @@ By default, the LLM’s responses will be placed under a header such as
 This option can be in the form of a string or a function that returns a string.
 If you opt for a function, the first parameter will always be the adapter from
 the chat buffer.
-
-The user role is currently only available as a string.
 
 
 MARKDOWN RENDERING
@@ -2699,7 +2699,7 @@ OpenAI adapter.
   as a great reference to understand how they’re working with the output of the
   API
 
-OPENAI�S API OUTPUT
+OPENAI’S API OUTPUT
 
 If we reference the OpenAI documentation
 <https://platform.openai.com/docs/guides/text-generation/chat-completions-api>
--- a/doc/configuration/chat-buffer.md
+++ b/doc/configuration/chat-buffer.md
@@ -295,8 +295,10 @@ require("codecompanion").setup({
         end,
 
         ---The header name for your messages
-        ---@type string
-        user = "Me",
+        ---@type string|fun(adapter: CodeCompanion.Adapter): string
+        user = function(adapter)
+			return vim.env.USER:gsub("^%l", string.upper)
+        end,
       }
     }
   }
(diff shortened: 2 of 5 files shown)


ravitemer:
This broke adding references in `codecompanion.nvim/lua/codecompanion/strategies/chat/references.lua`

```lua
  local parsed_buffer = ts_parse_buffer(self.Chat)

  if parsed_buffer then
    -- If the reference block already exists, add to it
    if parsed_buffer.capture == "refs" then
      add(self.Chat, ref, parsed_buffer.end_row - 1)
    -- If there are no references then add a new block below the heading
    elseif parsed_buffer.capture == "role" then
      add(self.Chat, ref, parsed_buffer.end_row + 1)
    end
  end
```

Will fix it

olimorris:
Hey @ravitemer hold fire on this PR. I think it's unlikely I'll merge it.

I thought about this quite a bit in #913. I resolved that a fix would result in me having to change the chat buffer beyond what I'm comfortable with. There's a lot of reliability that comes with using Tree-sitter, from a performance and a parsing perspective. Given the solution is four backticks or simply not writing any output in to the chat buffer itself, I'd rather we focus on that than altering the Tree-sitter parsing.

There was a lot of work that I did to resolve #552 which saw the inclusion of much smarter buffer parsing over specific ranges etc. I don't want to put that in jeopardy for an issue which I think can be mitigated through other ways.

ravitemer:
I understand your point completely. And fine with closing the PR. I used to have this fear of chat being broken and the need to start prompting from the scratch, particularly while using xml tools. So wanted to have a crack at it. Anyway I might be missing the whole picture here. 

Hear me out though 😅,  

Apart from yaml parsing for settings and buffer parsing for code blocks to allow "gy", to my knowledge we are currently using treesitter for finding the last user role line in one way or other. But, If we can just find the last user line we can parse the user message, render references, check references more reliably without the fear of syntax breaking. Using ts for the former 2 is absolutely necessary. For the latter, I thought this approach might be fast, reliable:  getting last 100 lines -> check for the last header -> found? happy else -> get next 100 lines and so on. Treesitter must be doing so much more than this internally just to get the last header line. 

Anyways, putting it on hold for now then unless it becomes a pain point.

olimorris:
These types of changes are things I'm really wary of just because their potential impact is difficult to test and often only becomes apparent months later (#552 is the best example of that). The chat buffer is quite a unique space and everyone uses them completely differently. I treat some as a scratch buffer and then have some open for days that go on for 10,000+ lines. 

I always evaluate features/fixes against "# of users impacted" and complexity to build vs benefit to the plugin. Put simply, I'm about 18 months into the plugin and this issue (which has always existed) got raised once and I'm still convinced we can counter this by updating the system prompt and asking our LLMs to use the four backticks.

ravitemer:
Totally agree with you. Yes it is a rare one.

> I'm really wary of just because their potential impact is difficult to test and often only becomes apparent months later

Yes, 💯 understood! No worries. we can just tweak the system prompt.

PR #1387, opened 2025-05-06
refactor(providers): centralize provider selection logic

dhruvinsh:
## Description

- Introduce `ProviderSpec` and `Providers` types to define provider configurations.
- Add a unified `find_provider` function to handle provider selection based on availability and optional conditions.
- Refactor all provider selection functions (`action_palette_providers`, `diff_providers`, `help_providers`, `pick_providers`) to use the new logic, improving maintainability and extensibility.
- Improve logic for selecting Snacks as a provider

## Related Issue(s)
None, but key highligh is, as a part of the refactor it improve picker selection specially in case of `Snacks`. It is possible that picker might be disable and we should respect that.

### DIFF
--- a/lua/codecompanion/config.lua
+++ b/lua/codecompanion/config.lua
@@ -117,7 +117,7 @@ local defaults = {
           description = "Insert open buffers",
           opts = {
             contains_code = true,
-            provider = providers.pickers, -- telescope|snacks|mini_pick|fzf_lua|default
+            provider = providers.pickers, -- telescope|fzf_lua|mini_pick|snacks|default
           },
         },
         ["fetch"] = {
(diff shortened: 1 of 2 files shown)


dhruvinsh:
Hi, 
There is no change in the terms of the end table creation, I have added only one condition `return snacks_module and snacks_module.config.picker.enabled` for snacks.

and the end table still be the same,
```lua
return {
  action_palette = action_palette_providers(),
  diff = diff_providers(),
  help = help_providers(),
  pickers = pick_providers(),
}
```

Few question,
- What kind of actions were you trying?
- Mind sharing your config?

olimorris:
I don't have anything specified for providers and only have Snacks installed (with picker enabled).

CodeCompanion config: https://github.com/olimorris/dotfiles/blob/main/.config/nvim/lua/plugins/coding.lua

Snacks config: https://github.com/olimorris/dotfiles/blob/main/.config/nvim/lua/plugins/ui.lua

leiserfg:
If the pickers are not using previewers, won't it be simpler to use `vim.ui.select` which will simplify the implementation?
The only cons, is not having access to previews but I don't think they are used currently.

dhruvinsh:
It should be fixed now. Apologies for delay.

olimorris:
Thanks so much for this. Really awesome PR and cleans things up hugely!

PR #1399, opened 2025-05-09
Better handle missing tools

ravitemer:
### BUG 

When LLM calls a tool that is missing, subsequent prompts fail with `Bad Request` error due to missing message with `tool` role. 

To address this we return the error mentioning that the tool is not found along providing it the available_tools as tool call response, so that on next prompts LLM corrects the tool call.

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/agents/init.lua
+++ b/lua/codecompanion/strategies/chat/agents/init.lua
@@ -107,12 +107,34 @@ function Agent:execute(chat, tools)
 
     local name = tool["function"].name
     local tool_config = self.tools_config[name]
+    local function handle_missing_tool(tool_call, err_message)
+      tool_call.name = name
+      tool_call.function_call = tool_call
+      log:error(err_message)
+      local available_tools_msg = next(chat.tools.in_use or {})
+          and "The available tools are: " .. table.concat(
+            vim.tbl_map(function(t)
+              return "`" .. t .. "`"
+            end, vim.tbl_keys(chat.tools.in_use)),
+            ", "
+          )
+        or "No tools available"
+      self.chat:add_tool_output(
+        tool_call,
+        string.format("Tool `%s` not found. %s", name, available_tools_msg),
+        string.format("**%s Tool Error**: %s", name, err_message)
+      )
+      return util.fire("AgentFinished", { bufnr = self.bufnr })
+    end
+    if not tool_config then
+      return handle_missing_tool(vim.deepcopy(tool), string.format("Couldn't find the tool `%s`", name))
+    end
 
     local ok, resolved_tool = pcall(function()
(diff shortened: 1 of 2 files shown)


lua/codecompanion/strategies/chat/agents/init.lua
```diff
@@ -107,12 +107,31 @@ function Agent:execute(chat, tools)
...
     local name = tool["function"].name
     local tool_config = self.tools_config[name]
+    local function handle_missing_tool(tool_call, err_message)
+      tool_call.name = name
+      tool_call.function_call = tool_call
+      log:error(err_message)
+      self.chat:add_tool_output(
+        tool_call,
+        string.format(
+          "You made an error in calling the %s tool: %s\n\nAvailable tools: %s",
```
olimorris, line 117: Can we be more concise and say `Tool %s not found. The available tools are....`

olimorris:
Thanks @ravitemer. Was this something you noticed often? Could we add some test coverage for this too?

ravitemer:
Not frequent but I have seen instances where the LLM confuses between `use_mcp_tool` and tool name of MCP Server and instead of calling 

```
use_mcp_tool
  server_name
  tool_name_in_server
  arguments
  ```
  
  It calls 
  
  ```
  tool_name_in_server
     arguments
 ``` 
 
 leading to no tool response for the tool call in the messages history and we need to tweak in the "gd" view to remove the llm tool call message so that we can continue the chat without "Bad Request" error

**Comment by eli0shin** [comment]
Thank you! This has been driving me a bit nuts lately. I often just tag the mcp tool but have regularly needed to add files as well when this issue comes up.

File: lua/codecompanion/strategies/chat/agents/init.lua [hunk 2]
```diff
@@ -107,12 +107,34 @@ function Agent:execute(chat, tools)
...
+    if not tool_config then
+      return handle_missing_tool(vim.deepcopy(tool), string.format("Couldn't find the tool `%s`", name))
+    end
 
     local ok, resolved_tool = pcall(function()
       return Agent.resolve(tool_config)
     end)
     if not ok or not resolved_tool then
-      return log:error("Couldn't resolve the tool(s) from the LLM's response %s", resolved_tool)
+      return handle_missing_tool(vim.deepcopy(tool), string.format("Couldn't resolve the tool `%s`", name))
```
eli0shin, line 137: When I see this error it usually comes from the line above as a ‘cannot index property of nil’ error. Have you been seeing the errors throw with the Couldn’t resolve tools message?

ravitemer: I too get the 'attempt to index tool a nil value'. It is from Agent.resolve line above when tool_config is nil. So I added check for that as well

PR #1409, opened 2025-05-10
Allow to override gemini parameters and add reasoning effort

petobens:
This basically addresses my comments here https://github.com/olimorris/codecompanion.nvim/discussions/1337#discussion-8269597

1. Since the gemini adapter uses the openai endpoint then trying to pass parameters as
```lua
gemini_flash = function()
            return adapters.extend('gemini', {
                env = { api_key = GEMINI_API_KEY },
                schema = {
                    model = { default = 'gemini-2.5-flash-preview-04-17' },
                    maxOutputTokens = { default = 2048 },
                },
            })
        end,
```
obviously would not work since the Openai API expects `max_tokens` rather than `maxOutputTokens`.

2. Added a new `reasoning_effort` parameter that allows to turn thinking off for the flash model as per https://ai.google.dev/gemini-api/docs/openai#thinking

### DIFF
--- a/lua/codecompanion/adapters/gemini.lua
+++ b/lua/codecompanion/adapters/gemini.lua
@@ -69,7 +69,7 @@ return {
       desc = "The model that will complete your prompt. See https://ai.google.dev/gemini-api/docs/models/gemini#model-variations for additional details and options.",
       default = "gemini-2.5-flash-preview-04-17",
       choices = {
-        "gemini-2.5-flash-preview-04-17",
+        ["gemini-2.5-flash-preview-04-17"] = { opts = { can_reason = true } },
         "gemini-2.5-pro-exp-03-25",
         "gemini-2.0-flash",
         "gemini-2.0-pro-exp-02-05",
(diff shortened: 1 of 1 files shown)


petobens:
I also removed the frequency_penalty parameter since it doesn't seem to available in 2.5 (pro or flash) models

olimorris:
Thanks for this. Are there any other schema items that we're missing? I note that `presence_penalty` references "logprobs"

petobens:
Yep the penalty parameter is not enabled for 2.5 models:
```
[{
  "error": {
    "code": 400,
    "message": "Penalty is not enabled for models/gemini-2.5-flash-preview-04-17",
    "status": "INVALID_ARGUMENT"
  }
}
```
Do you want me to remove it all together?

Dunno if we are missing any other relevant item since the `GenerationConfig` api doc is a bit opaque regarding what parameter is available for a specific model: https://ai.google.dev/api/generate-content#v1beta.GenerationConfig
But I guess we can always add them later on if needed
/

olimorris:
> Yep the penalty parameter is not enabled for 2.5 models:
> Do you want me to remove it all together?

Yep let's remove that. I suspect it's a rarely often used param and anyone hardcore can roll their own adapter.

petobens:
Done

PR #1426, opened 2025-05-13
Allow adapters to modify http requests as needed

parkin:
## Description

Adds two handlers to give adapters more flexibility in crafting the http request.
- `set_form` - Function which lets the adapter modify the form data sent in the http request
- `modify_request_opts` - Function which lets the adapter modify the http request parameters before the request is sent

Without this, the http requests are limited to sending content in the body, and the encoding is json.
This works for most LLM's, but not for LLM's with different request requirements.

The two new handlers can enable codecompanion to work with LLM's with non-standard request requirements.
For example, this enables adapters to send form data, and also to alter the encoding of the body before the request is sent.

## Checklist

### DIFF
--- a/doc/codecompanion.txt
+++ b/doc/codecompanion.txt
@@ -1,4 +1,4 @@
-*codecompanion.txt*          For NVIM v0.10.0         Last change: 2025 May 10
+*codecompanion.txt*          For NVIM v0.10.0         Last change: 2025 May 12
 
 ==============================================================================
 Table of Contents                            *codecompanion-table-of-contents*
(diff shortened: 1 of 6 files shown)


olimorris:
Hey @parkin thanks for this and apologies for the slow reply.

I've given this some thought. I think this is such a unique use case that I don't want to commit this to `main`. I'm thinking, instead, that we should allow users to provide a link to a separate curl module instead of `plenary.curl`. This is far less burdensome on myself and potentially more flexible than relying on me to merge PRs.

I'll put that on my todolist.

parkin:
Sounds good - agreed that makes more sense. I'll give that a try as well.

PR #1455, opened 2025-05-18
feat(adapters): add vision support

olimorris:
## Description

This PR will add vision support to Copilot and OpenAI (possibly more to follow).

It will allow users to share images in three primary ways, via:

1. URLs
2. Filesystem
3. Clipboard

It will introduce an `image` slash command to enable 1 and 2 and leverage [img-clip.nvim](https://github.com/hakonharnes/img-clip.nvim) for 3.

### Supported Adapters

- [x] OpenAI
- [x] Copilot
- [x] Anthropic
- [x] Gemini

### Supported Providers

- [x] Snacks.nvim
- [x] Default

## Related Discussion(s)

#1105

### URLs

https://github.com/user-attachments/assets/05b4c84d-a376-4b54-ab16-1a859bb2d78f

### Files

https://github.com/user-attachments/assets/c50ca419-d686-4a37-8819-cd31e3e74446

## Checklist


olimorris:
If anyone wants to try this out, would be much appreciated. I'm having some issues with `pngpaste` from `img-clip.nvim` working. It was working fine yesterday. The configuration you'll need to get started with CodeCompanion is:

```lua
{
  "HakonHarnes/img-clip.nvim", -- Share images with the chat buffer
  event = "VeryLazy",
  cmd = "PasteImage",
  opts = {
    filetypes = {
      codecompanion = {
        prompt_for_file_name = false,
        template = "[Image]($FILE_PATH)",
        use_absolute_path = true,
      },
    },
  },
},
```

olimorris:
> Thank you! I really like the way how we are using the content field for data keeping it to a string, without major refactoring. Also parsing inline images is super sick.

That was the plan to make it work nicely with CodeCompanion History. Then the adapters can do the forming of the messages into the correct structure.

> The only issue I faced was with the file choice picker which isn't updated to to use providers.images

That was intended because I'm not sure many users will have enabled that snacks module. You should still be able to view the images though. Are you facing issues with that? And there will be config options to load specific directories into the picker. I have a screenshots folder for example that sits in my Documents where all screenshots go. I wanted that to be always available via the slash command.

ravitemer:
> That was the plan to make it work nicely with CodeCompanion History. Then the adapters can do the forming of the messages into the correct structure.

Yeah, totally forgot. It just works with the history extension ( ♥ )

> Are you facing issues with that?
 
I am getting the following error if we don't explicitly set it to snacks: 
```txt
   Error  14:55:49 notify.error CodeCompanion [ERROR] 2025-05-20 14:55:49
Provider for the symbols slash command could not be found: telescope
```
with this config
```lua
		slash_commands = {
			["image"] = {
				opts = {
					dirs = { "/home/ubuntu/" }, -- Directories to search for images
					-- picker = "snacks"
				},
			},
		},
```

olimorris:
That would be because I haven't set Telescope support and I believe Telescope takes precedence over Snacks...but I would expect it to fallback to the default setting...hmmm. I'll investigate

olimorris:
@ravitemer the latest updates might have fixed this.

ravitemer:
Not unless we explicit set provider to snacks or change update the image slash_command default config to use `providers.images`. On exploring further:

In `slash_commands/image.lua` file:

```lua
local providers = {
  default = function(SlashCommand)
  end,

  snacks = function(SlashCommand)
  end,
}
---
  File = function(SlashCommand, SlashCommands)
    return SlashCommands:set_provider(SlashCommand, providers)
  end,
```

In `slash_commands/init.lua` file: 
```lua
function SlashCommands:set_provider(SlashCommand, providers)
  if SlashCommand.config.opts and SlashCommand.config.opts.provider then
    if not providers[SlashCommand.config.opts.provider] then
      return log:error(
        "Provider for the symbols slash command could not be found: %s",
        SlashCommand.config.opts.provider
      )
    end
    return providers[SlashCommand.config.opts.provider](SlashCommand) --[[@type function]]
  end
  return providers["default"] --[[@type function]]
end
```

As the `provider = providers.pickers` which resolves to `telescope` in the default image slash command config:

```lua
        ["image"] = {
          callback = "strategies.chat.slash_commands.image",
          description = "Insert an image",
          opts = {
            dirs = {}, -- Directories to search for images
            filetypes = { "png", "jpg", "jpeg", "gif", "webp" }, -- Filetypes to search for
            provider = providers.pickers, -- snacks|default
          },
        },
```

Not a major issue at this point though. I think you are eventually going to change the provider to `providers.images` once into production?

ravitemer:
I got it working with MCP Hub. I am using Stagehand MCP server for this. 

https://github.com/user-attachments/assets/a939be1e-7a77-495b-9727-54f110ec68b5

Just a note:

As there is no way to add the image using `add_tool_output` yet for images, I am adding images just as normal user messages. However, as we have to always send a tool output for a tool call to avoid "Bad Request" error, I am using a hack for now which works perfectly fine.

`````lua
            if result.images and #result.images > 0 then
                ---When the mcp call returns just images, we need to add the tool output
                if not tool_call_result_added then
                    local to_llm = string.format(
                        [[**`%s` Tool**: Returned the following:

````
%s
````]],
                        action_name,
                        string.format("%d image%s returned", #result.images, #result.images > 1 and "s" or "")
                    )
                    add_tool_output(action_name, self, agent.chat, to_llm, false, has_function_calling, opts)
                end
                -- local helpers = require("codecompanion.strategy.chat.helpers")
                for _, image in ipairs(result.images) do
                    local id = "<image>" .. string.format("mcp-%s", os.time()) .. "</image>"
                    -- helpers.add_image(agent.chat, result.images)
                    local config = require("codecompanion.config")
                    if image.data then
                        agent.chat:add_message({
                            role = config.constants.USER_ROLE,
                            content = image.data,
                        }, {
                            reference = id,
                            mimetype = image.mimeType,
                            tag = "image",
                            visible = false,
                        })

                        agent.chat.references:add({
                            bufnr = image.bufnr,
                            id = id,
                            --FIX:
                            path = image.path,
                            source = "codecompanion.strategies.chat.slash_commands.image",
                        })
                    end
                end

`````

**Comment by olimorris** [comment]
I've added a helper to `codecompanion.strategies.chat.helpers` called `add_image` that can take a filepath and base64 encode it. However I could add another one which makes it easier to add an image into the chat message structure. If that's helpful?

**Comment by ravitemer** [comment]
> I've added a helper to `codecompanion.strategies.chat.helpers` called `add_image` that can take a filepath and base64 encode it. However I could add another one which makes it easier to add an image into the chat message structure. If that's helpful?

I tried the helper but MCP responses are returning non-text data as direct base64 data.

Yeah `chat:add_image` will help if any future changes occur.

**Comment by olimorris** [comment]
> I tried the helper but MCP responses are returning non-text data as direct base64 data.

I don't know how we work around that as all LLM endpoints that I'm working with require it in base64

**Comment by ravitemer** [comment]
> > I tried the helper but MCP responses are returning non-text data as direct base64 data.
> 
> I don't know how we work around that as all LLM endpoints that I'm working with require it in base64

Sorry for not being clear enough. MCP protocol returns the image (non-text) as `base64 string` and `mimeType` which is exactly what we need.

The reason I said I `"tried"` is because the `helpers.add_image` expects `image.path` when we have the `base64` data already. 

I was thinking, If we can have a function `Chat:add_image({id,data,mimeType})` as all LLMs expects base64 data. `helpers.add_image` can also use this after convering path to base64.

**Comment by olimorris** [comment]
Ahh gotcha. Yes perfect. I'll add that in.

**Comment by olimorris** [comment]
@ravitemer I've added this in 👍🏼

**Comment by ravitemer** [comment]
@olimorris The copilot adapter's custom "Copilot-Vision-Request" header is gone from the latest commit. It is using openai adapter's form_messages instead. It was present in f31f324c480e95dcc3ba6bc741148ff91e6c5fcb commit

**Comment by olimorris** [comment]
I see it in my output:

```
{ "-sSL", "-D", "/tmp/plenary_curl_4fd740d0.headers", "-X", "POST", "-H", "Authorization: Bearer ABC123", "-H", "Editor-Version: Neovim/0.12.0", "-H", "Copilot-Integration-Id: vscode-chat", "-H", "Content-Type: application/json", "-H", "Copilot-Vision-Request: true", "-d", "@/var/folders/v2/78qsgrqj3c19g1z0vlsxjb3c0000gn/T/nvim.Oli/vppEs6/2.json", "--retry", "3", "--retry-delay", "1", "--keepalive-time", "60", "--connect-timeout", "10", "--tcp-nodelay", "--no-buffer", "https://api.githubcopilot.com/chat/completions" }
```

You'll get a big warning from them if it's not present:

```
Error: {"error":{"message":"missing required Copilot-Vision-Request header for vision requests","code":""}}
```

ravitemer:
Oops! Just saw that. I had a custom adapter that extends `copilot` and is declared as `copilot_custom`. As self.name becomes `copilot_custom` it is not adding the header.

olimorris:
Ahhhh. Okay. That's a fair point. I'll amend that and add it back to copilot's form_messages.

ravitemer:
Thought this is cool, MCP Hub allows previewing images in the chat if there are images in the response. 

If one has image.nvim plugin they can see images right inside the chat. Otherwise we have to remap clear chat to something other than `gx`

https://github.com/user-attachments/assets/70181790-e949-4df6-a690-c5d7a212e7d1

PR #1459, opened 2025-05-19
fix: user-extended adapters inheriting name

ravitemer:
## Description

Related this issue comment: https://github.com/ravitemer/codecompanion-history.nvim/issues/9#issuecomment-2885801123

When user declares some custom adapters like below, 
```lua
return {
	adapters = {
		openrouter = function()
			return require("codecompanion.adapters").extend("openai_compatible", {
				env = {
					url = "https://openrouter.ai/api",
					api_key = "OPENROUTER_API_KEY",
					chat_url = "/v1/chat/completions",
				},
				schema = {
					model = {
						default = "google/gemini-2.5-flash-preview",
					},
				},
			})
		end,
	},
}
```

With `Chat:new()` the adapter is resolved. But the `adapter.name` will be `openai_compatible` instead of `openrouter`. Most users don't provide name or provide a different name from the `config.adapters.[adapter_name]`. This leads to an issue when we need to restore the chat with the saved adapter name. In the above case the Chat will be initialized with `openai_compatible` adapter.

Resetting `adapter.name` while resolving it ensures adapter's name will be one of the `config.adapters`

## Checklist

- [x] I've read the [contributing](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md) guidelines and have adhered to them in this PR
- [ ] I've updated `CodeCompanion.has` in the [init.lua](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/init.lua#L239) file for my new feature
- [ ] I've added [test](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md#testing) coverage for this fix/feature
- [ ] I've updated the README and/or relevant docs pages
- [x] I've run `make all` to ensure docs are generated, tests pass and my formatting is applied

PR #1461, opened 2025-05-20
fix: adapters.resolve not using custom adapter, bug introduced with (#1459)

ravitemer:
## Description

Fixes bug as commented in https://github.com/olimorris/codecompanion.nvim/pull/1459#issuecomment-2892619215
 
image

## Checklist

- [] I've updated the README and/or relevant docs pages

### DIFF
--- a/lua/codecompanion/adapters/init.lua
+++ b/lua/codecompanion/adapters/init.lua
@@ -321,9 +321,8 @@ function Adapter.resolve(adapter)
   if type(adapter) == "table" then
     adapter = Adapter.new(adapter)
   elseif type(adapter) == "string" then
-    local name = adapter
-    adapter = Adapter.extend(adapter)
-    adapter.name = name
+    local user_adapter = config.adapters[adapter]
+    adapter = Adapter.extend(user_adapter or adapter, { name = adapter })
   elseif type(adapter) == "function" then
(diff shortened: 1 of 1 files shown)


kizza:
Cheers @ravitemer - I can confirm that works 🥳.   I actually thought it mightn't... given the return type of my config is `function` (but then `extend` [supports functions](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/adapters/init.lua#L280-L281) also).  

Noticing the `opts` there, I'm curious if you considered passing the `name` in via those? (or if that even works the way I think it might etc)

```lua
local name = "foo"
adapter = Adapter.extend(adapter)
adapter.name = name
```
with
```lua
adapter = Adapter.extend(adapter, { name = adapter })
```

ravitemer:
Thank you @kizza. That's much cleaner. Updated.

olimorris:
Thanks @kizza. We'll add some test coverage for this to make sure we have it can't happen again

kizza:
Sounds great.  I actually had a brief pass looking to write some... my mini-test-foo is quite lacking 🤔

I'll take the opportunity to say this is a _super_ exciting plugin.  Very grateful for the vision and hard work here 🙌✨

PR #1457, opened 2025-05-19
Feat/allow custom curl libs

olimorris:
## Description

Allow user's to provide custom http clients instead of `plenary.curl`. Update your config to be:

```lua
require("codecompanion").setup({
  adapters = {
    opts = {
      curl = "/Users/Oli/.dotfiles/.config/nvim/lua/curl.lua",
    },
  }
})
```

## Related Issue(s)

#1426

## Checklist

- [x] I've read the [contributing](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md) guidelines and have adhered to them in this PR
- [x] I've updated `CodeCompanion.has` in the [init.lua](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/init.lua#L239) file for my new feature
- [ ] I've added [test](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md#testing) coverage for this fix/feature
- [ ] I've updated the README and/or relevant docs pages
- [ ] I've run `make all` to ensure docs are generated, tests pass and my formatting is applied

PR #1469, opened 2025-05-21
set Adapter.opts.request to override Client.request

parkin:
## Description

Allows the adapter to implement a custom request protocol by overriding `Client:request`. Update the config to:

```lua
  adapters = {
    opts = {
      ---Send a HTTP request
      ---@param payload { messages: table, tools: table|nil } The payload to be sent to the endpoint
      ---@param actions CodeCompanion.Adapter.RequestActions
      ---@param opts? table Options that can be passed to the request
      ---@return table|nil The Plenary job
      request = function(client, payload, actions, opts)
          -- implement custom request logic here
      end
    },
  }
})
```

## Related Issue(s)

Above is another attempt at overriding the default http request logic - let me know what you think.

Here are the prior attempts for reference:
- #1457
- #1426 

## Checklist

### DIFF
--- a/lua/codecompanion/adapters/init.lua
+++ b/lua/codecompanion/adapters/init.lua
@@ -30,6 +30,7 @@ local log = require("codecompanion.utils.log")
 ---@field handlers.on_exit? fun(self: CodeCompanion.Adapter, data: table): table|nil
 ---@field handlers.teardown? fun(self: CodeCompanion.Adapter): any
 ---@field schema table Set of parameters for the generative AI service that the user can customise in the chat buffer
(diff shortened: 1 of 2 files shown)


olimorris:
This is _super_ clean, really like it and thanks for your efforts again. I'll merge this right away.

Are there any additional methods we need to attach to the `Client` object that your adapters will need to make use of? Everything right now is buried in `Client:request`.

parkin:
Great - thank you so much!
This is working well so far for my use case.

Agreed - maybe in the future we'll want to override some additional Client methods.
I'll test this out for now as is and maybe come back with a pull request for some tweaks if needed (or for an update to the docs).

PR #1472, opened 2025-05-21
fix(providers): only use `mini.diff` if initialized

toupeira:
## Description

If `mini.nvim` is installed CodeCompanion will try to use `mini.diff`, but it only works if `setup()` was called.

This changes the default so it's only used if initialized, and updates the docs to share how to use MiniDiff only with CodeCompanion.

## Checklist

- [-] I've updated `CodeCompanion.has` in the [init.lua](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/init.lua#L239) file for my new feature
- [-] I've added [test](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md#testing) coverage for this fix/feature

### DIFF
--- a/doc/installation.md
+++ b/doc/installation.md
@@ -152,6 +152,23 @@ To enable you to copy images from your system clipboard into a chat buffer via `
 
 ```
 
+### [mini.diff](https://github.com/echasnovski/mini.diff)
+
+To render generated diffs inline, but keep `mini.diff` disabled for Git diffs:
+
+```lua
+ {
+    "echasnovski/mini.diff",
+    config = function()
+      local diff = require("mini.diff")
+      diff.setup({
+        -- Disabled by default
+        source = diff.gen_source.none(),
+      })
+    end,
+},
+```
+
 ## Completion
 
 When in the [Chat Buffer](usage/chat-buffer/index), completion can be used to more easily add [variables](usage/chat-buffer/variables), [slash commands](usage/chat-buffer/slash-commands) and [tools](usage/chat-buffer/agents). Out of the box, the plugin supports completion with both [nvim-cmp](https://github.com/hrsh7th/nvim-cmp) and [blink.cmp](https://github.com/Saghen/blink.cmp). For the latter, on version <= 0.10.0, ensure that you've added `codecompanion` as a source:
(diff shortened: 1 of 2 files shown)


olimorris:
I don't really want to configure external plugins in CodeCompanion. What will this do to users who already have mini.diff setup?

Agree we should better instruct users about how to set up mini.diff for CodeCompanion but that should exist in the docs.

toupeira:
> I don't really want to configure external plugins in CodeCompanion. What will this do to users who already have mini.diff setup?

@olimorris nothing should happen, either `vim.g.minidiff_disable` is restored or it was never set at all.

I agree that calling `.setup()` is probably too intrusive, but temporarily disabling it when the user wants to use it with CodeCompanion seems fine to me.

For some context: I use `gitsigns.nvim` which is more full-featured, but like many others I have `mini.nvim` installed and so when I first tried CodeCompanion the diffing didn't work at all (because I was missing the `.setup()` call).

I think a good middle ground would be to:

- Only auto-enable the `mini_diff` provider in this plugin if MiniDiff is explicitly setup (= `_G.MiniDiff` is defined).
- Apply the change in this PR so the provider can still be used explicitly, like in my setup.
- Add a quick description somewhere in the docs, I wasn't sure what a good place is but if you have a suggestion I'd be happy to add that.

What do you think?

toupeira:
Oh I totally missed that in the docs, that works great too! :+1: 

I've reverted the previous changes, so now this PR changes the default to only use MiniDiff if initialized, and documents the setup you shared.

toupeira:
> I don't see why users, if this is particularly bothersome, wouldn't just set providers.diff = "default" in their config.

Just to clarify, previously if you had mini.nvim installed without using MiniDiff, diffing was broken in CodeCompanion (no error, just no diff appears at all). This PR fixes that so it falls back to Vim diffing in that case.

olimorris:
> Just to clarify, previously if you had mini.nvim installed without using MiniDiff, diffing was broken in CodeCompanion (no error, just no diff appears at all). This PR fixes that so it falls back to Vim diffing in that case.

This is much cleaner thanks for this and pushing this through.

toupeira:
Awesome, thank you too! :bow:

PR #1473, opened 2025-05-21
Fix/make params from schema on adapter resolve

ravitemer:
## Description

When chat is initialized with some adapter, the adapter doesn't have parameters field on it as `map_schema_to_params` is not called until `Chat:submit()`. I don't know if this is intentional.

This leads to a minor bug when users have a dynamic llm role. 

```lua
		roles = {
			llm = function(adapter)
				return string.format(
					"  %s%s",
					adapter.formatted_name,
					adapter.parameters and (adapter.parameters.model and " (" .. adapter.parameters.model .. ")" or "")
						or ""
				)
			end,
``` 
This is not noticed during regular workflows as llm header is only render after `Chat:submit()` which is not the case when using with history extension where we render all the headers on `Chat:new()`. Related issue from history extension: https://github.com/ravitemer/codecompanion-history.nvim/issues/21

This PR calls `map_schema_to_params` on `adapter.resolve()` so that `adapter.parameters` becomes available. Adds a test and fixes a case in `test_inline.lua` file as described in the commit.

I am not sure if this leads to any side effects, feel free to discard it.

## Checklist

### DIFF
--- a/lua/codecompanion/adapters/init.lua
+++ b/lua/codecompanion/adapters/init.lua
@@ -326,7 +326,7 @@ function Adapter.resolve(adapter)
   elseif type(adapter) == "function" then
     adapter = adapter()
   end
-
+  adapter:map_schema_to_params()
   return adapter.set_model(adapter)
 end
 
--- a/tests/strategies/inline/test_inline.lua
+++ b/tests/strategies/inline/test_inline.lua
@@ -10,7 +10,7 @@ T["Inline"] = new_set({
     pre_case = function()
       inline = h.setup_inline({
         adapters = {
-          fake_adapter = { name = "fake_adapter" },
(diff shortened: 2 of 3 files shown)


olimorris:
You should be able to get the model name from either `adapter.model` or `adapter.schema.model.default`. There _was_ a reason I don't set params until submit and I'm not sure I've got much test coverage to be confident this won't break it.

ravitemer:
Yeah, agree. Feels dangerous. 

I think the llm role was copied from a discussion way back. 
```lua
				return string.format(
					"  %s%s",
					adapter.formatted_name,
					" (" .. (adapter.schema.model.default or "unknown") .. ")"
				)
```

This works fine! Thank you!

PR #1486, opened 2025-05-22
fix openai vision support

petobens:
AFAICT the boolean was always set to false preventing vision feature to work with the openai adapter


olimorris:
Oooof I missed that! Thanks

PR #1503, opened 2025-05-26
Fix word boundary matching for tools and variables

ravitemer:
## Problem

I faced this issue when I am trying to make each MCP server into a tool groups and each tool in the server as codecompanion function tools for finer control and precision while keeping the old `@mcp`. MCP Hub has a `mcphub` server. Now when I create the individual tools, and type `@` in the buffer I get these 4. 

image
 
The current tool (`@tool`) and variable (`#variable`) detection system has issues with partial matching that affect both functionality and visual feedback:

### 1. Syntax Highlighting Partial Matches

image

When a user has both `@mcp` and `@mcphub` tools configured and types `@mcp @mcphub` in the buffer, the original syntax pattern `vim.cmd.syntax('match CodeCompanionChatTool "@' .. name .. '"')` matches any occurrence of the tool name substring. This causes only the `@mcp` portion within `@mcphub` to be highlighted, while the complete `@mcphub` tool name doesn't get its own distinct highlighting due to lack of proper word boundary detection.

### 2. Tool Detection with Word Boundaries  

image

When a user types `@mcphub__toggle_mcp_server`, the system incorrectly detects `@mcphub` as a valid tool because the underscore character (`_`) is treated as a word boundary by the current pattern matching. This leads to `@mcphub` tool group tools also being added when the user intended to use `mcphub_toggle_mcp_server` that simply contains the tool name as a prefix.

### 3. Pattern Matching Inconsistencies
The detection logic uses Lua pattern matching with `%f[%w@]@tool%f[%W]`, while syntax highlighting uses simple substring matching without word boundaries. These different approaches handle word boundaries inconsistently, creating discrepancies between what gets detected functionally versus what gets highlighted visually.

## Solution Design

image

image

### 1. Pattern Selection Strategy

**For Logic** (`agents/init.lua` and `variables/init.lua`):
```vim
\(^\|\s\)@tool\(\s\|$\)        # Tools
\(^\|\s\)#var\(\s\|$\|{\)      # Variables (with parameter support)
```

**For Syntax Highlighting** (`plugin/codecompanion.lua`):
```vim
\(^\|\s\)@tool\>               # Tools  
\(^\|\s\)#var\>                # Variables
```

### 2. Why Different Patterns?

**Logic uses `\(\s\|$\)`**: 
- ✅ Precise control over what constitutes a word boundary
- ✅ Intentionally restrictive (only spaces/end-of-line)
- ✅ Prevents accidental triggering in contexts like `@tool.` or `@tool!`

**Syntax uses `\>`**:
- ✅ Zero-width assertion (doesn't consume characters)
- ✅ Prevents overlapping match conflicts in `@mcp @mcphub`
- ✅ Efficient for real-time syntax highlighting

### 4. Character Consumption Problem

The key insight: **syntax highlighting has overlapping match issues**.

With `\(\s\|$\)` in syntax:
```
@mcp @mcphub
^^^^^        First match consumes the space after @mcp
     ^^^^^^^ Second match can't find required preceding space
```

With `\>` in syntax:
```
@mcp @mcphub  
^^^^          First match - \> is zero-width
     ^^^^^^^ Second match - space still available
```

### Agents (`lua/codecompanion/strategies/chat/agents/init.lua`)
```lua
local function is_found(tool)
  local pattern = "\\(^\\|\\s\\)" .. CONSTANTS.PREFIX .. tool .. "\\(\\s\\|$\\)"
  local regex = vim.regex(pattern)
  return regex and regex:match_str(message.content) ~= nil
end
```

### Variables (`lua/codecompanion/strategies/chat/variables/init.lua`)
```lua
local pattern = "\\(^\\|\\s\\)" .. CONSTANTS.PREFIX .. var .. "\\(\\s\\|$\\|{\\)"
local regex = vim.regex(pattern)
if regex:match_str(message.content or "") then
  table.insert(found, var)
end
```

### Syntax (`plugin/codecompanion.lua`)
```lua
vim.cmd.syntax('match CodeCompanionChatTool "\\(^\\|\\s\\)@' .. name .. '\\>"')
```

### Before (Broken)
```
Input: "@mcphub"
❌ Incorrectly matches: @mcp tool group (partial match)

Input: "@mcp @mcphub" 
❌ Syntax highlighting: only @mcp highlighted (overlapping issue)
```

### After (Fixed)
```
Input: "@mcphub"
✅ Correctly matches: @mcphub only
❌ No longer matches: @mcp (prevents false positive)

Input: "@mcp @mcphub"
✅ Logic: both tools detected correctly  
✅ Syntax: both tools highlighted correctly
```

## Edge Cases Handled

1. **Similar prefixes**: `@tool_group` vs `@tool_group_tool`
2. **Special characters**: `#foo://10-20-30:40` vs `#foo`
3. **Parameter support**: `#var{param}` detection
4. **Punctuation boundaries**: `@tool.` doesn't trigger (by design)
5. **Multiple tools**: `@mcp @mcphub` both work correctly

## Testing Strategy

- Added comprehensive test cases for word boundary scenarios
- Updated test configuration with conflicting tool/variable names
- Verified syntax highlighting behavior with visual tests
- Updated reference screenshots to reflect correct highlighting

## Backward Compatibility

**Logic**: ✅ Fully compatible - more precise matching prevents false positives
**Syntax**: ✅ Compatible - improved highlighting accuracy
**Performance**: ✅ `vim.regex()` is efficient for the use case

## Design Trade-offs

| Aspect | Trade-off | Decision |
|--------|-----------|----------|
| **Precision vs Flexibility** | Restrictive `\(\s\|$\)` vs permissive patterns | Chose precision to prevent false positives |
| **Consistency vs Performance** | Same pattern everywhere vs optimized per use case | Chose optimized patterns (different for logic vs syntax) |
| **Simplicity vs Correctness** | Simple Lua patterns vs complex Vim regex | Chose correctness with `vim.regex()` |

This approach ensures that tools and variables with similar names can coexist without conflicts while maintaining efficient syntax highlighting and precise logical detection.

## Checklist


ravitemer:
It's totally up to you.  I wanted to make it work without changing the current behavior. Typing `@{tool_name}` or `#{resource://uri}`  isn't that bad though.

> they're so brittle and so fickle and people have a surprising talent of nearly always being able to break them.

Agree but I doubt users intentionally declare special chars in their custom tools or variables names that the current testing couldn't catch. Only issue rare occasion that I can think of is if some other extension or plugin dynamically adding tools or variables with weird names that the current regex couldn't match.

I'm happy anyway as long as I can use groups and tool names that start with same prefix. 😅

lua/codecompanion/strategies/chat/agents/init.lua
```diff
@@ -188,6 +188,10 @@ function Agent:execute(chat, tools)
   end)
 end
 
+function Agent:_pattern(tool)
```
olimorris, line 208: Can we add a doc block with params and return types to this?

lua/codecompanion/strategies/chat/agents/init.lua
```diff
@@ -201,7 +205,11 @@ function Agent:find(chat, message)
   local tools = {}
 
   local function is_found(tool)
```
olimorris, line 226: Can we add a doc block with params and return types to this?

lua/codecompanion/strategies/chat/variables/init.lua
```diff
@@ -69,6 +70,13 @@ function Variables.new()
   return self
 end
 
+function Variables:_pattern(var)
```
olimorris, line 76: Can we add a doc block with params and return types to this?

lua/codecompanion/strategies/chat/variables/init.lua
```diff
@@ -79,9 +87,10 @@ function Variables:find(message)
 
   local found = {}
   for var, _ in pairs(self.vars) do
-    -- Escape the special characters in var
-    local escaped_var = var:gsub("[%(%)%.%%%+%-%*%?%[%]%^%$%/:]", "%%%1")
-    if message.content:match("%f[%w" .. CONSTANTS.PREFIX .. "]" .. CONSTANTS.PREFIX .. escaped_var .. "%f[%W]") then
+    -- Use vim.regex with explicit word boundaries
```
olimorris: We're repeating a lot of comments here. Let's remove them. Someone reviewing this code will navigate to `regex_find` and be able to understand what is happening

lua/codecompanion/utils/init.lua
olimorris, line 1: Can we move this logic into its own file called `utils/regex.lua`? Given these will be used infrequently compared to `utils/init.lua` and are super specific, they warrant their own file.

olimorris (changes requested):
Sorry for the delay in reviewing this. I think we're good to merge this soon.

A few comments regarding doc blocks, types etc and we don't need to be too generous with the comments in the files.

Thanks for your hardwork on this.

ravitemer:
I am done from my side. I have also updated the inline strategy variable find and replace logic. Please let me know if you need any more changes.

olimorris:
Noticed a small UI quirk this morning:

2025-06-07 10_44_41 - WezTerm@2x

It's highlighting the space after `#buffer`.

ravitemer:
Sorry didn't notice them due to my colortheme. Should be fixed now.

olimorris:
Really appreciate your efforts on this. Sorry, took me sometime to get my head into the change and the code. But super valuable addition with no overhead for the user. Great job.

olimorris:
@ravitemer heads up, I've just noticed that `Can we add a new method to #buffer with @insert_edit_into_file?` doesn't get parsed properly

ravitemer:
@olimorris Is this due to `?` at the end? If so, it was intentional.

There should be space after the tool or variable name or it must at the end of line. This was necessary in order to handle special characters in the names. As long as we can see the tool name or variable name highlighted, it should be parsed properly. 

Edit: Image

image

olimorris:
Cool okay and that makes sense!

tristan957:
I think having to add a space between the tool/variable and the punctuation is not great. Would love to see the bracket solution discussed earlier which is also how bash and zsh solve the problem.

Adding a space between a word and a punctuation mark is not great at all, and I'm sure people will run into this issue continually.

> It's totally up to you. I wanted to make it work without changing the current behavior. Typing @{tool_name} or #{resource://uri} isn't that bad though.

My understanding of the language server protocol is that a user can type something like `@tool_name`, and if they use the autocompletion option, the language server client can expand the selected entry into whatever the language server wants. So if a user typed `@tool`, codecompanion would tell the client to expand `@tool_name` to `@{tool_name}`.

I also evaluated how Cursor solves this problem, and it moves the context above the prompt, so it doesn't really have the same problem at all. In Cursor if I start typing `@...`, I can't even continue with my prompt until I choose one of the dropdown options provided by Cursor. Implementing the bracketed approach would be much more similar to Cursor in that regard.

olimorris:
@tristan957 - I echo your thoughts. It's becoming an inconvenience to myself too.

@ravitemer - I think moving to the `@{tool_name}` solution is the best way forward

tristan957:
@ravitemer definitely don't want to diminish your work in any way. Thanks for your contributions to CodeCompanion :D. Even recognizing the original issue at hand is a great first step!

ravitemer:
@tristan957 No worries! Thanks for the kind words. It was definitely not that clean to add a space between punctuation. Will open a PR in some time given this might be a major refactor mainly updating the tests.

>  So if a user typed @tool, codecompanion would tell the client to expand @tool_name to @{tool_name}.

Agree

PR #1516, opened 2025-05-28
feat(variables): support default params in configuration

toupeira:
## Description

This allows setting default parameters for variables in the global configuration, for example to auto-watch buffers.

## Related Issue(s)

Follow-up from https://github.com/olimorris/codecompanion.nvim/discussions/655#discussioncomment-13300545

## Checklist

- [-] I've updated `CodeCompanion.has` in the [init.lua](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/init.lua#L239) file for my new feature


toupeira:
@olimorris on second thought, I figured we can just add a `default_params` setting for all variables, instead of adding dedicated settings for `watch` and `pin` on buffers.

This turned out to be even easier, what do you think? If you like the approach I'll update the docs, and will see if I can also add a simple test.

toupeira:
I saw there's more variables in the `inline` strategy, but they don't seem to implement params? And at least for buffers I don't have a need to `watch` there anyway, since it's just one-shot :grinning:

olimorris:
Thanks so much @toupeira, this is an epic PR! Appreciate your efforts to add test coverage and document this, embracing the style of the current docs.

Now serious question, is that your adorable cat in your Github profile?

toupeira:
@olimorris thanks for the merge! :heart: 

> Now serious question, is that your adorable cat in your Github profile?

lol unfortunately it isn't, just a random cat I found on the internets :crying_cat_face:

PR #1542, opened 2025-05-31
feat(image): add Telescope provider for image picker

petobens:
Add support for using Telescope as a provider in the image slash command, allowing users to pick images via Telescope in addition to snacks.nvim and vim.ui.select. Update documentation and config to reflect the new provider option.

### DIFF
--- a/lua/codecompanion/config.lua
+++ b/lua/codecompanion/config.lua
@@ -165,7 +165,7 @@ local defaults = {
           opts = {
             dirs = {}, -- Directories to search for images
             filetypes = { "png", "jpg", "jpeg", "gif", "webp" }, -- Filetypes to search for
-            provider = providers.images, -- snacks|telescope|default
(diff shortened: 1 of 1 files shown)


olimorris:
Thanks for this, awesome to add Telescope support.

You'll need to update `providers/init.lua` and the `image_providers` function though.

petobens:
Done

lua/codecompanion/config.lua
```diff
@@ -165,7 +165,7 @@ local defaults = {
           opts = {
             dirs = {}, -- Directories to search for images
             filetypes = { "png", "jpg", "jpeg", "gif", "webp" }, -- Filetypes to search for
-            provider = providers.images, -- snacks|default
+            provider = providers.images, -- snacks|telescope|default
```
petobens, line 168: Is the order here relevant?

PR #1549, opened 2025-06-01
Add `<group>name</group>` references for tool groups

ravitemer:
## Description

This change is purely cosmetic and affects only the internal reference tracking for better organization. The functionality remains exactly the same. 

Tool groups are now tracked as `<group>name</group>` references instead of being immediately expanded into individual tools. This provides better organization and tracking of grouped tools in the chat buffer.

| Before | After |
|--------|-------|
| image | image |
| Tool groups were immediately expanded into individual `<tool>name</tool>` references | Tool groups now create unified `<group>name</group>` references |

## Checklist


olimorris:
I'm actually planning a PR which I think will need all these tools expanded.

Appreciate there's a lot of context but I can at least manually remove a tool I don't need as part of a group.

ravitemer:
Oh noo! There are some groups which have too many tools and just the references will fill up the buffer. Is there any other way to address this without disabling the `show_references` option?

image

olimorris:
Inevitably that's going to be the case. I think most users experience this if they use workspaces or add lots of buffers or files.

I'm particularly picky about any UI changes in CodeCompanion. My preference is to always keep the UI as simple as possible, hence why I like to keep tool output to a minimum. But there is a difference when a user has manually chosen to add references and context to the chat buffer, in the case of tools, groups, buffers etc. To not display them fully, is to not inform the user of what's going on under the hood.

I'm working on a PR which is going to add most of Copilot Chat's tools into CodeCompanion, under the `@agent` group. This is likely going to go well over 10+ tools and I think it's imperative to let users know at all times what the LLM could be using to generate their responses. And they could choose to delete one of the references an remove the tool from the conversation. Because the references are always above the cursor position and the user can quickly navigate to the LLM's response with `[[`, I'm fine with them being long.

Right now, I'm striving to keep the codebase as simple as possible because I have a lot planned in the coming months. Happy to discuss this at a later point in time.

ravitemer:
Makes sense. Thanks!

PR #1555, opened 2025-06-01
refactor(tools): split out files tool

olimorris:
## Description

This PR intends to split out the `@files` tool into three parts:

1. `@create_file`
2. `@read_file`
3. `@insert_edit_into_file`

Following GitHub's approach in Copilot Chat of giving tools a single responsibility. These will then be grouped into `@files`.

In playing around with Copilot Chat in VS Code for a number of weeks its become apparent that the LLMs seem to perform better when tools have a simple schema. I also note that no system prompt is included for any of the tools and instead is covered in some detailed instructions in an overall system prompt.

## Checklist

### DIFF
--- a/lua/codecompanion/config.lua
+++ b/lua/codecompanion/config.lua
@@ -74,9 +74,9 @@ local defaults = {
           callback = "strategies.chat.agents.tools.editor",
           description = "Update a buffer with the LLM's response",
         },
-        ["files"] = {
-          callback = "strategies.chat.agents.tools.files",
-          description = "Update the file system with the LLM's response",
+        ["create_file"] = {
+          callback = "strategies.chat.agents.tools.create_file",
+          description = "Create a file in the current working directory",
           opts = {
             requires_approval = true,
           },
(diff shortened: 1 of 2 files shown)


pratyushmittal:
Hey @olimorris, are you sure about this?

Generally, lesser tools are better than more tools. Though, more tools are better than a single tool with a complicated schema.

I think the `@files` schema is pretty simple to handle all CRUD operations in a single tool. Did you experience any cases where the tool didn't perform as it should?

jinzhongjia:
Perhaps it can be understood that the file tool has too many functions and llm cannot determine how to use it?、
I've encountered a situation where the tool result is provided but it doesn't use it

olimorris:
> Generally, lesser tools are better than more tools.

I'm seeing practical usage to the contrary. Sharing the GitHub Copilot Chat Agent prompt:

[Agent.json](https://github.com/user-attachments/files/20547935/Agent.json)

They have nearly 30 tools which make up their "Agent" mode (which as far as I can tell is just a bunch of tools wrapped in a very neat system prompt). The standout point is that the only tool they need system instructions for is their tool which edits files. Presumably because it's the most complicated. As is our `UPDATE` action in `@files`.

In my mind, it makes sense to decouple `CREATE` and `READ` from `UPDATE`. I also plan on removing the `DELETE` action as I think it bares too much risk for something which is likely not used by many people.

PR #1569, opened 2025-06-05
Reuse existing tabs when jumping to a file.

Davidyz:
## Description

This PR implement a `goto_file_action` that reuses opened tabs when jumping to a file under cursor. 
This avoids opening too many tabs in a neovim session.

## Checklist

- [x] I've read the [contributing](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md) guidelines and have adhered to them in this PR
- [ ] I've updated `CodeCompanion.has` in the [init.lua](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/init.lua#L239) file for my new feature
- [ ] I've added [test](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md#testing) coverage for this fix/feature
- [x] I've updated the README and/or relevant docs pages
- [ ] I've run `make all` to ensure docs are generated, tests pass and my formatting is applied

PR #1570, opened 2025-06-05
New tool: next edit suggestion

Davidyz:
## Description

This is a tool that allows the LLM to jump to a file (and optionally a particular line in that file).

This tool offers a more streamlined workflow for people who talk to AI for suggestions but still choose to write the code by themselves, instead of 100% vibe coding. This can also be handy when exploring a repository, where you can just ask the LLM to take you from place to place, as an alternative to the `gR` keymap.

- [x] if #1569 is merged, I'll use the same `tabnew_reuse` function as the default jump action for this PR.

## Checklist


olimorris:
This is an awesome idea and a great addition to the default tools.

I'm in the process of simplifying the tools in the plugin, starting in #1555. Examining GitHub Copilot's prompts I've noticed that they're incredibly simple - most don't even have a system prompt because they're explained adequately in the schema.

If we can achieve that with this tool, that would be great. I also think we should be explicit and call it `next_edit_suggestion` instead of `nes`.

Davidyz:
> I'm in the process of simplifying the tools in the plugin, starting in #1555.

I hesitated about whether I should include the ability to edit (or propose a diff) in this tool, but it'll be convenient if we can reuse existing tools to achieve this and just keep this one as simple as a jump.

> I also think we should be explicit and call it `next_edit_suggestion` instead of `nes`.

ok I can do that.

lua/codecompanion/strategies/chat/agents/tools/next_edit_suggestion.lua
```diff
@@ -0,0 +1,103 @@
...
+  name = "next_edit_suggestion",
+  schema = {
+    type = "function",
+    ["function"] = {
+      name = "next_edit_suggestion",
+      description = "Suggest a possible position in a file for the next edit.",
+      parameters = {
+        type = "object",
+        properties = {
+          path = {
```
olimorris: Can we change this to `filepath` and say the `The relative path to the file to edit, including its filename and extension.` as it's consistent with the work I'm doing in splitting out the `@files tool.

lua/codecompanion/strategies/chat/agents/tools/next_edit_suggestion.lua
```diff
@@ -0,0 +1,103 @@
...
+            type = "string",
+            description = "The path to the file",
+          },
+          line = {
```
olimorris, line 28: And, can we make this 0-based? Will be consistent with Neovim and Tree-sitter as well as GitHub Copilot. I suspect, LLMs are probably more "comfortable" with it as well given - although that's a guess on my part - no backing evidence. Also will be consistent with the work I'm doing in #1555.

olimorris:
Btw, I'm planning on adding a new a `background` strategy at some point which will execute a request in the background and silently output the data via an event. I imagine this could be very interesting for this tool.

lua/codecompanion/strategies/chat/agents/tools/next_edit_suggestion.lua
Davidyz: ok I'll rename it to `filepath`. I have reservations about "relative" though. In VectorCode, when it queries from other projects, the backend returns absolute paths. This avoids excessive `../../../`. I think it's better to just say "path" without requiring it to be relative.

lua/codecompanion/strategies/chat/agents/tools/next_edit_suggestion.lua
Davidyz: > Will be consistent with Neovim and Tree-sitter

I think neovim itself is NOT consistent :rofl: `nvim_win_set_cursor` is 1-based and `nvim_buf_set_lines` is 0-based. But ok I'll change this to align with other tools.

PR #1572, opened 2025-06-05
feat(keymaps): support `:map-arguments` in keymaps

seblyng:
## Description

Allow `opts` to keymap config option

I was a bit unsure where to potentially add the docs for this. Do you have any pointers on where you want it? I also wasn't able to run `make all` because I had some problems installing `pandoc`, but let's see if the pipeline succeeds.

I think the dedicated `description` option could be deprecated in favor or `opts.desc`, but it would be harsh to just remove it right away at least

## Related Issue(s)

Fixes https://github.com/olimorris/codecompanion.nvim/discussions/1537

## Checklist

### DIFF
--- a/doc/codecompanion.txt
+++ b/doc/codecompanion.txt
@@ -1,4 +1,4 @@
-*codecompanion.txt*         For NVIM v0.10.0         Last change: 2025 June 01
+*codecompanion.txt*         For NVIM v0.10.0         Last change: 2025 June 06
 
 ==============================================================================
 Table of Contents                            *codecompanion-table-of-contents*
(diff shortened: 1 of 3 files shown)


olimorris:
Great PR and thanks for contributing. I think we could probably put a single sentence [here](https://codecompanion.olimorris.dev/configuration/chat-buffer.html#keymaps).

seblyng:
Hopefully I did it correct. I got a bunch of other doc changes about jump action, and some other punctuation and unicode-stuff. I just ignored those changes for now at least since those were irrelevant to my changes, and I got the same changes on the `main`-branch

olimorris:
Killed it. Thanks for such a great PR.

seblyng:
Thanks for the plugin! Just recently starting to dive into AI plugins in neovim, and this is by far the most intuitive plugin I have found thus far😄

Will definitely contribute more in the future if I find time and some bugs/features that I think will make things better

PR #1574, opened 2025-06-05
fire RequestFinished event after fetch slash cmd request completion

petobens:
This is useful to clear a spinner indicator after a `/fetch` command fires a request.


olimorris:
This should automatically be handled by the http client. My spinner implementation duly clears this.

olimorris:
2025-06-06 11_02_11 - WezTerm

PR #1582, opened 2025-06-07
feat(fetch): only show "Cache" option if cached files exist

petobens:
Previously, the "Cache" option was always shown in the selection menu, even when no cached files were available. Now, "Cache" is only included if there are cached files, improving the user experience and preventing invalid selections.

Personally I don't generally cache urls so this is a small QoL for me. Feel free to dismiss it :)


olimorris:
Awesome thanks. Always appreciate your QOL PRs.

olimorris:
Bit of a nitpick, please stick to the PR template. It makes things slightly quicker for me when it comes to reviewing a PR and sets a good standard for others.

petobens:
Will do

PR #1586, opened 2025-06-07
refactor(tools): roll `@editor` functionality into `@insert_edit_into_file`

olimorris:
## Description

#1349 brought about some awesome changes which enabled LLMs to update files using diff patching. This PR looks to roll up the functionality and features of the `@editor` tool into the `@insert_edit_into_file` tool. With it, this will bring reliability, resiliency and a much simpler implementation.

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/agents/tools/insert_edit_into_file.lua
+++ b/lua/codecompanion/strategies/chat/agents/tools/insert_edit_into_file.lua
@@ -1,7 +1,12 @@
 local Path = require("plenary.path")
+local buffers = require("codecompanion.utils.buffers")
+local config = require("codecompanion.config")
+local keymaps = require("codecompanion.utils.keymaps")
 local log = require("codecompanion.utils.log")
 local patch = require("codecompanion.strategies.chat.agents.tools.helpers.patch")
+local ui = require("codecompanion.utils.ui")
 
+local api = vim.api
 local fmt = string.format
 
 local PROMPT = [[<editFileInstructions>
@@ -19,7 +24,7 @@ The system uses fuzzy matching and confidence scoring so focus on providing enou
 ---Edit code in a file
 ---@param action {filepath: string, code: string, explanation: string} The arguments from the LLM's tool call
 ---@return string
-local function edit(action)
+local function edit_file(action)
   local filepath = vim.fs.joinpath(vim.fn.getcwd(), action.filepath)
   local p = Path:new(filepath)
   p.filename = p:expand()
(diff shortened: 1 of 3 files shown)


olimorris:
I'm going to be continually tweaking this for a day or two but please feel free to test it out.

bassamsdata:
Hello @olimorris,

I tried out this PR a bit today, and it’s working well! I do have a couple of small suggestions:

1. When the LLM asks for permission to modify the buffer, it applies the changes and then shows `(tool processing...)`. To be honest, I didn't know it was waiting for me to accept or reject the changes. After I saw this message I noticed the changes stayed but I think the LLM thought that the changed has not bee applied. But then, I did `gta` and it worked.
Maybe a notification or some guidance at that point would help clarify what to do next.
Screenshot 2025-06-09 at 12 12 44 PM


2. One other minor thing: the buffer didn’t seem to be watched by default, I had to activate the watch manually. Not sure if that’s the intended behavior.

Please feel free to ignore these suggestions if they don’t align with the intended behavior or default design of this PR.
Thank you for your work.

olimorris:
@bassamsdata thanks so much for testing this out. Really appreciated.

1. Interesting and great point. It does send out a `vim.notify` however I forget that some users don't have these enabled. I'll add an extmark backup
2. The `#buffer` variable is watched automatically but the buffer slash commands are not. I'll add a fix for that. Presume this was how you were loading content into the chat buffer?

olimorris:
2025-06-09 22_00_40 - WezTerm@2x

Users now get a notification in the chat buffer letting them know the action resides with them as well as a notification in case they've navigated away from the chat buffer whilst it runs.

jacobdaddario:
Oh that's super interesting with the diff approval and rejection. I use mini-diff and sometimes have issues when applying diffs to buffers that aren't currently in view. I wonder if this will collide with that? I like the general idea behind this though. It seems to emphasize turn-taking more than the current paradigm.

As a note from a user, I generally like seeing the diff over LLM plans when using the `@editor` tool. Reviewing the actual diff can be easier than reviewing the LLM's plan. This would be a welcome change in that direction, as I generally have to append the following to all of my requests to get that sort of workflow:
> Please ignore any instruction to explain reasoning step-by-step or to use pseudocode. Instead, immediately perform any user-requested tool actions without delay or additional explanation, focusing only on the outcome.

olimorris:
@jacobdaddario I plan on adapting the plugin to mimic GitHub Copilot Chat's `modes` more. So by default the plugin will be in `ask` mode where it's geared more towards answering questions and reviewing code. If you were to bring in the `@agent` group into the chat buffer a whole new system prompt would overwrite the default one. So where you're appending that text everytime, you could have that in the `@agent` system prompt (although I expect I'll naturally take care of that when I curate it).

> I wonder if this will collide with that?

I have no idea.

> It seems to emphasize turn-taking more than the current paradigm.

It does with the default settings but nothing that cannot be changed in the config.

jacobdaddario:
> If you were to bring in the @agent group into the chat buffer a whole new system prompt would overwrite the default one. So where you're appending that text everytime, you could have that in the @agent system prompt (although I expect I'll naturally take care of that when I curate it).

Oh wow, that sounds so cool. I was thinking to myself that being able to place it into different attitudes would be super useful.

I'll look into the mini-diff thing if I am able to. Thanks as always for your hard work.

bassamsdata:
Hello @olimorris,

> but the buffer slash commands are not. I'll add a fix for that. Presume this was how you were loading content into the chat buffer?

Yes, this is how I tried it with the slash command. Thanks for that.

> Users now get a notification in the chat buffer letting them know the action resides with them as well as a notification in case they've navigated away from the chat buffer whilst it runs.

Thank you, this works beautifully now. And amazingly, it performs noticeably better than the `@editor` command and doesn’t miss locations at all (sonnet 4 and gpt 4.1). Really appreciate it.

One other suggestion though: would it be possible to show the range of lines changed in the chat buffer near the file that was modified? Similar to how Zed does it. If you're open to the idea but don’t have time, I’d be happy to open a complementary PR for it once this one is merged.

**Reason**: When I work in `@agent` style, I tend to operate in autopilot mode, letting the LLM apply changes freely. But when the tool is triggered multiple times, I sometimes lose track of where to start reviewing.

One small question as well: I'm not sure why I'm receiving all these notifications.

Screenshot 2025-06-09 at 5 45 47 PM

olimorris:
> One other suggestion though: would it be possible to show the range of lines changed in the chat buffer near the file that was modified? Similar to how Zed does it. If you're open to the idea but don’t have time, I’d be happy to open a complementary PR for it once this one is merged.

Obtaining this range should be super easy. This current PR should move that buffer to the start of those lines so you can visually see the changes.

My preference for something like this would be to emit an event that you can hook into which informs you of the buffer that's been edited and the line range. Then if you wanted to color that in your status column you could.

From a UI perspective, I'm cool with making changes to a chat buffer as I consider that CodeCompanion's domain but I'm reluctant to do anything outside of that as that's changing the aesthetic of the user's editor.

Was this where your thinking was at?

olimorris:
The buffer slash command is now watches buffers by default.

bassamsdata:
Thank you for the update.

> Obtaining this range should be super easy. This current PR should move that buffer to the start of those lines so you can visually see the changes.

Great.

> My preference for something like this would be to emit an event that you can hook into

This is great as well. What I meant is not to change anything in the editor. It’s just that when a file is changed, it would be helpful to show the range as simple text next to it, something like this in the chat buffer:  
edited one file: `core/statusline.lua` line range (100–112)

Basically, it’s just to be able to know where the LLM changed lines, especially when it’s editing multiple files.

One small error that seems to happen often: when I first ask the LLM to edit the file, it fails because it adds the path to the absolute path, so it doesn’t see the file (not sure if this is solvable, it's just an observation).

Then I just tell it to check again with the correct path I’ve sent. (This happens when I share files from the slash command.)
Screenshot 2025-06-10 at 1 12 55 PM

olimorris:
It will need to be a file in your current working directory. I've restricted the scope of the tool to that but you do need a better error message than that.

bassamsdata:
Ah, indeed, the file wasn't in the current workspace. Thanks.

olimorris:
> This is great as well. What I meant is not to change anything in the editor. It’s just that when a file is changed, it would be helpful to show the range as simple text next to it, something like this in the chat buffer:
edited one file: core/statusline.lua line range (100–112)

I hear you. I'll put this on the roadmap for a future PR sometime soon. There's a few other event related PRs I want to implement too.

Jufralice:
Hi @olimorris ,

thanks a lot for this feature.
But I already miss the @editor tool (using line numbers)
This new patch way doesn't work well for the gemini  'gemini-2.5-flash-preview-05-20' model.
It 'sometimes' work, but most of the time, the LLM just reply with its 'patch' response inside a 'normal' response message instead of actually calling the tool.
And then, when I ask it again to actually call the tool, it does it but often miss a 'Begin Patch' or 'End Patch' marker or other incorrect patch errors.

The old @editor way did work great nearly all the time though...
I guess that other gemini-2.5-flash users will have the same issue, so It would be great if we could still use the good old @editor tool.

Or, even better, if someone find a way to make the @insert_edit_into_file tool works better, by adapting it's system prompt maybe, that would be awesome.

But again, thank you for your great work!

olimorris:
@Jufralice - Feel free to open up a discussion with your findings. The system prompt can absolutely be tweaked

Jufralice:
Yes. Sorry for the duplicated message but I just realized that its place was maybe better to be in the discussion: https://github.com/olimorris/codecompanion.nvim/discussions/1608
I'll try to tweak and test the system prompt to see if I can come up with something....

PR #1601, opened 2025-06-10
Update slash-commands.md docs

pratyushmittal:
I didn't know we can select multiple files together until I read this: https://github.com/olimorris/codecompanion.nvim/discussions/1589#discussioncomment-13401712

This is so useful.

I never needed to select multiple files in Telescope before. Hence it didn't strike to me.

## Description

Update the documentation.

### DIFF
--- a/doc/usage/chat-buffer/slash-commands.md
+++ b/doc/usage/chat-buffer/slash-commands.md
@@ -23,14 +23,16 @@ The _fetch_ slash command allows you to add the contents of a URL to the chat bu
 
 ## /file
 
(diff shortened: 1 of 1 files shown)


olimorris:
My preference would be to remove the link to the comment and instead put the video below the header. The URL of the video is:

[https://github.com/user-attachments/assets/f1b8c90d-cbfd-4cda-a825-7bbd18c08a00]

olimorris:
You raise a good point. I think I need to do some screencasts which show people how I use CodeCompanion as it's grown so much and people may be missing super useful tips like this.

pratyushmittal:
Hey @olimorris, I have updated the PR.

It will look like this: https://github.com/pratyushmittal/codecompanion.nvim/blob/patch-1/doc/usage/chat-buffer/slash-commands.md

PR #1610, opened 2025-06-12
Add `default_tools` option to chat strategy

Davidyz:
## Description

This PR adds the option to automatically load tools when opening a new chat.

## Checklist

### DIFF
--- a/doc/configuration/chat-buffer.md
+++ b/doc/configuration/chat-buffer.md
@@ -224,12 +224,12 @@ require("codecompanion").setup({
     }
   }
 })
-
 ```
 
 ### Automatically Add Tools to Chat
 
-You can configure CodeCompanion to load tools to new chats by default:
+You can configure the plugin to automatically add tools to new chat buffers:
+
 ```lua
 require("codecompanion").setup({
   strategies = {
@@ -244,8 +244,7 @@ require("codecompanion").setup({
 })
 ```
 
-This works for both built-in tools and external tools configured by the users or
-[extensions](/configuration#extensions).
+This also works for [extensions](/configuration/extensions).
 
 ## Prompt Decorator
 
--- a/doc/configuration/chat-buffer.md
+++ b/doc/configuration/chat-buffer.md
@@ -228,7 +228,7 @@ require("codecompanion").setup({
 
 ### Automatically Add Tools to Chat
 
-You can configure the plugin to automatically add tools to new chat buffers:
+You can configure the plugin to automatically add tools and tool groups to new chat buffers:
 
 ```lua
 require("codecompanion").setup({
@@ -236,7 +236,10 @@ require("codecompanion").setup({
     chat = {
       tools = {
         opts = {
-          default_tools = { "my_tool" }
+          default_tools = { 
+            "my_tool",
+            "my_tool_group"
+          }
         },
       }
     }
--- a/doc/configuration/chat-buffer.md
+++ b/doc/configuration/chat-buffer.md
@@ -227,6 +227,26 @@ require("codecompanion").setup({
 
 ```
 
+### Automatically Add Tools to Chat
+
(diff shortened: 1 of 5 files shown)


doc/configuration/chat-buffer.md
```diff
@@ -227,6 +227,26 @@ require("codecompanion").setup({
...
+          default_tools = { "my_tool" }
+        },
+      }
+    }
+  }
+})
+```
+
+This works for both built-in tools and external tools configured by the users or
+[extensions](/configuration#extensions).
```
Davidyz: I'm not _quite_ sure whether this is the correct way to reference another section in the documentation.

olimorris:
This is a killer idea and implementation. Thanks so much.

Could we add support for groups as well? It would be much easier to put `files` in the `default_tools` than the individual ones.

ravitemer:
@Davidyz Thanks for the awesome addition! Aren't you having any trouble with the history extension when using default_tools? 


image

For me there are two reference blocks rendering on opening the chat due to history extension calling references.render. 
```lua
    local function _create_chat(adapter, settings)
        local chat = require("codecompanion.strategies.chat").new({
            save_id = save_id,
            messages = messages,
            context = context,
            settings = settings,
            adapter = adapter,
            title = title,
          })
        chat.refs = chat_data.refs or {}
        chat.references:render()
        chat.tools.schemas = chat_data.schemas or {}
        chat.tools.in_use = chat_data.in_use or {}
        chat.cycle = chat_data.cycle or 1
        chat.opts.title_refresh_count = chat_data.title_refresh_count or 0
        log:trace("Successfully created chat with save_id: %s", save_id or "N/A")
        return chat
    end
```

Also the history extension currently replaces the entire chat.refs. 

One way to address is to pass the refs in the `Chat.new()` and extend it with the `default_tools` and finally render the references in `self.ui:render`

@olimorris Is this the way forward or is there some other way? I can open a PR to address this.

olimorris:
Good point. This should be for new chats and disabled for chats we're restoring

Davidyz:
> Aren't you having any trouble with the history extension when using default_tools?

Is this the main branch or the summary/memory branch? I've been using the summary branch and I rarely load the whole chat, but only the summaries, so I might have missed this.

ravitemer:
> Good point. This should be for new chats and disabled for chats we're restoring

Can we have it while restoring as well? From what I see it doesn't need any changes from codecompanion side. Instead of calling `references:render()` doing `add()` solves all the problems! `add()` takes care of adding context block, appending refs, handling duplicate refs and also updating `chat.refs`. My bad I used `render()` in the first place 😅

```lua
         -- chat.refs = chat_data.refs or {}
        -- chat.references:render()
        for _, ref in ipairs(chat_data.refs or {}) do
            chat.references:add(ref)
        end
```

olimorris:
> Can we have it while restoring as well? 

I would oppose this on the basis that when I save a chat to disk, I expect to restore it to the same state I saved it. CodeCompanion adding tools on my behalf would be a frustration if I didn't intend for them to be there. If I need them, then I can add them manually and be confident that CodeCompanion-History will save and restore them when required.

ravitemer:
Yeah, agreed. Was tempted by not needing to do any changes at codecompanion. 

I guess then should we pass refs with Chat.new() and if they are > 0, avoid adding default tools?

arnm:
@ravitemer Hey, nice addition. Can this be configured for prompt library prompts?

Davidyz:
> @ravitemer Hey, nice addition. Can this be configured for prompt library prompts? 

You're suggesting automatically loading some prompt from the prompt library? I can give it a try, but could you explain how it's different from just configuring a different system prompt?

olimorris:
@arnm you could just add the tools to the prompt yourself via `@my_tool @my_tool2 my prompt to the LLM...`.

Seems overkill to add this feature in.

PR #1614, opened 2025-06-12
Add optional `collapse_tools` setting for tool groups

ravitemer:
## Description

Hey! Remember our conversation in #1549 about showing tool group references? I completely understood your perspective on keeping the UI transparent and showing users exactly what tools are available.

However, I've been thinking about this and realized there's a middle ground that respects both viewpoints. Instead of forcing the behavior change on everyone, what if we made it **optional**?

This PR adds a `collapse_tools` option to a tool group config that defaults to `false` (current behavior). Users who want the expanded view keep it as-is, while those who prefer cleaner references can opt-in:

```lua
groups = {
  ["full_stack_dev"] = {
    opts = {
      collapse_tools = true, -- Shows as single <group>full_stack_dev</group>
    },
  },
}
```

An important benefit here is that when `collapse_tools` is enabled and a user removes the group reference from the chat, **all tools in that group and their associated system messages are properly cleaned up together**. This actually solves a cleanup issue where individual tool removal from the chat didn't properly remove group's system message.

This way:
- Default behavior stays exactly the same (backward compatible)
- No forced UI changes that might affect your upcoming @agent group
- Users with many tool groups can choose to reduce clutter
- Full transparency is maintained - it's still user's choice what to show
- Better cleanup behavior when groups are removed

I know you mentioned keeping the codebase simple, but this feels like a reasonable compromise that gives users control without changing defaults. What do you think?

## Screenshots

Edit: 

With `collapse_tools` set to true for `@files` it looks like:

image

## Checklist


olimorris:
Now this is an interesting idea.

I was thinking about this yesterday as I was mulling over using a `conceal` instead. I use this when I have huge tailwind css references in html:

https://github.com/user-attachments/assets/2a67d96c-95e0-4978-afc2-60ebdce5b06e

It's actually just a Tree-sitter query that enables this:

```scm
((attribute
        (attribute_name) @att_name (#eq? @att_name "class")
        (quoted_attribute_value (attribute_value) @class_value) (#set! @class_value conceal "…")))
```

Would this be a cleaner way to go?

ravitemer:
That's a really interesting approach with Tree-sitter conceal! I can see how that would be elegant for the visual aspect, and I think it could actually work **alongside** the `collapse_tools` option to give users ultimate flexibility.

Let me share my perspective with the `@mcp` tool group. As you know it is a group of two tools that work together, and when added to chat, they come with a system message containing all active MCP server schemas, resources, etc. 

In the coming days, `mcphub` extension will have a `make_tools` option  which will make each MCP server its own tool group along with individual tools. The challenge is that when users delete individual tools from the chat, the group's system message stays behind (since other tools might still need it). But with MCP servers, it's more natural to add/remove entire servers as units.

This is where `collapse_tools` becomes really valuable - it matches the user's mental model. When I delete `<group>brave_search</group>`, I expect the entire Brave Search MCP server (all its tools + system message) to be removed as one logical unit.

The conceal feature would be fantastic for visual cleanup, and it could work beautifully with `collapse_tools`:
- Users who want individual tool visibility: default behavior 
- Users who want visual cleanup: conceal feature
- Users who want logical grouping: `collapse_tools = true`
- Users who want both: conceal + `collapse_tools` together

This gives maximum flexibility - get full transparency by default, get visual control through conceal, optional logical grouping with `collapse_tools`.

I personally would prefer the fourth `- Users who want both: conceal + collapse_tools together` option

olimorris:
Sorry I was brain dumping on how we could visually apply some additional UI stuff to the chat buffer and taking away from the core points of your PR.

> When I delete <group>brave_search</group>, I expect the entire Brave Search MCP server (all its tools + system message) to be removed as one logical unit.

100% agree. I've been reflecting on this in recent days as I move to a more unified way of grouping tools into "modes". I think this _now_ should be the default behaviour as you proposed in #1549. To extend on your sentence above, I think if a user tags a group into the chat buffer then they're aware of what that means for an LLM's behaviour.

If we resurrect #1549, where does that leave this PR in your mind? Sorry it's late and I'm wrambling somewhat.

ravitemer:
> If we resurrect https://github.com/olimorris/codecompanion.nvim/pull/1549, where does that leave this PR in your mind? Sorry it's late and I'm wrambling somewhat.

I feel this PR builds on #1549 and is better than that. It didn't consider cleaning up the tools or system prompt and is not customizable

> I think this now should be the default behaviour as you proposed in https://github.com/olimorris/codecompanion.nvim/pull/1549.

In order to achieve grouping by default setting `collapse_tools = true` for the `@full_stack_dev` and `@files` or in general should do it. Users can disable for individual groups if they want.

Please let me know if you want `collapse_tools` to be true by default.

olimorris:
Yeah let's set it to true then this looks good to me aside from the conflict to resolve 👍🏼

ravitemer:
@olimorris This is ready for review. All groups will now be collapsed by default and cleaned properly.

ravitemer:
https://github.com/olimorris/codecompanion.nvim/pull/1614/commits/8f4c9365ea4d782d9bb7417de46666609a7433e8 is a minor fix and is not related to this PR. But related to #1610 . Sorry for adding it here.

lua/codecompanion/strategies/chat/tools.lua
```diff
@@ -92,6 +98,36 @@ function Tools:add(tool, tool_config)
   return self
 end
 
+function Tools:add_group(group, tools_config)
```
olimorris, line 105: Only feedback, can we add the doc block with params and return types to this?

Great spot to move this here btw.

PR #1622, opened 2025-06-12
fix(inline): revert keymap changes

olimorris:
## Description

During testing I changed the inline diff keymaps. This reverts that.

## Checklist

### DIFF
--- a/lua/codecompanion/config.lua
+++ b/lua/codecompanion/config.lua
@@ -411,7 +411,7 @@ local defaults = {
       keymaps = {
         accept_change = {
           modes = {
-            n = "gda",
+            n = "ga",
           },
           index = 1,
           callback = "keymaps.accept_change",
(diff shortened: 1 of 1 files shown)


igbanam:
Hi Oli 👋🏽 

Fantastic work with this. I don't use it cos I don't use Neovim, but I am watching cos it has some ideas I would like to see adopted in the Vim community eventually. Thanks for taking the bold step to bring LLMs into Vim.

I wanted to call out that these mappings may interfere with existing mappings in some configs. I don't know how widely used coc.nvim is in the Neovim community, but `gr` is their "Go To References" mapping.

One suggestion I have is to offer configurability for these maps. — _p.s: if this exists, and I am lacking context, please let me know! I'm, in this PR, just an enthusiast._ 🙏🏽

olimorris:
They're throw away mappings. They're used once, then deleted and tied to specific buffers 👍.

And yep, all in the config.lua file for people to change as they wish.

PR #1628, opened 2025-06-13
Improve @insert_edit_into_file tool prompt (mainly move system_prompt into tool's description) to fix issues with gemini-2.5-flash not correctly calling the tool

Jufralice:
## Description

Some tweaks in the system_prompt and tool description of the @insert_edit_into_file tool.
The most important part is the move of the Patch Format specifications into the tool description.
Doing so greatly help some LLM models like gemini-2.5-flash to call that tool with success. This model indeed used to not call the tool at all and instead return it's patch code inside a normal LLM response.

I only tested with 'gemini-2.5-flash-preview-05-20', so it would be great if other users tries to test these change with other models before we merge this.

You can read more info about the experiments I did and the tweaks I apply to this prompt in the below related discussion.

## Related Discussion:

  - Fixes #1619  

## Checklist


archie-judd:
Thanks for this, I am not using gemini, but copilot, and thought I would leave my experience here, because I was also not getting successful patch formatting in 16.2.2.

Some notes after asking it to add a docstring to a function in `#buffer`:
- The LLM prints a response plan and the `tool_code()` call, including the patch diff, and asks in the chat before before actually attempting to make an edit. I'm not sure if printing the entire diff is deliberate, but assume it's because of this line:
> For each file, first give a short description of what needs to be changed
- When I tell it to proceed it says it cannot find the buffer and I need to provide an absolute path.
- When I tell it use the absolute path, it infers one and it does actually successfully make an edit.
- When I ask it to make multiple edits in the same file, it makes a single diff, rather than asking the user to accept / reject hunks sequentially. Again, not sure if this is deliberate, but assume that's because of this line:

> You can use the tool multiple times in a response, but *only* if you need to modify *different* files.


Appreciate my experience will vary because I'm not using gemini.

lua/codecompanion/strategies/chat/agents/tools/insert_edit_into_file.lua
```diff
@@ -10,17 +10,39 @@ local wait = require("codecompanion.strategies.chat.agents.tools.helpers.wait")
...
-Use the insert_edit_into_file tool to modify files.
-NEVER show the code edits to the user - only call the tool. The system will apply and display the changes.
-For each file, give a short description of what needs to be changed, then use the insert_edit_into_file tools. You can use the tool multiple times in a response, and you can keep writing text after using a tool.
-The insert_edit_into_file tool is very smart and can understand how to apply your edits to the user's files, you just need to follow the patch format instructions carefully and to the letter.
+You have access to the `insert_edit_into_file` tool to modify files.
+For each file, first give a short description of what needs to be changed, then make the `insert_edit_into_file` tool call. You can keep writing text after using a tool.
+## Example Tool Call
+When you want to edit a file, your output should use the `insert_edit_into_file` tool like this:
+```
+tool_code('insert_edit_into_file', {
```
olimorris, line 20: We shouldn't need to tell the LLM how to call the tool. They've gone through a long process of training in order to do that. Also, unsure of where you've got `tool_code` from?!

olimorris:
Sharing the below conversation I've had just now...

Me:

````md
> Context:
> -  <buf>lua/codecompanion/strategies/chat/agents/tools/helpers/patch.lua</buf>
> - <buf>lua/codecompanion/strategies/chat/agents/tools/insert_edit_into_file.lua</buf>
> - <url>https://cookbook.openai.com/examples/gpt4-1_prompting_guide</url>
> - <file>tests/strategies/chat/agents/tools/test_insert_edit_into_file.lua</file>

Can you do a critical analysis which looks at the prompting guide from OpenAI and compares it to how I've implemented file edits with `patch.lua` and `insert_edit_into_file.lua`. Explicitly, I'd like you to review the system prompt. Please share your feedback
````

claude-sonnet-4:

````md
Looking at your implementation compared to OpenAI's GPT-4.1 prompting guide, I'll provide a critical analysis of your patch format and system prompt.

## Format Comparison

**Your format vs OpenAI's V4A format:**

Your format is actually quite similar to OpenAI's recommended V4A diff format, with some key differences:

### Similarities ✅
- Both avoid line numbers (excellent choice)
- Both use `@@` identifiers for context
- Both provide before/after context lines
- Both use `+/-` prefixes for changes
- Both support multiple `@@` statements for nested context

### Key Differences

**Patch Boundaries:**
- **Yours:** `*** Begin Patch` / `*** End Patch`
- **OpenAI's:** More explicit action headers like `*** Update File: path/file.py`

**Context Handling:**
- **Your approach:** Focus lines + pre/post context in a single block
- **OpenAI's approach:** Separate context blocks per change with explicit file actions

## System Prompt Analysis

### Strengths of Your Current Prompt

1. **Clear format specification** - You embed the full `FORMAT_PROMPT` which is comprehensive
2. **Fuzzy matching emphasis** - Good mention of confidence scoring
3. **Practical guidance** - "focus on providing enough context to uniquely identify the location"

### Critical Issues with Your Current Prompt

#### 1. **Missing Agentic Instructions**
According to the OpenAI guide, you should include these three key reminders:

```lua
-- Missing from your prompt:
- Persistence: "Keep going until the user's query is completely resolved"
- Tool-calling: "If you are not sure about file content, use your tools to read files"
- Planning: "You MUST plan extensively before each function call"
```

#### 2. **Insufficient Behavioral Guidance**
Your current prompt lacks specific instructions about:
- When to use the tool vs when to ask for clarification
- How to handle multiple files
- Error recovery strategies

#### 3. **No Examples in System Prompt**
The OpenAI guide recommends including examples in an `# Examples` section rather than just in the description field.

## Recommended System Prompt Improvements

Here's how I'd restructure your system prompt following OpenAI's best practices:

```lua
local IMPROVED_PROMPT = [[
# Role and Objective
You are a code editing agent that modifies files using precise patch formats. Your goal is to help users make accurate code changes while following instructions to the letter.

# Core Instructions
- Always use the insert_edit_into_file tool to modify files - NEVER show code edits to users directly
- You are an agent - keep going until the user's query is completely resolved before ending your turn
- If you are not sure about file content or codebase structure, use your tools to read files first: do NOT guess or make up an answer
- You MUST plan extensively before each function call and reflect on outcomes

# Workflow Steps
1. **Understand the Request**: Analyze what changes are needed
2. **Gather Context**: Read relevant files if needed to understand current state
3. **Plan Changes**: Break down edits into logical, safe modifications
4. **Apply Patches**: Use insert_edit_into_file with proper patch format
5. **Verify**: Confirm changes were applied correctly

# Patch Format Requirements
]] .. patch.FORMAT_PROMPT .. [[

# Critical Guidelines
- Provide 3 lines of context before and after changes (minimum for unique identification)
- Use `@@` identifiers when 3 lines aren't sufficient for uniqueness
- Multiple `@@` statements can be chained for deeply nested contexts
- Focus on providing enough context to uniquely identify locations - the system uses fuzzy matching
- For multiple changes in one file, use a single tool call with multiple patch blocks

# Examples
## Simple Change
*** Begin Patch
   function calculateTotal() {
     let sum = 0;
-    return sum;
+    return sum + tax;
   }
*** End Patch

## With Context Identifiers
*** Begin Patch
@@class UserService
@@  updateUser(id, data) {
     if (!id) return null;
-    return this.save(data);
+    return this.save(id, data);
     }
*** End Patch

# Error Prevention
- If you cannot uniquely identify a location, ask the user for clarification
- If a patch fails, analyze the error and try a different approach
- Always provide clear explanations of what you're changing and why
]]
```

## Implementation Recommendations

### 1. **Add Planning Enforcement**
```lua
-- In your tool execution, consider adding a planning step:
if not args.explanation or #args.explanation < 10 then
  return output_handler({
    status = "error", 
    data = "Please provide a detailed explanation of the changes being made"
  })
end
```

### 2. **Improve Error Messages**
Your current error handling could be more instructive:

```lua
-- Instead of just:
error(fmt("Bad/Incorrect diff:\n\n%s\n\nNo changes were applied", patch.get_change_string(change)))

-- Consider:
local suggestion = "Try adding more context lines or using @@ identifiers to uniquely locate the code"
error(fmt("Could not locate patch target:\n\n%s\n\n%s", patch.get_change_string(change), suggestion))
```

### 3. **Consider Tool Chaining**
The OpenAI guide emphasizes tool usage. You might want to encourage reading files first:

```lua
-- Add to your prompt:
"Before making changes, consider using read_file to understand the current state, especially for complex modifications."
```

## Overall Assessment

Your implementation is **fundamentally sound** and follows many best practices from the OpenAI guide. The main gaps are in the **system prompt's behavioral instructions** rather than the technical patch format itself.

**Priority fixes:**
1. Add the three agentic instruction types (persistence, tool-calling, planning)
2. Include concrete examples in the system prompt
3. Add more specific workflow guidance
4. Improve error messaging to guide users toward solutions

Your patch parsing logic is robust and handles edge cases well. The fuzzy matching approach is exactly what OpenAI recommends for dealing with formatting variations.

Would you like me to help implement any of these specific improvements?

````

**Comment by olimorris** [comment]
@archie-judd appreciate your feedback but this complicates this PR somewhat so removing your comment to be tight on scope.

@pratyushmittal - Keen to get your views on this PR before we do any merging. It's clear some Gemini users have struggled and I know it's not a strong performer when it comes to function calling so I'm mindful of making concessions that might impact other LLMs.

File: lua/codecompanion/strategies/chat/agents/tools/insert_edit_into_file.lua (hunk 1 above)
Comment (Jufralice) [reply lua/codecompanion/strategies/chat/agents/tools/insert_edit_into_file.lua:20]: yeah... also not sure about that tool_code(). It was a suggestion of the LLM and I noticed that adding that like he proposed did give better results, I then kept it. But maybe we should indeed remove that 'tool_code' function call.

About the fact that we should not tell the LLM how to call tool, I agree. But for some reason adding that example seems to help a lot to reduce the amount of LLM responses where he would put his Patch code into a normal response message instead of in the 'code' parameter of a tool call.

But I agree that we should probably not merge these changes since it's probably not necessary (and possibly more confusing and costly) for the users that uses a more advanced model than gemini-2.5-flash.

We should probably keep the default prompt and description like it was and let the users change the tool's prompt and/or description in their personal config. 
 
The question now is maybe: should we completely get rid of the system_prompt and put everything in the tool's description? and then we should probably add the possibity to change that description in our personal codecompanion config like we can do for the system_promp... ?

**Comment by olimorris** [comment]
Sharing some of my findings with Gemini...

With the following prompt:

````md
Can you add a comment block to #buffer at the top describing what this file does using the @insert_edit_into_file tool?
````

where the buffer is the `minimal.lua` file in the root of the project.

With `gemini-2.5-flash-preview-05-20`:

```
[WARN] 2025-06-16 09:31:33
Tool insert_edit_into_file: ...companion/strategies/chat/agents/tools/helpers/patch.lua:125: Invalid patch format: missing Begin/End markers
```

and with `gemini-2.0-flash`:

```
[WARN] 2025-06-16 09:31:58
Tool insert_edit_into_file: ...n/strategies/chat/agents/tools/insert_edit_into_file.lua:95: Bad/Incorrect diff:

1:  ---@diagnostic disable: missing-fields
+-- This file is a minimal example configuration for using codecompanion.nvim
+-- It sets up lazy.nvim, installs codecompanion along with its dependencies,
+-- and configures it to use the copilot adapter.
+
2:  
3:  --[[

No changes were applied
```

No issues with calling the tool. Just issues with understanding the tool.

olimorris:
I've been giving some thought to how we solve the this issue and unfortunately I don't think this PR is the answer. We run the risk of playing whac-a-mole (a phrase I'm unfortunately using far too often recently), whereby we get it working for Gemini and make it less performant for [insert name of currently working LLM].

We need to move towards an [Aider style](https://github.com/olimorris/codecompanion.nvim/discussions/1619#discussioncomment-13539715) approach whereby users can specify a separate algorithm to the default. This will be useful for this tool and no doubt future ones.

Really appreciate your efforts on this. For now, I'd recommend rolling your own custom tooling for Gemini.

Issue #1631, opened 2025-06-13
[Bug]: @insert_edit_into_files_tool can't provide a relative path because it doesn't know the cwd.


archie-judd:
### Did you check the docs and existing issues?

- [x] I have read all the docs
- [x] I have updated the plugin to the latest version before submitting this issue
- [x] I have searched for existing issues/discussions
- [x] I have searched the existing issues of plugins related to this issue

### Neovim version (nvim -v)

0.11.2

### Operating system/version

Macos

### Adapter and model

Copilot and sonnet-4

### Describe the bug

Hi,

I've been using the `@insert_edit_into_file` tool, which is great, but have noticed its instructed to use the relative path when inserting edits into the provided file or buffer.

I don't think that the LLM knows the relative path when the tool is invoked, and it seems to take an educated guess based on the absolute path of the file its editing. In my experience, it often just passes the filename as the relative path, which can cause issues in `get_bufnr_from_filepath`.

As an example: say there is a file at `/home/user/workspaces/repo/src/file.py`, and the cwd is `/home/user/workspaces/repo/`, if the LLM guesses a relative path of just `file.py`, [this line](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/utils/buffers.lua#L83) will return `/home/user/workspaces/repo/file.py`, and so no buffer will be found.

This has happened to me a few times and I often end up with these responses using copilot and sonnet-4:

![Image](https://github.com/user-attachments/assets/a1a08f06-4892-4680-b661-bb35519c2e19)

I wonder if a more robust approach could be to make the LLM use an absolute path instead? I have tested by changing the "relative" on [this line](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/strategies/chat/agents/tools/insert_edit_into_file.lua#L195) to "absolute", and it works perfectly for me. 

Can you think of any issues with doing this?

I'd be happy to make a PR if this change sounds sensible!

### Steps To Reproduce

Try editing a file from with a cwd that is a child directory of the workspace root. 
Inspect the `args` on this line:
https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/strategies/chat/agents/tools/insert_edit_into_file.lua and see if the LLM correctly provides the relative path from the cwd.

### Expected Behavior

The args.filename attribute should be the relative path of the file to be edited, but in my experience it often isn't.

### init.lua file

```lua
[...]

PR #1633, opened 2025-06-13
fix: paths in variables and slash commands are now relative

olimorris:
## Description

Currently, LLMs receive absolute paths to buffers via variables and slash commands. This gives them knowledge of a user's disk. We also ask, in the `insert_edit_into_file` tool, for the LLM to give the relative path to a file to edit. This led to the issues listed below.

This PR seeks to unify CodeCompanion and only send paths relative to the current working directory.

This PR will also ensure that it's much clearer to the LLM to understand what a buffer or file's filepath is, preventing the issues below.

## Related Issue(s)

#1630
#1631

## Checklist


lua/codecompanion/strategies/chat/variables/buffer.lua
```diff
@@ -89,7 +98,7 @@ end
 ---@return string
 function Variable.replace(prefix, message, bufnr)
   local bufname = buf_utils.name_from_bufnr(bufnr)
```
toupeira, line 78: This currently returns just the basename, which is probably the main reason why the LLMs get misled since it's so prominent in the prompt. Changing this to a relative path seems to help a lot:

```diff
diff --git i/lua/codecompanion/utils/buffers.lua w/lua/codecompanion/utils/buffers.lua
index 0ad1523..8ffc7a9 100644
--- i/lua/codecompanion/utils/buffers.lua
+++ w/lua/codecompanion/utils/buffers.lua
@@ -34,7 +34,7 @@ end
 ---@param bufnr number
 ---@return table
 function M.name_from_bufnr(bufnr)
-  return vim.fn.fnamemodify(api.nvim_buf_get_name(bufnr), ":t")
+  return vim.fn.fnamemodify(api.nvim_buf_get_name(bufnr), ":.")
 end
 
 ---Get the information of a given buffer
```

olimorris:
When you share a buffer or a file, the LLM will now see:

````md
<attachment filepath="lua/codecompanion/strategies/chat/agents/tools/helpers/patch.lua" buffer_number="9">
User's current visible code in a file, this should be the main focus:
```lua
1:  local M = {}
2:  
3:  M.FORMAT_PROMPT = "some code"
```
</attachment>
````

The message will change depending on whether you've shared a file or a buffer and if it's come from a slash command or a variable.

The system prompt also contains a reference to attachments as well. This now falls inline with GitHub Copilot Chat's way of sharing context.

File: lua/codecompanion/utils/buffers.lua [hunk 2]
```diff
@@ -76,6 +76,54 @@ function M.get_open(ft)
...
+
+  local message = opts.message or "Buffer content"
+
+  local formatted_content = string.format(
+    [[<buffer filepath="%s" number="%s">%s:
+%s</buffer>]],
+    relative_path,
+    bufnr,
+    message,
+    relative_path,
```
toupeira: Small mistake here, looks like `string.format` won't complain if you pass too many arguments :grinning: 

```suggestion
```

toupeira: With this change applied this seems to work great, thanks! :heart:

PR #1642, opened 2025-06-15
refactor(create_file): use `vim.uv` functions

jinzhongjia:
## Description

- replace plenary.path with native `vim.uv` functions for file and directory operations

Plenary creates files on Windows often fail to create parent successfully.

Using vim's built-in method is more convenient

## Related Issue(s)

- https://github.com/nvim-lua/plenary.nvim/issues/393

## Checklist

### DIFF
--- a/lua/codecompanion/adapters/copilot/helpers.lua
+++ b/lua/codecompanion/adapters/copilot/helpers.lua
@@ -30,9 +30,9 @@ end
 ---Get a list of available Copilot models
 ---@param adapter CodeCompanion.Adapter
 ---@param get_and_authorize_token_fn function Function to get and authorize token
----@param github_token table The github token object
+---@param oauth_token string The oauth token
 ---@return table
-function M.get_models(adapter, get_and_authorize_token_fn, github_token)
+function M.get_models(adapter, get_and_authorize_token_fn, oauth_token)
   if _cached_models and _cache_expires and _cache_expires > os.time() then
     return _cached_models
   end
@@ -44,13 +44,16 @@ function M.get_models(adapter, get_and_authorize_token_fn, github_token)
     _cached_adapter = adapter
   end
 
-  get_and_authorize_token_fn(adapter)
-  local url = github_token.endpoints.api or "https://api.githubcopilot.com"
+  if not get_and_authorize_token_fn(adapter) then
+    return {}
+  end
+
+  local url = "https://api.githubcopilot.com/models"
   local headers = vim.deepcopy(_cached_adapter.headers)
-  headers["Authorization"] = "Bearer " .. github_token.token
+  headers["Authorization"] = "Bearer " .. oauth_token
 
   local ok, response = pcall(function()
-    return curl.get(url .. "/models", {
+    return curl.get(url, {
       sync = true,
       headers = headers,
       insecure = config.adapters.opts.allow_insecure,
@@ -58,13 +61,13 @@ function M.get_models(adapter, get_and_authorize_token_fn, github_token)
     })
   end)
   if not ok then
-    log:error("Could not get the Copilot models from " .. url .. "/models.\nError: %s", response)
+    log:error("Could not get the Copilot models from " .. url .. ".\nError: %s", response)
     return {}
   end
 
   local ok, json = pcall(vim.json.decode, response.body)
   if not ok then
-    log:error("Error parsing the response from " .. url .. "/models.\nError: %s", response.body)
+    log:error("Error parsing the response from " .. url .. ".\nError: %s", response.body)
     return {}
   end
 
--- a/lua/codecompanion/adapters/copilot/init.lua
+++ b/lua/codecompanion/adapters/copilot/init.lua
@@ -300,7 +300,7 @@ return {
             return { ["gpt-4.1"] = { opts = {} } } -- fallback
           end
         end
-        return copilot_helper.get_models(self, get_and_authorize_token, _github_token)
+        return copilot_helper.get_models(self, get_and_authorize_token, _oauth_token)
       end,
     },
     ---@type CodeCompanion.Schema
--- a/lua/codecompanion/strategies/chat/ui/init.lua
+++ b/lua/codecompanion/strategies/chat/ui/init.lua
@@ -148,7 +148,7 @@ function UI:open(opts)
   end
 
   log:trace("Chat opened with ID %d", self.chat_id)
-  util.fire("ChatOpened", { bufnr = self.chat_bufnr })
+  util.fire("ChatOpened", { bufnr = self.chat_bufnr, id = self.chat_id })
 
   self.tools = require("codecompanion.strategies.chat.ui.tools").new({
     chat_bufnr = self.chat_bufnr,
@@ -176,7 +176,7 @@ function UI:hide()
     vim.cmd("buffer " .. vim.fn.bufnr("#"))
   end
 
-  util.fire("ChatHidden", { bufnr = self.chat_bufnr })
+  util.fire("ChatHidden", { bufnr = self.chat_bufnr, id = self.chat_id })
 end
 
 ---Follow the cursor in the chat buffer
--- a/tests/adapters/copilot/test_helpers.lua
+++ b/tests/adapters/copilot/test_helpers.lua
@@ -120,4 +120,77 @@ T["Copilot Helper Stats"]["show_copilot_stats handles overage_permitted for prem
   h.eq(true, found_overage, "Expected overage_permitted to be displayed in stats")
 end
 
+T["Copilot Helper Get Models"] = new_set()
+
+T["Copilot Helper Get Models"]["retrieves models with correct structure"] = function()
+  local mock_models = {
+    {
+      id = "model1",
+      capabilities = { type = "chat", supports = { streaming = true, tool_calls = true, vision = true } },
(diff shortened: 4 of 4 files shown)


jinzhongjia:
The related disscussion: https://github.com/olimorris/codecompanion.nvim/discussions/1641

jinzhongjia:
Actually, I tried to fix the plenary problem, but honestly, the plenary's code is a bit hard to maintain

lua/codecompanion/strategies/chat/agents/tools/create_file.lua
```diff
@@ -8,26 +7,33 @@ local fmt = string.format
...
-        data = fmt("**Create File Tool**: `%s` already exists as a directory", action.filepath),
-      }
-    else
-      return {
-        status = "error",
-        data = fmt("**Create File Tool**: File `%s` already exists", action.filepath),
-      }
-    end
+  -- Check if path already exists
+  if vim.fn.isdirectory(filepath) == 1 then
```
olimorris: Where possible, I'd like to not use `vim.fn` commands as they are old Vimscript functions. I think we can use `vim.uv.fs_stat` instead.

lua/codecompanion/strategies/chat/agents/tools/create_file.lua
```diff
@@ -8,26 +7,33 @@ local fmt = string.format
...
+    return {
+      status = "error",
+      data = fmt("**Create File Tool**: `%s` already exists as a directory", action.filepath),
+    }
+  end
+
+  if vim.fn.filereadable(filepath) == 1 then
```
olimorris: Let's use `vim.uv.fs_stat` instead.

olimorris:
> Plenary creates files on Windows often fail to create parent successfully.

This is the first I've heard of this and we've had this functionality in the plugin for over a year now. Can you cross-reference the Plenary issue that exists for this.

Plenary is deeply embedded in the plugin so I'm surprised this is the one area you've been experiencing problems with.

jinzhongjia:
Reproduction is easy, just ask codecompanion to use @files to create a non-existent file (the parent folder does not exist either)

like: `luas/test.txt`

lua/codecompanion/strategies/chat/agents/tools/create_file.lua
jinzhongjia: Why do you have to use uv? Although uv will bring a small amount of performance improvement, it is not convenient to use.

Even the old vim fn functions, I think they execute well, there is nothing wrong with that

olimorris:
> Reproduction is easy, just ask codecompanion to use

I don't have a Windows machine so reproduction isn't easy at all.

This needs to be a bug report filed with Plenary so someone can fix this upstream.

The changes in this PR are minimal but I'm unwilling to change core functionality because a user says they're having a problem without providing any evidence or willing to fix it for the wider community. As I mentioned before, this is the first time I've had a windows issue raised with this functionality.

jinzhongjia:
The main reason is that there is a problem with the parsing parent function of plenary. I see that it uses regular matching
For example, this path `C:/Users/jin/code/LspUI.nvim/luas/test.txt` will give `C:\` after passing parent

jinzhongjia:
Unfortunately, plenary doesn't seem to be developing actively at the moment, and it uses more ancient methods, such as using vim.loop

In fact, throughout 2024, Plenary has few submissions, and many PRs have been unaudited and not merged.

And this is plenary releated issue: https://github.com/nvim-lua/plenary.nvim/issues/393

Despite claims to be repaired, similar issues remain

jinzhongjia:
I'll change this pr to use uv later

jinzhongjia:
Now use vim.uv and it can ensure that the function is the same as it was originally

lua/codecompanion/strategies/chat/agents/tools/create_file.lua
```diff
@@ -26,8 +26,64 @@ local function create(action)
   end
 
   local ok, result = pcall(function()
-    p:touch({ parents = true })
-    p:write(action.content, "w")
+    local parent_dir = vim.fs.dirname(filepath)
+
+    -- Improved recursive directory creation
+    local function create_dir_recursive(path)
```
olimorris: Let's move this into a `utils/files.lua` and add some test coverage

lua/codecompanion/strategies/chat/agents/tools/create_file.lua
```diff
@@ -26,8 +26,64 @@ local function create(action)
...
+        error(fmt("Incomplete write: expected %d bytes, wrote %d bytes", content_length, bytes_written))
+      end
+
+      return bytes_written
+    end)
+
+    -- Always close the file regardless of write success
+    local close_success, close_err, close_errname = vim.uv.fs_close(fd)
+    if not close_success then
+      error(fmt("Failed to close file: %s (%s)", close_err, close_errname))
```
olimorris: This means nothing to an end user. The tool creates files and we're notify the user that we couldn't "close a file". Let's insert a generic error message but log:error what's actually happened

lua/codecompanion/strategies/chat/agents/tools/create_file.lua
```diff
@@ -26,8 +26,64 @@ local function create(action)
...
+      end
+
+      local parent = vim.fs.dirname(normalized)
+      if parent ~= normalized and not vim.uv.fs_stat(parent) then
+        create_dir_recursive(parent)
+      end
+
+      local success, err, errname = vim.uv.fs_mkdir(normalized, 493)
+      if not success and errname ~= "EEXIST" then
+        error(fmt("Failed to create directory %s: %s (%s)", normalized, err, errname))
```
olimorris: You're not following the return types here

olimorris (changes requested):
I accept this is a Plenary issue. However, for this to be added into CodeCompanion I want this to exist as a utility file - something that we can reuse across the plugin. It's no use existing in one tool if this issue is going to arise again in the future.

The impact of this is we will need test coverage for that utility file as well.

So unfortunately there's a bit of work to get this PR in a state that I'd be happy to accept.

jinzhongjia:
OK, then according to my understanding, what I need to do next is
- move `create_dir_recursive` to `utils/files.lua` and add unit test
- The return value of the `create` does not follow the convention

jinzhongjia:
Maybe now the requirement is met

lua/codecompanion/strategies/chat/agents/tools/create_file.lua
```diff
@@ -26,11 +27,61 @@ local function create(action)
...
   local ok, result = pcall(function()
-    p:touch({ parents = true })
-    p:write(action.content, "w")
+    local parent_dir = vim.fs.dirname(filepath)
+
+    -- Ensure parent directory exists
+    if not vim.uv.fs_stat(parent_dir) then
+      local success, err_msg = files.create_dir_recursive(parent_dir)
+      if not success then
+        error(err_msg) -- This error is specific to directory creation
```
olimorris: This method has the following return type:

```lua
@return {status: "success"|"error", data: string}
```

We need to return a status of `error` in order to populate `stderr` on the agent executor class. This is the emitted to the tool. Throwing `error` counteracts this.

Can we replace all instances of `error` with this approach? See what I've done in `file_search.lua` and `grep_search.lua`

tests/utils/test_files.lua
olimorris, line 1: Unfortunately this breaks convention for _all other_ tests in CodeCompanion. Please reference our mini.test tests.

I have a specific workspace for tests - you can literally load that in a chat buffer and ask the LLM to convert your existing case to a mini.test version.

lua/codecompanion/strategies/chat/agents/tools/create_file.lua
jinzhongjia: The external use of pacall for wrapping, and the error information can be obtained effectively, and then return it in a unified manner below.

jinzhongjia:
I've corrected the code according to the specification

I carefully checked the specifications of other codes and the implementation is really great

Thank you for your work

jinzhongjia:
I also tried to run unit tests on Windows, but found that there were basically no errors. The code specifications were written really well. The problem was that there were some problems with the backslash processing, but it did not affect the use, but it would only cause some unit tests to be unable to pass on Windows

olimorris:
Sorry for going quiet on this. Will 100% get to this but will be next week sometime

olimorris:
Just taken a look at this. Great PR and happy to merge. I've just merged #1665, can you update this for a similar style of output? Tbh, I think this should just be one line such as "Created `lua/test.lua`" or something.

jinzhongjia:
OK, I'll make the changes, and it looks like I need to resolve the conflict

jinzhongjia:
I noticed that only `success` uses the new folding effect, other cases still use the old output style

jinzhongjia:
Now the result of creating the file will collapse normally, and the collapsed content is the file content

jinzhongjia:
I made some additional modifications. llm output will only tell llm that the llm file was created successfully and will not bring content to llm, which can reduce token consumption a little

But the user can see the collapse (the content of the file)

olimorris:
> I noticed that only success uses the new folding effect, other cases still use the old output style

It works for all handlers but it detects words in the `failure_words` array in the config for anything outside of the `success` handler.

PR #1650, opened 2025-06-16
fix(patch): allow inserting content at end of file with the insert_edit_info_file tool

Jufralice:
## Description

I noticed a bug when trying to use the insert_edit_info_file tool to append some content at the end of a file. 
I spotted that the issue was because the 'get_match_score' function in patch.lua needs to iterate with an extra 'i' in order to find the best location for the 'pre' lines.

example:
you have a file with ['line1', 'line2']
you try to patch with:
*** Begin Patch
line1
line2
+new_line3
*** End Patch

the 'get_best_location' function wont find a location and will return '0'
we need to iterate a last time with i=3 in the get_match_score function so that 'i - #change.pre' is 3-2=1 and then we will be able to get the scores of or 2 pre lines

I added a test that fail before the fix (and obviously not after)

note that you may not be able to reproduce that issue if you test with a file instead of a buffer. This is because when you save a file, an extra \n is often added at the end, then the 'edit_file' function in insert_edit_info_file.lua will get an extra empty lines when it splits the content by '\n'. We can then by chance iterate our needed extra time in the get_best_location' (except if the LLM kept that empty line in his 'pre' lines).

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/agents/tools/helpers/patch.lua
+++ b/lua/codecompanion/strategies/chat/agents/tools/helpers/patch.lua
@@ -202,7 +202,6 @@ local function get_best_location(lines, change)
   -- trim_spaces mode solves all of these
   local best_location = 1
   local best_score = 0
-  -- we need to add an extra iternation (one more than the line count) so that we cover the case where we try to add content at the end of the file.
   for i = 1, #lines + 1 do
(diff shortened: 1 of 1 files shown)


olimorris:
This is a great PR. Many thanks indeed.

PR #1654, opened 2025-06-16
fix(patch): Support lines starting by '-' (e.g for bullets points in markdown files) for the insert_edit_info_file tool

Jufralice:
This commit updates the regex in `M.parse_changes` to correctly only catch newlines around the patch delimiters. A new test case has been added to ensure correct parsing when lines start with characters like '-', something which is common in markdown files.

## Description

The previous patch parsing logic expected new-lines or whitespace around the "Begin Patch" and "End Patch" delimiters. But I think that we want to only catch new_lines so that we can keep the potential white space at the start of the next lines (the LLM is instructed to start the unchanged lines in a patch by a space)

Don't hesitate to let me know if you think that this should be fixed in another way.

## Checklist


olimorris:
This looks like another great catch too, thanks.

olimorris:
Let's fix up these conflicts and then we're good to go.

Jufralice:
Thanks a lot @olimorris ,
I'll fix the conflicts right now

PR #1655, opened 2025-06-16
Use the same diff format as in patch

pratyushmittal:
## Description

It is highly recommended to not use line-numbers for diffs in LLMs. The diffs generated in watched files used line numbers. This can confuse the LLMs.

This PR fixes that by using a common format.

## Related Issue(s)

It addresses some of the issues mentioned here: https://github.com/olimorris/codecompanion.nvim/discussions/1619#discussioncomment-13473490

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/watchers.lua
+++ b/lua/codecompanion/strategies/chat/watchers.lua
@@ -105,7 +105,7 @@ end
 ---@param old_content table
 ---@param new_content table
 ---@return string
-local function format_changes_as_diff(old_content, new_content)
+function Watchers:format_changes_as_diff(old_content, new_content)
   -- Convert line arrays to strings for vim.diff
   local old_str = table.concat(old_content, "\n") .. "\n"
   local new_str = table.concat(new_content, "\n") .. "\n"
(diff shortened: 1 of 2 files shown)


olimorris:
Love this! Thanks @pratyushmittal

bassamsdata:
Hello @pratyushmittal,

After this update, manually changing the buffer seems to cause the LLM to apply diffs incorrectly, often resulting in wrong changes. I mainly work with Claude Sonnet 4.

I understand it’s better to keep diffs consistent, but I remember when I updated watchers, I faced a similar issue with Claude. That’s why I chose to keep line numbers, Sonnet 4 even suggested that it works better that way.

Have you encountered any problems when changing multiple places inside the buffer manually? And have you run into any issues using diffs with line numbers as before?

olimorris:
I too am having some issues with sonnet-4 and it keeping track of what changes have been made.

Unfortunately with CodeCompanion and its LLM agnostic approach, we're playing whack-a-mole.

Do we think line numbers at the top of the diff made a real different to any other tools? Or, can this be managed through prompt engineering?

My connection will be patchy but I propose rolling back this change for now

pratyushmittal:
Ah! It makes sense to revert this. I didn't think about the watchers in buffers before doing this.

I don't think I ever saw the LLMs generate a diff which contained line numbers. So we can go ahead.

A better solution might be to update our patch parser which can ignore line numbers in `@@ linnumbers @@` if any.

PR #1657, opened 2025-06-16
refactor(tools): improve system prompt in `insert_edit_into_file` tool

olimorris:
## Description

Further refinements to make the tool more robust.

## Related Issue(s)

#1619

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/agents/tools/helpers/patch.lua
+++ b/lua/codecompanion/strategies/chat/agents/tools/helpers/patch.lua
@@ -57,7 +57,6 @@ Example with multiple blocks of changes and `@@` identifiers:
 *** End Patch
 
 This format is similar to the `git diff` format; the difference is that `@@[identifiers]` uses the unique line identifiers from the preceding code instead of line numbers. We don't use line numbers anywhere since the before and after context, and `@@` identifiers are enough to locate the edits.
-Be mindful that the user may have shared attachments that contain line numbers, but these should not be used in your patch.
 IMPORTANT: Be mindful that the user may have shared attachments that contain line numbers, but these should NEVER be used in your patch. Always use the contextual format described above.]]
 
 ---@class Change
--- a/lua/codecompanion/strategies/chat/agents/tools/helpers/patch.lua
+++ b/lua/codecompanion/strategies/chat/agents/tools/helpers/patch.lua
@@ -11,12 +11,28 @@ The `[PATCH]` is the series of diffs to be applied for each change in the file. 
 +[new code]
  [3 lines of post-context]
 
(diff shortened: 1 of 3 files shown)


pratyushmittal:
Hey @olimorris, any particular reason for `(including line numbers)` in buffers? I just wanted know your thoughts around it.

Does Copilot or others user line-numbers too or is it recommended anywhere?

olimorris:
So I personally don't use the `insert_edit into_file` tool. My workflow typically consists of sharing a buffer with an LLM and conversing around a feature or a refactor I want to perform. I then review and critique the LLM's output, oftentimes using part of it, but not all of it.

If I take the chat buffer file, for example, that's nearly 1500 lines long. It's really advantageous to have the LLM be able to reference specific line numbers so I know what it's referring to and where to perform any edits myself. (My logic for this is I just won't remember what the LLM's done if I haven't reviewed or been personally invested in the edits).

I was playing around in Cursor yesterday and it's quite frustrating at how incorrect its line number referencing is.

Jufralice:
Hi @olimorris , @pratyushmittal 

Thanks a lot for these improvements.

I totally agree about reviewing every LLM suggestions, discuss it, adapt it and only take part of it...

But I'm a bit confused... I though that it was the goal of the diff patch logic and the nice insert_edit_into_file tool:
I mean, If we have the LLM proposing changes in multiple places in multiple files, I personnaly find it quite hard to review all these changes directly in the chat buffer because it can become hard to see the context around each proposed change.

We have a tool that allow the LLM to directly integrate his proposed changes in the correct files at the correct places, and the important part is that it doesn't apply the changes directly thanks to the diff-patch logic. We can review all the changes directly in the files, accept/reject/adapt/,... any part of them and continue the discussion with the LLM (thanks to the watcher which keep the LLM up to date with our final changes).

But my confusion comes from the fact that now, since we replaced the @editor tool by the @insert_edit_into_file, we have a 'countdown' to accept or reject the changes and if we dont do that in time, a timeout is returned saying to the LLM that the user did not accept the change in time.

So, indeed, if we don't have all the time we want to accept/reject/adapt the changes before applying them, then I understand that the workaround is to plan ahead in the chat buffer. 

I then have the feeling that the 'countdown' to accept the changes kind of break the powerfullness of the insert_edit_into_file tool.

This is just my feelings but I'm wondering what you think about this 'countdown' and why it is needed, if we can increase it or even completely disable it and what will happen if we does disable it,...

Thanks a lot.

olimorris:
> But my confusion comes from the fact that now, since we replaced the @editor tool by the @insert_edit_into_file, we have a 'countdown' to accept or reject the changes and if we dont do that in time, a timeout is returned saying to the LLM that the user did not accept the change in time.

All configurable by the way - you can turn this off or increase the time. Adapt this to suit your workflow. This is a precursor to `agent` mode where CodeCompanion will be running in an autonomous mode using a whole suite of tools to work on a codebase. It's nice to be able to allow the user to review edits/changes - accept or reject them - before handing it back to the agent to carry on

PR #1663, opened 2025-06-17
fix(tools): @insert_edit_into_file tool: Allow patch starting by empty line before a @@ identifier. 

Jufralice:
This is especially usefull when the LLM forget to add the begin and end markers. It indeed uses to start with for example "\\n@@Focus()" we previously assumed that this was the start of a new focus and we then inserted the ongoing 'patch' in the table. But in this case, the current patch is empty and trying to parse an empty patch returned an error. With this change, we only insert the ongoing patch into the table if it is not empty.

A new test case has been added to validate the correct handling of patches that start with empty lines. A duplicate test name was also resolved for clarity.

## Description

This PR is a little complement to the nice work done by @pratyushmittal  in this #1653 
It allows patches starting by empty line before a @@ identifier. 

## Checklist


lua/codecompanion/strategies/chat/agents/tools/helpers/patch.lua
```diff
@@ -103,8 +103,6 @@ local function parse_changes_from_patch(patch)
       -- empty lines can be part of pre/post context
       -- we treat empty lines as new change block and not as post context
       -- only when the next line uses @@ identifier
-      table.insert(changes, change)
-      change = get_new_change()
     elseif line:sub(1, 1) == "-" then
```
pratyushmittal: This is a good catch.

Just a small nitpick, can we add a noop `do end` to show we need to do nothing in this `else` block?

Something like:

```
    elseif line == "" and lines[i + 1] and lines[i + 1]:match("^@@") then
      -- empty lines can be part of pre/post context
      -- we treat empty lines as new change block and not as post context
      -- only when the next line uses @@ identifier
      -- skip this line and do nothing
      do
      end
    elseif line:sub(1, 1) == "-" then
```

olimorris:
I'm viewing this on my mobile so may have missed some detail but can we also update the tests to be patch tests after @pratyushmittal's test refactor was merged.

lua/codecompanion/strategies/chat/agents/tools/helpers/patch.lua
Jufralice: good idea. thank you @pratyushmittal .
✅

Jufralice:
> I'm viewing this on my mobile so may have missed some detail but can we also update the tests to be patch tests after @pratyushmittal's test refactor was merged.

Hi @olimorris ,
of course. 
done ✅

PR #1664, opened 2025-06-17
feat(chat): add_buffer

daliusd:
## Description

This is feature to expose `add_buffer` as public API. E.g. it can be
used to add current buffer to the most recent chat like this:

```lua
function()
  local bufnr = vim.api.nvim_get_current_buf()
  require('codecompanion').last_chat():add_buffer({ bufnr = bufnr })
end
```

Change itself is quite simple. I simply moved code from
`slash_commands/buffer.lua` to `init.lua`, added path resolution in case
it is not supplied and documented it in comments. As well moved code is
called in `slash_commands/buffers.lua` output. Path resolution was
added to make `add_buffer` usage simpler.

This is part of functionality discussed here:
https://github.com/olimorris/codecompanion.nvim/discussions/1546

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/init.lua
+++ b/lua/codecompanion/strategies/chat/init.lua
@@ -43,6 +43,7 @@
 ---@field tokens? table Total tokens spent in the chat buffer so far
 
 local adapters = require("codecompanion.adapters")
+local buf = require("codecompanion.utils.buffers")
 local client = require("codecompanion.http")
 local completion = require("codecompanion.providers.completion")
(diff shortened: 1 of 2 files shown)


olimorris:
Really not keen on a PR that moves well placed logic from one place to another i.e. slash command -> chat buffer.

I appreciate this is useful for yourself and happy to support your endeavour's but not something I want in `main`.

daliusd:
So long, and thanks for all the fish

PR #1665, opened 2025-06-17
Tool output can be folded in the chat buffer

olimorris:
## Description

Currently, most tools choose to display a minimal amount of output in the chat buffer so as not to clutter the UI.

This PR will automatically fold tool output and overlay extmarks indicating the success of the execution.

## Screenshots

2025-06-17 22_36_21 - WezTerm

## Checklist


Jufralice:
Wow I like that a lot. That will be more convenient than having to go look in the debug window. 
Thanks a lot @olimorris

lua/codecompanion/adapters/openai.lua
```diff
@@ -250,13 +250,12 @@ return {
         return nil
       end
 
-      local output = {}
-      output.role = delta.role or nil
-      output.content = delta.content or ""
```
olimorris, line 255: FYI: This was added last year by someone wanting to support an OpenAI Compatible endpoint. This falsely hides an LLM's output so removing this

olimorris:
## Some notes on this PR

For anyone that reads these pull requests, I thought I'd put some notes down and also some potential notes to my future self.

This was possibly one of the harder PRs that I've ever had to do for CodeCompanion, simply because the `add_buf_message` method in the chat buffer had gotten out of hand. I say out of hand, it was perfectly fine until I decided that I wanted to fold tool output, at which point it became clear that it just couldn't scale whatsoever. And testing it and getting it to work reliably was a nightmare, and debugging felt impossible at times.

So whilst this PR principally adds the ability to fold tool output, it also introduces a builder pattern for how we format output in the chat buffer. It should be much more scalable and if we ever add additional output types, and right now we have tool reasoning and just standard output, then this should flow logically too.

One thing I've not been able to solve/decided it's not worth my time to solve is some of the additional spacing that's created after tool outputs folded. This isn't much of an issue other than there's just more spacing between the tool output and an LLM output than we may like.

PR #1674, opened 2025-06-19
Add tool support to azure_openai adapter

brian-a-esch:
## Description

First off, thank you for this plugin, clearly lots of effort has gone into building it. This PR adds tool support the the azure adapter, as they are now supported in their chat completions api https://learn.microsoft.com/en-us/azure/ai-services/openai/reference#chat-completions. I used the same implementation for tools support as openai_compatible.lua has.

As a side note, I spent quite a bit of time trying to diagnose this issue (thinking the problem could be with azure, or the model I was selecting or not setting up mini diff correctly, etc.). How do you feel about notifying the user if their LLM does not support tool use, but they requested it? It gets silently dropped [here](https://github.com/olimorris/codecompanion.nvim/blob/229ca916ed0c9cfb5d97a06b61bd7ebb69f5d63e/lua/codecompanion/http.lua#L96) if the adapter doesn't support it, so that's one spot a message could be added. Another option is when they add the tool [here](https://github.com/olimorris/codecompanion.nvim/blob/229ca916ed0c9cfb5d97a06b61bd7ebb69f5d63e/lua/codecompanion/strategies/chat/agents/init.lua#L222-L222), we could check if `chat.adapter.opts.tools == true` somewhere & log.

## Related Issue(s)

N/A

## Checklist

### DIFF
--- a/lua/codecompanion/adapters/azure_openai.lua
+++ b/lua/codecompanion/adapters/azure_openai.lua
@@ -10,6 +10,7 @@ return {
   },
   opts = {
     stream = true,
+    tools = true,
     vision = true,
   },
   features = {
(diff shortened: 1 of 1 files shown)


olimorris:
Unfortunately it's just impossible to support this in the plugin.

I'm mulling removing the adapter full stop and would ask users to adapt the current one to suit their own needs.

brian-a-esch:
What is the rationale for it being impossible to support? Lack of testability?

olimorris:
Near infinite endpoints all that supposedly have OpenAI support but in reality can differ quite a bit

PR #1702, opened 2025-06-23
fix(inline): user_prompt mode with dressing and telescope

powerman:
## Description

I suppose there is a conflict between telescope and dressing triggered by switching from Action Palette (telescope) to "Custom Prompt" (dressing). The "Custom Prompt" so far is the only inline prompt with `user_prompt = true`, so it triggers `vim.ui.input()` right after leaving telescope selection menu. Telescope selection menu was in INSERT mode, and I think it tries to leave it on exiting from menu. Dressing in turn tries to enable INSERT mode on showing `vim.ui.input` - and fails (it calls `vim.cmd("startinsert!")` without any effect).

If I disable any one of telescope or dressing this issue disappears.

Adding `vim.schedule` fixes the issue. But I didn't checked the whole codebase - there are may be other similar issues somewhere too.

I've no idea is it possible to write a test for the fix.

## Screenshots

Pressing enter here:
изображение
result in this (note **block** cursor meaning NORMAL mode):
изображение

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/inline/init.lua
+++ b/lua/codecompanion/strategies/inline/init.lua
@@ -294,15 +294,17 @@ function Inline:prompt(user_prompt)
   -- From the prompt library, user's can explicitly ask to be prompted for input
   if self.opts and self.opts.user_prompt then
     local title = string.gsub(self.context.filetype, "^%l", string.upper)
-    vim.ui.input({ prompt = title .. " " .. config.display.action_palette.prompt }, function(input)
-      if not input then
-        return
-      end
+    vim.schedule(function()
+      vim.ui.input({ prompt = title .. " " .. config.display.action_palette.prompt }, function(input)
+        if not input then
+          return
+        end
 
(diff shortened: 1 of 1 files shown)


powerman:
> there are may be other similar issues somewhere too

Same issue here:
https://github.com/olimorris/codecompanion.nvim/blob/dbefc41dba2f46720ecb6ecec34c0c5f80022f2b/lua/codecompanion/strategies/chat/ui.lua#L76-L78

When opening default chat from Action Palette (with telescope) option `start_in_insert_mode = true` does not work. Adding `vim.schedule()` there fixes this.

olimorris:
Hey @powerman. So sorry I've been slow at getting to these. I don't use Telescope myself (and too lazy to install it and test this out 😆). I do recall in one of my other plugins that `vim.schedule` did seem to solve most issues.

This doesn't seem to affect non-telescope users so happy to merge. Thanks for the PR.

PR #1726, opened 2025-06-27
fix(inline): skip think tags output by qwen3

powerman:
## Description

LLM qwen3:8b running using Ollama always return `<think>...</think>` at beginning of the response. Even if prompt includes `/no_think` instruction (in this case there will be no content between the tags, but tags are still present).

This breaks inline strategy parser.

Example output when using `/no_think` in the prompt:

```text
<think>

</think>

{
  "code": "This is my Neovim config.\n",
  "language": "markdown"
}
```

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/inline/init.lua
+++ b/lua/codecompanion/strategies/inline/init.lua
@@ -523,6 +523,9 @@ end
 ---@param output string
 ---@return table|nil
 function Inline:parse_output(output)
+  -- Remove any think tags that may have been added by the LLM (e.g. qwen3).
+  output = output:gsub("<think>.-</think>", "")
+
   -- Try parsing as plain JSON first
   output = output:gsub("^```json", ""):gsub("```$", "")
   local _, json = pcall(vim.json.decode, output)
--- a/tests/strategies/inline/test_inline.lua
+++ b/tests/strategies/inline/test_inline.lua
@@ -28,6 +28,23 @@ T["Inline"]["can parse json output correctly"] = function()
   h.eq("add", json.placement)
 end
 
+T["Inline"]["can parse output with think tags correctly"] = function()
+  local json = inline:parse_output([[
+<think>
+Cogito,
+ergo sum.
+</think>
+
+{
+  "code": "function test() end",
+  "placement": "add"
+}
+]])
+
+  h.eq("function test() end", json.code)
+  h.eq("add", json.placement)
+end
+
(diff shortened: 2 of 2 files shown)


olimorris:
I would rather we managed this through prompt tweaking. What if the user is asking the LLM to generate think tags as output?

Some models will unfortunately not be well suited to the inline assistant or will need a separate adapter which removes thinking via parameters.

powerman:
> I would rather we managed this through prompt tweaking. 

I've tried hard enough to fix this using prompt tweaking - no luck.
The current prompt is https://github.com/powerman/config.nvim/blob/df59bd5d677d3da47f862ce1839b522116918bfe/lua/plugins/codecompanion.lua#L518-L560

> What if the user is asking the LLM to generate think tags as output?

Is this actual for the inline strategy at all?

olimorris:
Then my advice would be to manage this through a separate adapter which disables reasoning/thinking for that model.

Otherwise, the inline strategy will become a list of rules and edits to try and fix every possible quirk in an LLM's output.

If we can't fix it through prompting then I would question whether it's well suited to be used as an inline adapter.

powerman:
> whether it's well suited to be used as an inline adapter

Maybe not. But it's the only model in 8b class which produce reasonable results. I've tried many others when I realized there is no way to work around this qwen3 issue without patching codecompanion, but all others mostly return junk. :disappointed: 

I've nothing against fixing this issue by creating qwen3-specific adapter for inline strategy, but I probably need some example how to create this kind of adapters because all existing are provider-specific, not model+strategy specific.

powerman:
Maybe just add another one user callback option to run before parsing inline strategy response?

olimorris:
Customising parameters on an adapter is covered in the docs, so whatever parameter turns off reasoning you can apply that in an adapters schema.

powerman:
Thanks! This works:

```lua
                ollama_qwen_3_inline = function()
                    return require('codecompanion.adapters').extend('ollama', {
                        handlers = {
                            inline_output = function(...)
                                local openai = require 'codecompanion.adapters.openai'
                                local res = openai.handlers.inline_output(...)
                                if res and res.output then
                                    res.output = res.output:gsub('<think>.-</think>', '', 1)
                                end
                                return res
                            end,
                        },
                        schema = {
                            model = {
                                default = 'qwen3:8b',
                                choices = {},
                            },
                            num_ctx = {
                                default = 40960,
                            },
                        },
                    })
                end,
```

PR #1740, opened 2025-06-28
Migrate to Explicit `@{tool}` and `#{variable}` Syntax for Tools and Variables

ravitemer:
## Description

This PR implements the consensus from [discussion in PR #1503](https://github.com/olimorris/codecompanion.nvim/pull/1503) and follow-up conversations. The previous approach using `@tool` and `#variable` with word-boundary matching led to ambiguity, need for explicit word boundaries. The new approach uses explicit curly-brace syntax for both tools and variables, e.g.:
- Tools: `@{tool_name}`
- Variables: `#{variable_name}` and `#{variable_name}{params}`

### What’s Changed

- **Completion Providers:** All completion providers (cmp, blink, coc) now insert the new explicit syntax (`@{tool}` and `#{variable}`) when a completion item is selected.
- **Highlighting:** Syntax highlighting matches only the new forms.
- **Detection & Replacement:** All logic for finding and replacing tools/variables in chat and inline strategies now exclusively supports the curly-brace syntax.
- **Parameter Support:** Variables with parameters are supported as `#{variable}{params}`.
- **Tests:** All tests updated to use the new syntax. Obsolete tests for word boundaries, spaces, and partial matches have been removed.

### Breaking Change

- All user inputs, prompts must now use `@{tool}` and `#{variable}`.
- The old `@tool` and `#variable` forms are no longer recognized or highlighted.
- Users should update any custom prompts, workflows, or documentation accordingly.

## Related Issue(s)

#1503 

## Screenshots

image

image

## Checklist


ravitemer:
I have tried to update the docs with the new syntax. I didn't update when the tools are in the heading but only inside the markdown blocks. It might cause some confusion to new users having both `@tool` and `@{tool}`. 

I think they need some care before finalizing this. @olimorris How do you want them?

olimorris:
Firstly, thanks so much for this, I really appreciate it. This is so much work and effort

> I didn't update when the tools are in the heading but only inside the markdown blocks. It might cause some confusion to new users having both @tool and @{tool}.

Yeah the docs are tricky. I've opted to remove all references to `@tool` and instead just call them `tool`. Technically, users can still type `@tool` and the completion will wrap it in brackets for them. So for some, it won't be any change in their workflow, but agree with your view - it may cause some confusion for others. I think we have enough examples in the getting started guide though for new users to be able to pick it up.

Sigh...I will have to re-record the videos at some point.

olimorris:
This all looks good to me. Let me know if you're happy for me to merge.

ravitemer:
Thank you! Surprisingly it wasn't as difficult as I thought to be. Because of all those awesome tests and well designed architecture of the plugin. 

> Technically, users can still type @tool and the completion will wrap it in brackets for them

Yes. Just a thought, currently cmp menu is relying solely on the trigger characters. As we get used to the new syntax typing `@{` might be common. Currently it hides the cmp menu after `{`. In the future showing the menu while inside @{} or #{} might be useful. 


For now, Looks good to me!

olimorris:
> Because of all those awesome tests and well designed architecture of the plugin.

Great to hear it! They've bailed me out sooooo many times recently.

Thanks again for all your efforts

PR #1753, opened 2025-06-30
Filter hidden tools, vars and normalize descriptions in help window

ravitemer:
## Description

- **Problem:** Some tool descriptions can be multiline (e.g., for better highlighting with Treesitter in cmp providers like `blink.cmp`). Currently, multiline descriptions cause errors when opening the help/options window.  
  **Solution:** Descriptions are now sanitized by replacing newlines with spaces.

- **Problem:** Extensions like `mcphub` add many dynamic tools and groups, cluttering the help window and making it hard to use.  
  **Solution:** Added a `hide_in_help_window` option for tools, variables, and groups. When set to `true`, these entries are filtered out and not shown in the help/options window. This allows only the static `mcp` group to be visible, hiding all dynamic tools/groups.

I could not find a clear place in the documentation to add this option. Since this feature is mainly for extension authors and not typical users, is it okay to skip documenting the `hide_in_help_window` option?

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/keymaps.lua
+++ b/lua/codecompanion/strategies/chat/keymaps.lua
@@ -49,6 +49,10 @@ M.options = {
       return str .. string.rep(" ", max_length - #str + (offset or 0))
     end
 
+    --- Cleans and truncates a string to a maximum width.
+    ---@param desc string? The description to clean
+    ---@param max_width number? The maximum width to truncate the description to
+    ---@return string The cleaned and truncated description
     local function clean_and_truncate(desc, max_width)
       if not desc then
         return ""


bassamsdata:
Today, I was planning to make a PR to fix an error when showing tools in the help menu. Some extensions don’t have descriptions for their tools, and honestly, it would be nice to have a fallback message like “No description for this tool” instead of throwing an error and preventing the help from displaying.

I’m not sure if this PR will get merged, but if possible, adding feedback would be awesome. If not, no worries, I’ll create a follow-up PR for that.

ravitemer:
@bassamsdata This PR in it's current state handles `nil` descriptions with an empty string like below

image

However, we can have some placeholder also like `(No description)` or something.

image

Personally, I prefer having an empty string but happy to change if you or olimorris want

bassamsdata:
Thank you so much, @ravitemer, this is awesome!

> I prefer having an empty string but happy to change if you or olimorris want

I think an empty string works fine for me, as long as the error is gone, that’s what matters most.

olimorris:
Thanks both for this. Super reasonable PR and fixes a nuisance in the workflow. This is good to me so happy to merge when you're ready.

ravitemer:
Nothing more for this from my side! Please merge it.

PR #1754, opened 2025-06-30
feat(llm_iteration): llm iteration basical support

jinzhongjia:
## Description

Add iterative function for context overflow

Take the same solution as vscode copy chat

## Checklist


olimorris:
There's not a path for this in the plugin right now. I've got a large roadmap of things to add over the next 6 months which will likely change how something like this is implemented. Adding this now would slow down that roadmap.

Appreciate your efforts and as per the contributing guidelines. I'd always advise raising a discussion so your hard work isn't for nothing.

jinzhongjia:
Maybe you should announce your follow-up roadmap so that everyone can participate in it, with more people and more power

jinzhongjia:
This feature iteration is very important, especially in copilot, which can prevent us from consuming another advanced request for no reason.

olimorris:
> This feature iteration is very important, especially in copilot, which can prevent us from consuming another advanced request for no reason.

If you've updated your plugin you shouldn't be experiencing this.

jinzhongjia:
I mean when the context exceeds the limit
Copilot can correctly perform iterative summary, which avoids us needing to manually enable a new chat when the context limit is exceeded (this behavior will obviously consume a new request)

I have used the latest plugin

image

PR #1795, opened 2025-07-07
fix: set foldmethod=manual in window context before fold creation

ThorstenRhau:
This fixes an error that occurred when creating folds with CodeCompanion in buffers where `foldmethod` is not set to "manual" (for example, when using Treesitter or expr-based folding).

- Finds a window displaying the chat buffer
- Temporarily sets `foldmethod=manual` in that window before creating the fold
- Restores the original foldmethod after

Resolves issues like `E350: Cannot create fold with current 'foldmethod'` seen in expr-based folding setups.

Related discussion: [#1788](https://github.com/olimorris/codecompanion.nvim/discussions/1788)

## Description
This fixes an error that occurred when creating folds with CodeCompanion in buffers where `foldmethod` is not set to "manual" (for example, when using Treesitter or expr-based folding).

- Finds a window displaying the chat buffer
- Temporarily sets `foldmethod=manual` in that window before creating the fold
- Restores the original foldmethod after

## Related Issue(s)
Resolves issues like `E350: Cannot create fold with current 'foldmethod'` seen in expr-based folding setups.

Related discussion: [#1788](https://github.com/olimorris/codecompanion.nvim/discussions/1788)

## Checklist


amadeus:
Finally had some time to test out this PR, and as I expected, the original fold never actually gets set, so the code path that exits early is ultimately what's happening when the chat buffer is closed with this PR:

CleanShot 2025-07-07 at 22 36 37

I think the ideal fix is still to try and apply that fold when the chat buffer is not being shown (it's unclear to me whether it's fully unloaded or if just hidden, also unclear to me if a fold can even be performed while a buffer is hidden).

Another alternate fix might be to bring back the chat buffer on accept and then apply the fold

ThorstenRhau:
@amadeus I have updated my PR according to your suggestions.

amadeus:
Oh sick, I’ll have to test this out.

FWIW, might be worth not even bothering with a temp window and just bring back the chat buffer like normal. I think it’s also good from a UX perspective because it got auto hidden, and it’s good to bring it up after we’ve cleaned up the diff so users can see how the AI continues to progress. Newer users may be confused on how to progress or continue the conversation.

Another random thought I had — wondering if some of this logic to set folds and what not should be wrapped in a pcall. If any of this fails than the entire chat buffer session is just broken, would be good to not get ourselves into this state in general.

Ultimately a lot of that rests on olimorris and whether he agrees that’s a good idea tho (maybe there’s something higher up the call stack that can be pcall’d for this

olimorris:
Hi both and sorry for coming to this so late. Thanks for taking a look at this and being so proactive with a PR as this wasn't something I had encountered, unfortunately.

Is there a reason we can't use the window number (`winnr`) from _tools.lua_?

ThorstenRhau:
> Hi both and sorry for coming to this so late. Thanks for taking a look at this and being so proactive with a PR as this wasn't something I had encountered, unfortunately.
> 
> Is there a reason we can't use the window number (`winnr`) from _tools.lua_?

Updated PR incoming. I hope that I got it right this time :-)

lua/codecompanion/strategies/chat/ui/tools.lua
```diff
@@ -97,15 +97,19 @@ function Tools.fold_tools()
...
 ---@param start_line number The starting line of the fold (0-based)
 ---@param end_line number The ending line of the fold (0-based)
 ---@param foldtext string The text to display in the fold
 ---@return nil
-function Tools:create_fold(start_line, end_line, foldtext)
-  if not config.strategies.chat.tools.opts.folds.enabled then
-    return
-  end
-  if start_line > end_line then
+function Tools.create_fold(self, start_line, end_line, foldtext)
```
olimorris, line 105: It's much cleaner if we don't pass in `self` as the first param and instead use `Tools:create_fold`

lua/codecompanion/strategies/chat/ui/tools.lua
```diff
@@ -97,15 +97,19 @@ function Tools.fold_tools()
...
+  local api = vim.api
+  local CONSTANTS = CONSTANTS -- already defined at top of module
```
olimorris, line 107: Not sure what the logic for bringing this into the method was either?

lua/codecompanion/strategies/chat/ui/tools.lua
```diff
@@ -116,13 +120,52 @@ function Tools:create_fold(start_line, end_line, foldtext)
...
+      col = 0,
+      style = "minimal",
+      focusable = false,
+      noautocmd = true,
+    })
+    created_temp = true
+  end
+
+  -- In that window, switch to manual, fold, then restore
+  api.nvim_win_call(win, function()
```
olimorris, line 158: Do we not pass `self.winnr` into this call in order to, in effect, alter the fold properties on the chat buffer?

lua/codecompanion/strategies/chat/ui/tools.lua
```diff
@@ -116,13 +120,52 @@ function Tools:create_fold(start_line, end_line, foldtext)
...
+      if api.nvim_win_get_buf(w) == self.chat_bufnr then
+        win = w
+        break
+      end
+    end
+  end
+
+  local created_temp = false
+  if not win then
+    win = api.nvim_open_win(self.chat_bufnr, false, {
```
olimorris, line 144: Upstream in "ui/init.lua" we have an open method. I think it's preferable to use that rather than re-implementing this in the tools file. But...I would much prefer we don't do any sort of window manipulation at all. Do we really need to open a closed window in order to get this fix to work?

olimorris (commented):
Few comments about decision made below.

My overarching thoughts are, if the genuine solution to this issue is detecting windows, opening windows, applying window specific fold expressions and doing some function deferrals, then perhaps we should not be folding at all. Instead, determine the foldexpr and disable it on the chat buffer.

amadeus:
Random thought -- I think if we auto close the chat on smaller windows, I do think it would be nice if we bring back the chat buffer on accept/reject change, which mostly would solve these problems anyways? Or even if the user closed the window themselves to look at the diff, it just feels natural to then `return` to the chat buffer to see what the AI continues to do

olimorris:
> I do think it would be nice if we bring back the chat buffer on accept/reject change, which mostly would solve these problems anyways

That's a very fair point. I can pull something together later tonight.

olimorris:
I've added a PR in #1816 which will ensure the chat buffer is open after an edit is accepted.

ThorstenRhau:
Thank you for all your input and feedback to this @olimorris 
I hope that we are good to go now :-)

olimorris:
Thanks so much for your hard work and investigation on this. My preference has been to go with #1816 as it solves two problems of the folding issue and forcing a user to manually open the chat buffer again.

ThorstenRhau:
> Thanks so much for your hard work and investigation on this. My preference has been to go with #1816 as it solves two problems of the folding issue and forcing a user to manually open the chat buffer again.

Happy to see the issue getting resolved. That is what matters. 

Keep on the good work with this plugin, I appreciate it.

Jufralice:
Hi @ThorstenRhau ,

I just tested your pull-request and it seems to work fine. 
I'll merge it into my local working branch. Thanks a lot for your great work!

@olimorris , should we maybe consider re-opening this PR since it seems to be a good fix for the folding issue when the chat buffer is not visible?
Your #1816 PR is great but I think that it solves the folding issue only when using the @insert_edit_info_file tool but not for any other tool uses see my comment in https://github.com/olimorris/codecompanion.nvim/pull/1816#issuecomment-3069619356

olimorris:
There's a lot of re-implemented window logic in this PR. If this is still causing issues for peoples workflow then I propose we wrap the folding in a pcall or detect that the chat buffer is not visible and don't run any of the folding logic.

amadeus:
I agree with oli in that re-implementing temp window is not worth it.

What about simply a check if window does not exist, attempt the chat buffer restore call from #1816

I do agree it also makes sense to wrap folding in a pcall in general just in case that logic can fail for any other reason

PR #1798, opened 2025-07-07
feat(keymaps): sort options

powerman:
## Description

Amount of keymaps is increasing with time (I've 26 now), and using Options window (`?`) became complicated because the order is different on each Neovim run. This PR adds basic sorting to keymaps/vars/tools.

## Screenshots

изображение
изображение

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/keymaps.lua
+++ b/lua/codecompanion/strategies/chat/keymaps.lua
@@ -64,6 +64,22 @@ M.options = {
       return desc
     end
 
+    local function sorted_pairs(tbl, comp)
+      local keys = {}
+      for k in pairs(tbl) do
+        table.insert(keys, k)
+      end
+      table.sort(keys, comp)
+      local i = 0
+      return function()
+        i = i + 1
+        local key = keys[i]
+        if key ~= nil then
+          return key, tbl[key]
+        end
+      end
+    end
+
     -- Workout the column spacing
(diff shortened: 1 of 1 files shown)


lua/codecompanion/strategies/chat/keymaps.lua
```diff
@@ -108,7 +124,11 @@ M.options = {
     -- Keymaps
     table.insert(lines, "### Keymaps")
 
-    for _, map in pairs(keymaps) do
+    local function comp_keymaps(a, b)
```
olimorris: What does comp keymaps mean?

olimorris:
This is a great PR! Very much appreciated.

lua/codecompanion/strategies/chat/keymaps.lua
powerman: It's a short of comparator/comparison function. To me it sounds like usual term in Lua world, used in many similar places including [table.sort()](https://neovim.io/doc/user/luaref.html#table.sort()). Can be renamed, if you don't like the name.

olimorris: Let's be explicit and call it `compare_keymaps` if that's what it does. That's generally how I try and do things in CodeCompanion. I'll go for months without looking at things like this you see.

powerman: No prob. Done.

PR #1809, opened 2025-07-10
Fix/watcher on unlisted buffer deletion

Davidyz:
## Description

The deletion of unlisted buffers doesn't seem to trigger `BufDelete`, and therefore doesn't trigger `watchers:unwatch(bufnr)`.

I use the watchers to track _unlisted_ scratch buffers that contain information generated from the [dap extension](https://github.com/Davidyz/codecompanion-dap.nvim), so that when the tools are called and the results are mostly the same, only the diffs are sent to the LLM.

When I delete the scratch buffers following the termination of a DAP session, the watchers don't unwatch them due to the aforementioned reason.

## Checklist


olimorris:
That is an awesome catch! Thanks for PR'ing this.

PR #1815, opened 2025-07-10
fix(chat): start_in_insert_mode with telescope

powerman:
## Description

There is a conflict between telescope and `display.chat.start_in_insert_mode = true` option.
This option works when new Chat is opened using `:CodeCompanionChat` and works when new Chat is opened from Action Palette without telescope but does not work when new Chat is opened from Action Palette with telescope.

## Related Issue(s)

PR #1702 is partially related to this issue - the root cause for both is same: telescope enter/leave INSERT mode and because of this other plugins should use `vim.schedule` before changing mode after leaving telescope picker.

## Checklist


olimorris:
Happy to merge this too. Could we just add a comment explaining why we're using vim.schedule here?

powerman:
Sure, good idea! Done.

PR #1816, opened 2025-07-10
add restore function to ensure a chat buffer can be made visible

olimorris:
## Description

This ensures that the chat buffer is visible after a user accepts an edit with the `insert_edit_into_file` tool.

## Related Issue(s)

#1788

## Checklist

### DIFF
--- a/lua/codecompanion/init.lua
+++ b/lua/codecompanion/init.lua
@@ -209,6 +209,25 @@ CodeCompanion.toggle = function()
   chat.ui:open({ toggled = true })
 end
 
+---Make a previously hidden chat buffer, visible again
+---@param bufnr integer
+---@return nil
+CodeCompanion.restore = function(bufnr)
+  if not bufnr or not api.nvim_buf_is_valid(bufnr) then
+    return log:error("Chat buffer %d is not valid", bufnr)
+  end
+
+  local chat = require("codecompanion.strategies.chat").buf_get_chat(bufnr)
+  if not chat then
+    return log:error("Could not restore the chat buffer")
+  end
+
+  if chat.ui:is_visible() then
+    return
+  end
+  chat.ui:open()
+end
+
 ---Return a chat buffer
 ---@param bufnr? integer
 ---@return CodeCompanion.Chat|table
--- a/lua/codecompanion/strategies/chat/agents/tools/insert_edit_into_file.lua
+++ b/lua/codecompanion/strategies/chat/agents/tools/insert_edit_into_file.lua
@@ -156,7 +156,9 @@ local function edit_buffer(bufnr, chat_bufnr, action, output_handler, opts)
             vim.cmd("silent! w")
           end)
         end)
+        -- NOTE: This is required to ensure folding works for chat buffers that aren't visible
         codecompanion.restore(chat_bufnr)
+
         return output_handler(success)
       end
       return output_handler({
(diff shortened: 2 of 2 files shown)


amadeus:
Just did a quick test, and this branch seems to fix the original issue perfectly. Also I think it'll be a nice UX improvement overall! 👍

olimorris:
Great suggestion on your part, @amadeus! Thanks for testing this.

Jufralice:
Hi @olimorris @amadeus ,

Thanks for this PR!

It does indeed fix the original issue when we accept the proposed changes. But when I tested when rejecting the changes, I still have that "Cannot create fold with current 'foldmethod'" error.
@amadeus, @ThorstenRhau, does it work for you when you reject the proposed changes?

A quick fix would be to also add the 'codecompanion.restore(chat_bufnr)' outside of the "if result.accepted then" clause.

I believe that it make sense to also reopen the chat buffer on changes rejection right?
If this fix is ok for you, here is the PR: #1834 

Thanks a lot.

ThorstenRhau:
No, it currently does not work well. I asked the chat and inline codecompanion.nvim to write a python program showing argparse. The result can be seen in the attached screenshots.

Maybe make the code folding an optional and configurable parameter that is off by default?

Screenshot 2025-07-14 at 12 29 06

Screenshot 2025-07-14 at 12 30 52

Screenshot 2025-07-14 at 12 32 51

This is the configuration I use:

```lua
---@module "lazy"
---@type LazySpec
return {
  'olimorris/codecompanion.nvim',
  dependencies = {
    'nvim-lua/plenary.nvim',
    'nvim-treesitter/nvim-treesitter',
  },
  cmd = {
    'CodeCompanion',
    'CodeCompanionActions',
    'CodeCompanionChat',
    'CodeCompanionCmd',
  },
  keys = {
    { '<leader>ci', '<cmd>CodeCompanion<cr>', desc = 'CodeCompanion Inline Assistant' },
    { '<leader>cc', '<cmd>CodeCompanionChat Toggle<cr>', desc = 'CodeCompanion Chat Buffer' },
    { '<leader>cx', '<cmd>CodeCompanionCmd<cr>', desc = 'CodeCompanion CMD Generator' },
    { '<leader>cX', '<cmd>CodeCompanionActions<cr>', desc = 'CodeCompanion Actions Palette' },
  },
  opts = {
    adapters = {
      openai = function()
        return require('codecompanion.adapters').extend('openai', {
          env = { api_key = vim.env.OPENAI_API_KEY },
          parameters = {
            model = 'gpt-4.1-mini',
            temperature = 0.1,
            stream = true,
            max_tokens = 8192,
            top_p = 1,
            presence_penalty = 0,
            frequency_penalty = 0,
          },
        })
      end,
    },
    strategies = {
      chat = { adapter = 'openai' },
      inline = { adapter = 'openai' },
      cmd = { adapter = 'openai' },
    },
  },
}
```

Jufralice:
Hi @ThorstenRhau ,
Yes, I think that this PR and the #1834 are not real fixes for the folding issue that we encounter. 
They are kind of a workaround that only works for the @insert_edit_into_file tool but we will still have the issue for any other scenarios where a fold is applied when the chat buffer is not visible. 
For example I just tested the new edit_file tool of MCP Hub (see #1833) and I obviously again get the folding error if my chat buffer is not visible when I accept or reject the change. 
It will also occur if we close our chat buffer during a tool processing...

amadeus:
A similar type of fix needs to be added for reject changes, the chat buffer should be restored to maintain parity with accept

Jufralice:
> A similar type of fix needs to be added for reject changes, the chat buffer should be restored to maintain parity with accept

Hi @amadeus ,
Yes it has been done with #1834

PR #1827, opened 2025-07-12
fix(symbols): also use file contents to detect filetype

toupeira:
## Description

This makes the `/symbols` slash command work with files that don't have an extension, e.g. a Python script without `.py`.

Since we already load the file contents to get the symbols we can just pass it to `vim.filetype.match` to improve the detection.

Use `unpack()` in the tests to get multiple return values through `lua_get()`, which only seems to support returning a single value.

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/slash_commands/helpers.lua
+++ b/lua/codecompanion/strategies/chat/slash_commands/helpers.lua
@@ -22,8 +22,22 @@ end
 ---@param target_kinds? string[] Optional list of symbol kinds to include (default: all)
 ---@return table[]|nil symbols Array of symbols with name, kind, start_line, end_line
 ---@return string|nil content File content if successful
+---@return string|nil filetype File type if successful
 function M.extract_file_symbols(filepath, target_kinds)
-  local ft = vim.filetype.match({ filename = filepath })
+  local ok, content = pcall(function()
+    return Path.new(filepath):read()
+  end)
+
+  if not ok then
+    log:error("[chat::slash_commands::helpers] Could not read the file at %s", filepath)
+    return nil, nil, nil
+  end
+
+  local ft = vim.filetype.match({
+    filename = filepath,
+    contents = vim.fn.split(content, "\n"),
+  })
+
   if not ft then
     local base_name = vim.fs.basename(filepath)
     local split_name = vim.split(base_name, "%.")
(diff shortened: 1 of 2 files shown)


lua/codecompanion/strategies/chat/helpers.lua
```diff
@@ -105,8 +105,21 @@ end
 ---@param target_kinds? string[] Optional list of symbol kinds to include (default: all)
 ---@return table[]|nil symbols Array of symbols with name, kind, start_line, end_line
 ---@return string|nil content File content if successful
+---@return string|nil filetype File type if successful
 function M.extract_file_symbols(filepath, target_kinds)
-  local ft = vim.filetype.match({ filename = filepath })
+  local ok, content = pcall(function()
+    return path.new(filepath):read()
```
olimorris: I don't like the fact we're reading a file regardless of whether we have a filetype match or not. Unless I'm mistaken, there is going to be a latency penalty for _all_ users if I merge this PR

toupeira: Well yeah, but there's only a penalty if you actually call `/symbols` on an unsupported file. I imagine that's something you'd only do by accident?

toupeira: Hey @olimorris, sorry I forgot about this PR 😀 Happy to rebase, but wanted to first check in with you if you're okay with the current approach.

As I said above there's only a performance penalty on unsupported files, in the old code if we got a match we'd end up reading the file anyway. I feel like that's an acceptable trade-off vs. not working at all on files without extensions. WDYT?

olimorris: I hear you. Let's go for it 👍🏼

toupeira: @olimorris hmm I see now that `tests/strategies/chat/test_helpers.lua` was deleted in 0a5c13c, was that deliberate or an over-eager LLM? :grinning: 

Let me know if I should add the file back.

edit: The rest is rebased now and tested manually.

toupeira:
@olimorris friendly ping :grinning: 

This is ready for merging, unless you want to restore the deleted helper tests.

olimorris:
Thanks so much for this 👍🏼

PR #1829, opened 2025-07-13
Native Ollama API

Davidyz:
## Description

This PR implements a new Ollama adapter that uses [native API](https://github.com/ollama/ollama/blob/main/docs/api.md). 

The main advantage of this is that it supports the [thinking](https://github.com/ollama/ollama/blob/main/docs/api.md#parameters-1) field that allows concealing the thinking process for supported models (like qwen3 and deepseek r1).

This PR also enables the use of other [hyperparameters](https://github.com/ollama/ollama/blob/main/docs/modelfile.md#valid-parameters-and-values). Many of these options, for example `num_ctx`, are not supported in the openai-compatible API, so the existing adapter that uses openai methods won't work with them.

## Screenshots

Screenshot of qwen3 with thinking enabled and reasoning hidden from the chat buffer:
image

I put the native API in a separate file because it makes testing easier, but if you want to just replace the original `ollama.lua` with the new one, I'm also ok with it.
I'll add the tests and relevant docs once it's been decided.

## Checklist


olimorris:
I think this should replace the default Ollama adapter as long as we have plenty of test coverage and it follows the structure for the existing adapter tests. Thanks for this, looks awesome

Davidyz:
> it follows the structure for the existing adapter tests

I'd need to rewrite the stubs, though. 

While we're at it, I'll also try to get vision working for ollama. From the API documentation, it looks like the images are also base64 encoded, so it probably won't be too hard.

Davidyz:
Added image/vision support (tested with gemma3).

image

Davidyz:
Seems like the native `api/tags` endpoint contains model capability (vision, thinking, etc.). I'll try to rewrite the `get_models` function so that it can automatically detect whether a model supports thinking/vision.

Davidyz:
Although some hyperparams (`microstat`, `logit_bias`, etc.) were dropped, I think we can still mark this as a NON-breaking change because [the openai-compatible endpoints](https://github.com/ollama/ollama/blob/main/docs/openai.md#v1chatcompletions) didn't support these anyways. They were simply ignored by the server, even if the users had these in their adapter config.

olimorris:
Thanks so much for this 🎁. It's a serious undertaking.

Tested with Qwen3 and function calling and works like a charm.

Issue #1823, opened 2025-07-11
[Bug]: "Cannot create fold with current 'foldmethod'" error when the chat buffer is not visible


Jufralice:
### Pre-submission checklist

- [x] I have read the [documentation](https://codecompanion.olimorris.dev)
- [x] I have updated the plugin and all dependencies to the latest versions
- [x] I have searched for existing [issues](https://github.com/olimorris/codecompanion.nvim/issues) and [discussions](https://github.com/olimorris/codecompanion.nvim/discussions)
- [x] My issue is not a minor or cosmetic quirk (e.g., formatting, spacing, or other non-functional details)

### Neovim version (nvim -v)

0.11.1

### Operating system/version

macOs 15.5

### Adapter and model

Gemini and gemini-2.5-flash-preview-05-20

### Describe the bug

This bug was apparently introduced by the recent feature that folds the tool outputs into the chat buffer.

When you accept or reject a proposed change to a file when your chat buffer is not open (I use to have my chat buffer in floating window, I then always toggle it off when I need to accept or reject changes to a buffer) you will receive that error:

Error executing lua callback: ...nion.nvim/lua/codecompanion/strategies/chat/ui/tools.lua:128: Error executing lua: vim/_editor.lua:0: User Autocommands for "CodeCompanionDiffAccepted"..script nvim_exec2() called at User Autocommands for "CodeCompanionDiffAccepted":0, line 1: Vim(fold):E350: Cannot create fold with current 'foldmethod'

Don't hesitate to let me know if you need more info or if you cannot reproduce it.

### Steps to reproduce

- ask the LLM to use the @insert_edit_info_file tool to change anything to your #{buffer}
- when the LLM is done (but before you accept or reject the proposed changes), close the chat buffer
- accept or reject the proposed changes
=> you should see the error

### Expected behavior

There should not be an error, and the fold should be there when you toggle back on your chat buffer.

### minimal.lua file

```lua
I have the feeling that the above reproduction steps should be enough. But don't hesitated to let me know if you need a minimal.lua file anyway
```

### Minimal reproduction confirmation

- [x] Yes, I have tested and provided a `minimal.lua` file that reproduces the issue

olimorris:
Please see #1788 and PRs  #1795 and #1816 which are underway. I'm going to lock this as active discussion is taking place across all.

PR #1834, opened 2025-07-14
fix(chat): also restore the chat buffer on edit changes rejection. This fix the folding error also on change rejection.

Jufralice:
## Description

This is a fix for #1816 which fixed #1788 and #1823 but only when user accepted the proposed edit changes.
I believe that it make sense to also restore the chat buffer on change rejection.

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/agents/tools/insert_edit_into_file.lua
+++ b/lua/codecompanion/strategies/chat/agents/tools/insert_edit_into_file.lua
@@ -149,6 +149,7 @@ local function edit_buffer(bufnr, chat_bufnr, action, output_handler, opts)
 
     -- Wait for the user to accept or reject the edit
     return wait.for_decision(diff_id, { "CodeCompanionDiffAccepted", "CodeCompanionDiffRejected" }, function(result)
+      local response
       if result.accepted then
         -- Save the buffer
         pcall(function()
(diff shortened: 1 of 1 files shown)


lua/codecompanion/strategies/chat/agents/tools/insert_edit_into_file.lua
```diff
@@ -149,22 +149,24 @@ local function edit_buffer(bufnr, chat_bufnr, action, output_handler, opts)
...
             vim.cmd("silent! w")
           end)
         end)
-        -- NOTE: This is required to ensure folding works for chat buffers that aren't visible
-        codecompanion.restore(chat_bufnr)
-
-        return output_handler(success)
+        response = success
+      else
+        response = {
```
olimorris, line 162: Can we put this response in the output handler. No need for a separate var

Jufralice: This was to be more DRY and avoid the duplication of the 'codecompanion.restore' line.
You mean that you'd prefer something like that:
```
     if result.accepted then
        -- Save the buffer
        pcall(function()
          api.nvim_buf_call(bufnr, function()
            vim.cmd("silent! w")
          end)
        end)
        -- NOTE: This is required to ensure folding works for chat buffers that aren't visible
        codecompanion.restore(chat_bufnr)

        return output_handler(success)
      end
      -- NOTE: This is required to ensure folding works for chat buffers that aren't visible
      codecompanion.restore(chat_bufnr)
      return output_handler({
        status = "error",
        data = result.timeout and "User failed to accept the changes in time" or "User rejected the changes",
      })
```
?

olimorris: Yep

olimorris: Ignore me. What you've done is fine.

PR #1857, opened 2025-07-17
fix(tools): handle missing action in get_changed_files

jinzhongjia:
I meet this error, this maybe fix it ?

```
You made an error in calling the get_changed_files tool: "...n.nvim/lua/codecompanion/strategies/chat/agents/init.lua:173: Expected value but found T_END at character 1"

...anion/strategies/chat/agents/tools/get_changed_files.lua:43: attempt to index local 'action' (a nil value)
```

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/agents/tools/get_changed_files.lua
+++ b/lua/codecompanion/strategies/chat/agents/tools/get_changed_files.lua
@@ -40,6 +40,12 @@ end
 ---@param opts {max_lines: number}
 ---@return {status: "success"|"error", data: string}
 local function get_changed_files(action, opts)
+  if not action then
+    return {
+      status = "error",
+      data = "Invalid arguments provided to get_changed_files tool",
+    }
+  end
   local states = action.source_control_state or { "unstaged", "staged", "merge-conflicts" }
   local output = {}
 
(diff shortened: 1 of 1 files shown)


olimorris:
Can you share the tool call from the LLM and provide the adapter and model?

EDIT: Can you stick to the PR template as well, adhering to the headers. It's important that we're able to trace the reasoning for decision.

jinzhongjia:
I can't reproduce this problem stably, and temporarily convert it to draft mode

ravitemer:
I've experienced this issue as well. In my case, it happens when Claude tries to parallel call multiple tools.

It tries to send the calls as:

```
{
      {
        arguments = '{"arg1": ""}',
        name = "github__get_issue_comments"
      },
      id = "toolu_vrtx_01TpzKpbTZoP3RF6NQ3i56kn",
      type = "function"
}{ second call ... }
```

This causes JSON parsing issues. However, the first call almost works fine - maybe due to during streaming the first function part segment is valid? But the second call fails with T_END error etc.

It seems to be something to do with the model rather than the plugin I guess.

jinzhongjia:
Since I can't reproduce, I temporarily close the PR

PR #1852, opened 2025-07-16
Failed/cancelled/rejected tool calls no longer block the execution queue

Davidyz:
## Description

This PR makes sure all `tool_calls` have a corresponding response by:

- make sure interrupted tool calls doesn't abort the subsequent tool calls.
- add fallback tool output handlers so that tools always send something to the LLM to indicate the result status.

## Related Issue(s)

Fix #1850 

## Checklist

- [x] I've read the [contributing](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md) guidelines and have adhered to them in this PR
- [ ] I've updated `CodeCompanion.has` in the [init.lua](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/init.lua#L239) file for my new feature
- [ ] I've added [test](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md#testing) coverage for this fix/feature
- [x] I've updated the README and/or relevant docs pages
- [ ] I've run `make all` to ensure docs are generated, tests pass and my formatting is applied

PR #1865, opened 2025-07-20
Use func executor for cmd-based tools

Davidyz:
## Description

This PR drops the cmd tool executor and uses a wrapper to convert cmd-based tools to function-based tools that call `vim.system`.

## Related Issue(s)

A followup of #1852 

## Checklist


tests/strategies/chat/agents/executor/test_cmd.lua
```diff
@@ -91,26 +91,26 @@ T["Agent"]["cmds"]["can run multiple commands"] = function()
...
-        }
-      },
-    }
-    agent:execute(chat, tool)
-    vim.wait(100)
-  ]])
-
-  h.eq({ testing = true }, child.lua_get("agent.chat.tools.flags"))
-end
+-- T["Agent"]["cmds"]["can set test flags on the chat object"] = function()
```
Davidyz: Not sure what the flag does, so I'm leaving this out for now.

Davidyz: nvm just got this working (as in, the test passed)

Davidyz:
@olimorris the "tool transformer" is done. Since it's (almost) transparent, I think we can keep the cmd tool tests. I haven't done much about the executor (both the base class and the func executor) because it's more about design choices. I didn't merge them because we might want to create new executors in the future.

For example, I think it be helpful to have an "adapter-based" tool executor (in a separate PR, ofc). It's mostly because:
1. currently it's kinda mouthful to invoke an LLM when calling a tool (all the param handling, etc.), and we can put the param handling and message building logic into the executor to simplify this process for tool devs.
2. for tools that sends a lot of requests (my dap extension, for example), the adapter-based executor can implement a throttling mechanism (my toy implementation here: https://github.com/Davidyz/codecompanion-dap.nvim/blob/main/lua/codecompanion/_extensions/dap/timer.lua) that avoids sending requests too fast.

About the adapter-based executor, @ravitemer any thoughts on this from a tool dev's perspective? I'm asking because I saw you also had to handle this by yourself in codecompanion-history.

ravitemer:
Thanks @Davidyz for fixing the #1852 and this. 

> For example, I think it be helpful to have an "adapter-based" tool executor (in a separate PR, ofc). It's mostly because:

> About the adapter-based executor, @ravitemer any thoughts on this from a tool dev's perspective? I'm asking because I saw you also had to handle this by yourself in codecompanion-history.

I remember olimorris mentioning about a background strategy for making it easy to send custom requests with adapters. Would that cover this? I'm afraid I've no idea how the adapter-based tool executor would look that's on your mind.

olimorris:
This is wayyyyyy more cleaner than I anticipated it being. Great job @Davidyz.

> For example, I think it be helpful to have an "adapter-based" tool executor (in a separate PR, ofc)

Potentially agree on this. I put on the roadmap my plan to allow an LLM to use orchestrate other LLMs and an executor _could_ be one way to handle this. Ultimately, we'd end up in a situation where we might need to spawn additional queues so that could change this whole setup completely.

It would most definitely use the background strategy, to @ravitemer's point.

From my side, this is quite far into the future as I need to get agent mode out the door first.

olimorris:
Running `pytest` with the `@{cmd_runner}` tool I notice that no error output is being shared back with the tool

olimorris:
I've tested agentic workflows and they're now working as expected.

I'll continue to play with this for another day and update the docs.

Davidyz:
> Ultimately, we'd end up in a situation where we might need to spawn additional queues so that could change this whole setup completely.

Let's leave it to the future then. Maybe after the background strategy.

PR #1533, opened 2025-05-30
fix: support endpoint redirection in copilot

80avin:
## Description

On our enterprise version of copilot, all requests other than authorization go to different server communicated by the `<auth_response>.endpoints.api` property.

This PR implements supporting the same endpoint redirection.

## Checklist

- [x] I've read the [contributing](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md) guidelines and have adhered to them in this PR
- [ ] I've updated `CodeCompanion.has` in the [init.lua](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/init.lua#L239) file for my new feature
- [ ] I've added [test](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md#testing) coverage for this fix/feature
- [ ] I've updated the README and/or relevant docs pages
- [x] I've run `make all` to ensure docs are generated, tests pass and my formatting is applied

PR #1873, opened 2025-07-21
Add support for GitHub Copilot Enterprise API

dyamon:
## Description

In the Copilot adapter code, the URL `github.com` is hardcoded in a few places. This makes it possible to use the plugin with other providers, namely Github Copilot Enterprise instances.

The only required configuration is the following:

```lua
local p = require 'codecompanion'
p.setup {
  adapters = {
    copilot = function()
      return require("codecompanion.adapters").extend("copilot", {
        opts = {
          provider_url = "acme.ghe.com",            -- defaults to 'github.com'
        },
      })
    end,
  },
}
```

The `provider_url` is used for token retrieval from `$XDG_CONFIG_HOME/github-copilot/{apps,hosts}.json` and authentication.
Thanks to #1533 other queries should be correctly redirected to the (possibly custom) API endpoint.

## Related Issue(s)

N.A.

## Screenshots

N.A.

## Checklist

## Notes

I'm still testing this and might have missed something.
Also, there are some parts of the code that need to be parametric on user options, but I'm not sure what is the best way to inject them. Maybe you can help me with that.

### DIFF
--- a/lua/codecompanion/adapters/copilot/init.lua
+++ b/lua/codecompanion/adapters/copilot/init.lua
@@ -63,7 +63,7 @@ end
 ---specifically for GitHub Codespaces. If not found, it then attempts to load
 ---the token from configuration files located in the user's configuration path.
 ---@return CopilotOAuthToken
-local function get_token()
+local function get_token(self)
   if _oauth_token then
     return _oauth_token
   end
@@ -94,7 +94,7 @@ local function get_token()
 
       local userdata = vim.json.decode(userdata)
       for key, value in pairs(userdata) do
-        if string.find(key, "github.com") then
+        if key:find(self.opts.provider_url, 1, true) == 1 then
           return value.oauth_token
         end
       end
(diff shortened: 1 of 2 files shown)


lua/codecompanion/adapters/copilot/init.lua
```diff
@@ -163,11 +169,13 @@ return {
...
+    provider_url = 'github.com',
   },
   features = {
     text = true,
     tokens = true,
   },
+  api = "https://api.githubcopilot.com",
   url = "https://api.githubcopilot.com/chat/completions",
   env = {
     ---@return string|nil
```
dyamon, line 181: Here `authorize_token` is called without `self` and I'm not sure if such a refactoring is possible.

dyamon:
Also in `helpers.lua` the function `get_copilot_stats` should have a parametric API URL and should be passed the adapter configuration. Again, this doesn't seem to be straightforward with the current way the function is called.

olimorris:
My advice would be to be roll your own GitHub Copilot Enterprise adapter and just reference the handlers on the Copilot adapter.

We've built the Copilot adapter without any design considerations for the Enterprise version and I suspect you'll spend more time refactoring and me testing than if we went a separate route.

dyamon:
> My advice would be to be roll your own GitHub Copilot Enterprise adapter and just reference the handlers on the Copilot adapter.

This makes sense. Just like Copilot uses the `openai` adapter under the hood, I guess?

olimorris:
Closing this based on our discussion above.

PR #1789, opened 2025-07-06
Fix: copilot adapter unauthorized token expired

somnam:
## Description
Fix `Unauthorized: token expired` error after prompting inactivity.

The issue seems to be caused by [models cache](https://github.com/olimorris/codecompanion.nvim/blob/3527a9c85b58a8db2fee6d7dcde72077ba790a4a/lua/codecompanion/config.lua#L33) refresh after some time. The copilot adapter function [is given a _github_token](https://github.com/olimorris/codecompanion.nvim/blob/3527a9c85b58a8db2fee6d7dcde72077ba790a4a/lua/codecompanion/adapters/copilot/init.lua#L303C48-L303C71) but that token can be outdated and refreshed by `get_and_authorize_token` call inside the function. Despite this the function would still use the outdated token from input when calling the models endpoint.

## Related Issue(s)

- Fixes #1052

## Checklist

- [x] I've read the [contributing](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md) guidelines and have adhered to them in this PR
- [ ] I've updated `CodeCompanion.has` in the [init.lua](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/init.lua#L239) file for my new feature
- [x] I've added [test](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md#testing) coverage for this fix/feature
- [ ] I've updated the README and/or relevant docs pages
- [ ] I've run `make all` to ensure docs are generated, tests pass and my formatting is applied

PR #1899, opened 2025-07-27
fix(copilot): correctly retrieve token in get_models

bassamsdata:
## Description

There’s an issue where Copilot doesn’t retrieve models correctly because the `get_models` GitHub endpoint expects a GitHub token rather than an OAuth token. It’s still surprising that the API didn’t return any errors when we used the wrong token.

This PR addresses the issue, but I would appreciate it if you could test it before merging to ensure everything is working smoothly.

Thank you

## Related Issue(s)

#1874

## Checklist

### DIFF
--- a/lua/codecompanion/adapters/copilot/helpers.lua
+++ b/lua/codecompanion/adapters/copilot/helpers.lua
@@ -30,7 +30,7 @@ end
 ---Get a list of available Copilot models
 ---@param adapter CodeCompanion.Adapter
 ---@param get_and_authorize_token_fn function Function to get and authorize token
----@param authorize_token_fn function Function to get fresh github token
+---@param authorize_token_fn function Function to get fresh Github token
 ---@return table
 function M.get_models(adapter, get_and_authorize_token_fn, authorize_token_fn)
   if _cached_models and _cache_expires and _cache_expires > os.time() then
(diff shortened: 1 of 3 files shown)


Eckii24:
Oh good catch - it's actually not the URI, it's the token. Nice!

Working for me again:
image

I'm wondering if it nevertheless would make sense to use the endpoint provided by the token:
E.g. for me, my company is blocking requests to api.githubcopilot.com and only letting api.business.githubcopilot.com requests through, to ensure everyone is using the correct policies.

```json
{
  [...]
  endpoints = {
    api = "https://api.individual.githubcopilot.com",
    ["origin-tracker"] = "https://origin-tracker.individual.githubcopilot.com",
    proxy = "https://proxy.individual.githubcopilot.com",
    telemetry = "https://telemetry.individual.githubcopilot.com"
  },
  token = "...",
  [...]
}
```

olimorris:
Agree with @Eckii24, we should be using the endpoint that's specified in the token. It's become clear that this is unique to some folk and was what we were doing before the refactor.

bassamsdata:
@olimorris, noted, I’ve reverted it to the state before PR #1789.  
@Eckii24 This should work fine now, but please let me know if anything else is needed.

olimorris:
Tested this out and can confirm that `o4-mini` and `gemini-2.5-pro` have returned.

Thanks @bassamsdata!

PR #1903, opened 2025-07-27
feat(chat): Allow for multiline welcome messaage.

slg95:
## Description

The virt_text option for nvim_buf_set_extmark() places the virtual text all on one line, even if the virtual text contains line breaks. This makes specially formatted custom welcome messages not wrap in the way a user may want.

This replaces the set_virtual_text() in
lua/codecompanion/strategies/chat/ui/init.lua with a new function set_virtual_text_lines() in lua/codecompanion/utils/ui.lua. The old set_virtual_text() in lua/codecompanion/utils/ui.lua was not removed in case code outside of this project depends on it, even though it's never called in this project.

Single line welcome messages are unchanged. Multiline welcome messages will be placed above the user's cursor.

## Related Issue(s)

Closes #1894

## Checklist

- [X] I've run `make all` to ensure docs are generated, tests pass and my formatting is applied

### DIFF
--- a/lua/codecompanion/strategies/chat/ui/init.lua
+++ b/lua/codecompanion/strategies/chat/ui/init.lua
@@ -54,7 +54,13 @@ function UI.new(args)
     once = true,
     desc = "Clear the virtual text in the CodeCompanion chat buffer",
     callback = function()
-      api.nvim_buf_clear_namespace(self.chat_bufnr, CONSTANTS.NS_VIRTUAL_TEXT, 0, -1)
+      -- Clear intro message if it exists
+      if self.intro_extmark_ids and self.intro_line_count then
+        self:clear_intro_message()
+      else
+        -- Otherwise just clear the namespace
+        api.nvim_buf_clear_namespace(self.chat_bufnr, CONSTANTS.NS_VIRTUAL_TEXT, 0, -1)
+      end
     end,
   })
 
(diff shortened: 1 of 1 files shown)


olimorris:
I don't like the way the cursor has to move around the virtual text:

https://github.com/user-attachments/assets/1c441f23-b3c3-43c0-88a1-dc852edcbe7e

Unlike currently:

2025-07-27 22_45_27 - WezTerm@2x

slg95:
Thanks for the feedback! I was playing around with the `virt_lines` too a bit and I do see what you mean. This is why I have the `virt_text` for the single line welcome text. The single line welcome text looked weird with `virt_lines`.

If we want to use the `virt_lines` approach, I believe that is how it has to work with the cursor skipping around. Another option could be 

1. Count the number of lines in the welcome text (x lines)
2. Fill the buffer with x lines
3. Create a separate `virt_text` for each lines
4. Clear all of the `virt_text`s when it's time to no longer show the virtual text.

I can try that as another option and then we can compare. Unless you have any better ideas :)

olimorris:
I don't to merge a PR that causes the cursor to skip around, it doesn't feel great from a UX perspective. If that can't be achieved (and I use extmarks and virtual text so infrequently in Neovim that I'm no expert) then I'll revert back to how it is currently.

slg95:
> I don't to merge a PR that causes the cursor to skip around, it doesn't feel great from a UX perspective. If that can't be achieved (and I use extmarks and virtual text so infrequently in Neovim that I'm no expert) then I'll revert back to how it is currently.

Totally understand, the UX needs to be right to add any new feature.

I updated my PR with a temporary commit for a quick prototype of what I was thinking. I let AI write it and didn't really check the output, so I will clean it up if you think the UX is acceptable.

olimorris (changes requested):
Image

### Summary
This PR enhances the chat UI by allowing multi-line intro messages with virtual text overlays, inserting empty buffer lines as needed, and providing a cleanup routine (`clear_intro_message`). It also factors out a new utility `set_virtual_text_lines` to support both single- and multi-line virtual annotations.

### Strengths
- Introduces support for multi-line intro messages, improving UX for longer prompts.
- Prevents duplicate `InsertEnter` autocmds with an `intro_autocmd_created` flag.
- Provides a comprehensive cleanup in `clear_intro_message` (removes extmarks and inserted lines).
- Adds `set_virtual_text_lines` in `utils/ui.lua` for general-purpose virtual text/virtual lines support.

### Issues Found
1. Buffer Option Handling (init.lua): The code uses `vim.bo[self.chat_bufnr].modifiable` which is invalid because `vim.bo` does not accept a buffer id key. The recommended approach is to use `vim.api.nvim_buf_get_option(bufnr, 'modifiable')` and `vim.api.nvim_buf_set_option(bufnr, 'modifiable', value)`.

2. Removal of `UI:set_virtual_text`: The old method is removed but ensure no remaining code relies on it or else migrate those usages.

3. Docstrings and Parameter Descriptions (utils/ui.lua): The new `set_virtual_text_lines` function lacks detailed documentation about the `opts` parameter fields, overflow behavior, and usage examples.

4. Test Coverage: There are no tests covering the new multi-line intro message support or the new utility function. Adding unit or integration tests is recommended.

5. Buffer-Autocmd Cleanup: Although the autocmd is created with `once = true`, consider storing or managing autocmd ids if future removal is needed.

### Suggestions
- Replace incorrect `vim.bo[self.chat_bufnr]` calls with appropriate Neovim buffer option API calls.
- Verify and update any calls to the deprecated `UI:set_virtual_text` method.
- Enhance documentation of the new utility to clarify options and expected behavior.
- Add test cases for multi-line intro message handling and the utility `set_virtual_text_lines`.
- Consider logging errors or edge cases silently caught in `clear_intro_message` for easier debugging.

Overall, this PR introduces valuable enhancements but requires fixes in buffer option handling and improved documentation before approval.

**Recommendation:** requestChanges

olimorris:
Looking at the PR to date, this is far more complicated than I hoped it would be for something which uses will likely see for a small duration of time. I propose we don't proceed any further with it.

slg95:
Not a problem, thanks for the review!

PR #1917, opened 2025-07-31
fix: set nowait as default for reject

seblyng:
## Description

Following up on this https://github.com/olimorris/codecompanion.nvim/pull/1572. I think it makes sense to set this as default to make rejection of a diff happen immediately. By default this conflicts with the builtin lsp mappings which have `gr` as prefix. Therefore it waits for if a third keypress happens or not.

Probably doesn't make sense to wait for a third keypress while in diff mode? So I think it will enhance the experience of not "lagging" when rejecting changes

## Checklist


olimorris:
This is an awesome catch. Thanks

PR #1923, opened 2025-08-01
fix(ui): Debug Chat: allow `q` and `<Esc>` to be able to close the window

sidkshatriya:
## Description

When I want to quit from Debug Chat, it is tedious to have to do `:q<return>` (three keypresses). It should be possible to just quit using `q` or `Esc`. That feature exists but `ignore_keymaps` is set to `true` when creating the floating window. Set `ignore_keymaps` to false. 

## Checklist


olimorris:
Great spot. Thanks for this.

PR #1949, opened 2025-08-09
feat(adapter): Add Support for Gitlab Duo

Kraust:
## Description

I recently discovered that I have [Gitlab Duo](https://docs.gitlab.com/api/chat/) access and as a CodeCompletaion.nvim user, I felt it necessary to add support for it in my workflow.

I'm relatively new to this, so if there's anything else that needs to be added I'm willing to do it. Just provide me any guidance for anything you're unsatisfied with.

Also, currently  Gitlab Duo has a 1000 character limit per message so the default system prompt does not work so I set it to an empty string in my CodeCompanion config. Any thoughts on how to
work around this (Maybe allowing for a smaller default prompt)?

Gitlab Duo API Reference: https://docs.gitlab.com/api/chat/

Additional Context is interesting. I don't know how this could be integrated into the workflow, but it provides for Agentic coding.

I'm also thinking that concatenating the message list is a god idea, maybe I should always submit the most recent message?

## Related Issue(s)

https://github.com/olimorris/codecompanion.nvim/discussions/1947

## Screenshots

image

## Checklist

### DIFF
--- a/doc/codecompanion.txt
+++ b/doc/codecompanion.txt
@@ -62,7 +62,7 @@ LLMs, in Neovim.
 FEATURES                                      *codecompanion-welcome-features*
 
 -  Copilot Chat <https://github.com/features/copilot> meets Zed AI <https://zed.dev/blog/zed-ai>, in Neovim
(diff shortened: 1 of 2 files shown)


olimorris:
As per #800, I'm keeping adapter scope limited in CodeCompanion. Please share this in the Community discussions though as I'm sure many will find it useful.

Kraust:
Completely understandable. I will publish my own plugin as an extension to code companion. Have already written it like that for testing. Thanks for the quick reply.

PR #1966, opened 2025-08-12
Support for GitHub Copilot Enterprise

dyamon:
## Description

As discussed in #1873, this PR adds support for GitHub Copilot Enterprise. It has been designed as a separate adapter that pulls as many reusable pieces from the standard `copilot` adapter.

I have been using it for a while and it seems to be working just fine.

Here is a minimal configuration for some organization `acme`:

```lua
  adapters = {
    copilot_enterprise = function()
      return require("codecompanion.adapters").extend("copilot_enterprise", {
        opts = {
          provider_url = "acme.ghe.com",
        },
      })
    end,
  },
  strategies = {
    chat = {
      adapter = {
        name = "copilot_enterprise",
        model = "claude-sonnet-4",
      },
    },
    inline = {
      adapter = "copilot_enterprise",
    },
    cmd = {
      adapter = "copilot_enterprise",
    }
  },
})
```

### User statistics

`copilot` supports user statistics through the `/copilot_internal/user` REST API. By debugging the VS Code Copilot extension it seems that the same is available for the Enterprise edition (through the custom API endpoint), but all my attempts returned "404 Not found". This seems to be confirmed in VS Code as well.

This is the popup when logged with a personal account:

image

while this is the same one when logged with an enterprise account:

image

I think the feature is there, but not exposed yet. As such, I decided to take the following steps:

- I slightly generalized the `{get,show}_copilot_stats`, turning them into a more generic `{get,show}_stats`.
- Implemented them for the `copilot_enterprise` adapter.

It will return an error for now, but will possibly be beneficial in the future.

### Testing

Tests have been ported from the `copilot` adapter and adapted to the "enterprise" version, but nothing major has changed. I have been having difficulties running `make test` in my local environment due to my Nix installation on NeoVim. If someone could just run `make all` once for me, it would be super :) I don't foresee any problem, tho.

## Related Issue(s)

N.A.

## Screenshots

N.A.

## Checklist

- [X] I've updated `CodeCompanion.has` in the [init.lua](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/init.lua#L239) file for my new feature
- [X] I've added [test](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md#testing) coverage for this fix/feature
- [X] I've updated the README and/or relevant docs pages

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,7 +29,7 @@ Thank you to the following people:
 ## :sparkles: Features
 
 - :speech_balloon: [Copilot Chat](https://github.com/features/copilot) meets [Zed AI](https://zed.dev/blog/zed-ai), in Neovim
-- :electric_plug: Support for Anthropic, Copilot, GitHub Models, DeepSeek, Gemini, Mistral AI, Novita, Ollama, OpenAI, Azure OpenAI, HuggingFace and xAI LLMs (or [bring your own](https://codecompanion.olimorris.dev/extending/adapters.html))
+- :electric_plug: Support for Anthropic, Copilot, Copilot Enterprise, GitHub Models, DeepSeek, Gemini, Mistral AI, Novita, Ollama, OpenAI, Azure OpenAI, HuggingFace and xAI LLMs (or [bring your own](https://codecompanion.olimorris.dev/extending/adapters.html))
 - :heart_hands: User contributed and supported [adapters](https://codecompanion.olimorris.dev/configuration/adapters#community-adapters)
 - :rocket: [Inline transformations](https://codecompanion.olimorris.dev/usage/inline-assistant.html), code creation and refactoring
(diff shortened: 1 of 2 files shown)


olimorris:
@dyamon sorry for not getting to this sooner.

Mulling this over, I think it makes sense for this to be a community adapter. I have no way of being able to test it so don't want it being part of `main` and users looking to me to support.

dyamon:
Fair enough. I'm moving the code to [codecompanion-copilot-enterprise.nvim](https://github.com/dyamon/codecompanion-copilot-enterprise.nvim). I guess we can mention it in the "Community adapters" section. When it is stable enough I'll open a thread in the Discussions as well.

PR #1696, opened 2025-06-22
feat: list_code_usages tool

lazymaniac:
### Why
To provide precise and exact context required by the LLM to solve coding tasks without guesses and assumptions about code fragments. It's specifically designed to solve smaller and direct tasks, and it achieves good results.

### What
This is a new tool for CodeCompanion called `list_code_usages`. Previously I had it in my personal config, but I decided to create a PR as the tool is more or less polished. It works similarly to the `vectorcode` tool, but instead of querying vector databases, it utilizes semantic knowledge of LSP.

### How it works
1. Finds symbols with LSP workspace/symbol function
2. Finds symbols with grep (this supplements LSP results)
3. Goes to the file where the symbol is found and places the cursor at the symbol position
4. Calls LSP methods: 'textDocument/definition', 'textDocument/references', 'textDocument/implementation', 'textDocument/declaration', 'textDocument/typeDefinition', 'textDocument/hover'
5. For all the results from LSP and grep, uses Tree-sitter locals to find the closest most significant code block (function, class, struct, etc.)
6. If Tree-sitter fails, uses fallback to manually find the code block for results

I know it's a little bit bloated and complex compared to other tools and the general philosophy of this plugin, but it's helpful for me and maybe others can benefit from it as well.

## Screenshots

Screenshot 2025-07-23 at 08 01 32
Screenshot 2025-07-23 at 08 01 50
Screenshot 2025-07-23 at 08 04 38

## Checklist

- [x] I've read the [contributing](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md) guidelines and have adhered to them in this PR
- [x] I've updated `CodeCompanion.has` in the [init.lua](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/init.lua#L239) file for my new feature
- [x] I've added [test](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md#testing) coverage for this fix/feature
- [x] I've updated the README and/or relevant docs pages
- [x] I've run `make all` to ensure docs are generated, tests pass and my formatting is applied

PR #1977, opened 2025-08-14
CodeCompanionComplete command for tab-autocompletion

pratyushmittal:
## Description

This PR adds a `CodeCompanionComplete` command to trigger a code completion on demand.

Though I added it as a test POC (proof-of-concept), I ended us using it extensively.

I use it with this setting in my `init.lua`:

```
vim.keymap.set({ 'i' }, '<C-f>', '<cmd>CodeCompanionComplete<cr>', { noremap = true, silent = true, desc = 'Complete [F]orward' })
```

The above triggers a tab-style code completion on pressing `Ctrl+f` in insert mode. No need to pass any chat instructions. Usually the comments and context around the cursor is sufficient to provide a good completion.

## Screenshots

https://github.com/user-attachments/assets/b4c6acea-9c6b-4c38-9354-beaf9f221b0b

## Checklist


pratyushmittal:
@olimorris I am not sure if this fits within the scope of the library. I personally have been finding this feature very useful.

The current PR is a draft. I can work on improving it if you too find it useful.

Improvements to make:
- Add some visual feedback to show when loading a autocomplete.
- Try to use [`list_code_usages`](https://github.com/olimorris/codecompanion.nvim/pull/1696) tool along with for even better suggestions.

lua/codecompanion/strategies/inline/init.lua
```diff
@@ -469,6 +469,10 @@ function Inline:done(output)
     self:start_diff()
     pcall(vim.cmd.undojoin)
     self:output(json.code)
+    -- After auto-applying inline edits (e.g., no_diff), move cursor to end of insertion
```
olimorris: This is a very nice QOL feature!

olimorris:
Hey @pratyushmittal 

> I am not sure if this fits within the scope of the library

It definitely does and would be an awesome addition. I'm not sure I'd want to have this as a separate command though. As this utilizes the inline assistant, I think it would make sense for it to be `:CodeCompanion complete`. This is because the current "main" commands all map to strategies i.e. `:CodeCompanionChat` and `:CodeCompanionCmd`. It _could_ always be its own context variable `:CodeCompanion #{complete}` (like we have this for `#{buffer}` and `#{chat}` right now). This would keep it nicely structured and contained but appreciate its doing something a bit more than just sharing context so isn't that intuitive to a user.

Bottom line, totally want this in CodeCompanion just need to workout what the UX would be.

> Add some visual feedback to show when loading a autocomplete.

I wouldn't want this directly in CodeCompanion but we could fire some events with the buffer number, line numbers etc so users could hook into that themselves. I will always endeavour for CodeCompanion to not force anything on a user from a UI perspective beyond what happens in the chat buffer.

> Try to use list_code_usages tool along with for even better suggestions.

This would be so good and so powerful, but I suspect this could be quite complicated to achieve. How are you envisioning this would work?

olimorris:
Hey @pratyushmittal hope all is well. I've changed tack with regards to what I want to allow in `main` on CodeCompanion. As such, my preference would be for a feature like this to exist as its own separate plugin.

PR #1989, opened 2025-08-17
fix(tools): Handle empty parameters in tools

Davidyz:
## Description

The current anthropic adapter (and the `Tools` class) aren't happy when the tool call has no parameters. This is problematic when the model chooses to return `""` for "no params".

Addendum:

When the tool has no parameters or all parameters are optional, the model may decide to call it with the argument being `""`, which is not parsable by `vim.json.decode` and causes the error. This PR just converts `""` to `"{}"` so that it doesn't break.

## Related Issue(s)

See details at https://github.com/Davidyz/VectorCode/discussions/270.

## Todo

- [x] test the patch with a live anthropic API call (tested by @bnwa)
- [ ] check whether other adapters need a similar fix

## Checklist


Davidyz:
@olimorris the fix for the anthropic adapter is done (with a new test case adapted from the existing tool streaming test). Could you take a look and lmk whether it looks ok (and whether I should implement empty param tests for the other adapters)?

olimorris:
I've not come across this before.

What's causing it? Is the answer not to update the tool schema of the tool that this is happening in?

Davidyz:
> What's causing it? Is the answer not to update the tool schema of the tool that this is happening in?

When the tool has no parameters or all parameters are optional, the model may decide to call it with the argument being `""`, which is not parsable by `vim.json.decode` and causes the error. This PR just converts `""` to `"{}"` so that it doesn't break.

olimorris:
Got it. I've put that in the initial post in case I reference this again in the future.

Davidyz:
@olimorris I've updated tool tests with new stubs that contain empty string arguments (`""`). From the tests, it seems like there's no need to modify other adapters. I've tested this against Copilot, Gemini, ollama and llama.cpp (openai-compatible) and they're all working fine.

olimorris:
Sorry for holding off on this. I wanted to get `v17.18.0` released as knew that would change the adapters massively. I'll get this merged tomorrow

olimorris:
@Davidyz thanks again for this and all your work recently. Really appreciated!

PR #1993, opened 2025-08-19
feat(adapter): support for OpenAI gpt-5 models

chrisgrieser:
## Description
The gpt-5 models reason by default, making it relevant to adjust `reasoning_effort` since you may not want reasoning for quick simple tasks.

Also, the gpt-5 models introduced a new level for reasoning called `minimal`: https://platform.openai.com/docs/api-reference/responses/object#responses/object-reasoning

Being able to set reasoning effort is especially relevant for the gpt-5 models, since they use medium reasoning effort by default, resulting in considerably slower responses than the gpt-4 models: https://community.openai.com/t/gpt-5-is-very-slow-compared-to-4-1-responses-api/1337859

## Checklist

### DIFF
--- a/lua/codecompanion/adapters/openai.lua
+++ b/lua/codecompanion/adapters/openai.lua
@@ -328,6 +328,9 @@ return {
       ---@type string|fun(): string
       default = "gpt-4.1",
       choices = {
+        ["gpt-5"] = { opts = { has_vision = true, can_reason = true } },
+        ["gpt-5-mini"] = { opts = { has_vision = true, can_reason = true } },
+        ["gpt-5-nano"] = { opts = { has_vision = true, can_reason = true } },
         ["o4-mini-2025-04-16"] = { opts = { has_vision = true, can_reason = true } },
         ["o3-mini-2025-01-31"] = { opts = { can_reason = true } },
         ["o3-2025-04-16"] = { opts = { has_vision = true, can_reason = true } },
@@ -365,6 +368,7 @@ return {
         "high",
         "medium",
         "low",
+        "minimal",
       },
     },
     temperature = {


chrisgrieser:
somewhat related, it seems gpt-5 models require organizational verification if you want to use streaming: #1951

So it's maybe worth a thought of disabling streaming by default for the gpt-5 models, since the gpt-5 models throw an error when you have not verified your org yet. But since that is a debatable decision, I did not add that to this PR yet.

olimorris (changes requested):
Image

### Summary
The PR extends the OpenAI adapter by adding three new GPT-5 variants (`gpt-5`, `gpt-5-mini`, `gpt-5-nano`) and a new `minimal` option for the `top_p` setting. Broadly, these changes follow the existing pattern in the codebase and appear functionally correct.

### Strengths
- Consistent pattern: Each new model entry uses the same `opts = { has_vision = true, can_reason = true }` structure.
- Nonbreaking: These additions are additive and shouldn’t affect existing configurations.
- Minimal diff: The PR makes small, focused changes in one file.

### Issues Found
1. **Documentation Missing** (lua/codecompanion/adapters/openai.lua)
   - No updates to the README or adapter documentation to mention the new GPT-5 models or the `minimal` top_p level.
2. **Test Coverage**
   - There are no accompanying unit tests or integration tests that ensure the new models and `minimal` top_p value are recognized and handled correctly.
3. **Alphabetical Ordering**
   - Existing model keys in the `choices` table appear roughly sorted. The new entries (`gpt-5`, `gpt-5-mini`, `gpt-5-nano`) are appended at the top of the list. Consider inserting them in alphabetical order for consistency.
4. **Runtime Validation**
   - The adapter does not validate that a chosen model is supported by the remote API or provide a clear error if a user selects an unsupported variant.

### Suggestions
- **Documentation**: Update the README and any relevant docs to list the new GPT-5 variants and the new `top_p = "minimal"` option, including any notes on vision and reasoning capabilities.
- **Tests**: Add unit tests to cover:
  - Loading each new model name in the configuration schema.
  - Ensuring `top_p = "minimal"` passes schema validation.
- **Sorting**: Reorder the `choices` table to keep model names in alphabetical order so that future maintainers can more easily find entries.
- **Validation**: Add a runtime check or clear error message when a model that your local environment or the OpenAI API does not support is selected.

Once the above are addressed, this PR will be in good shape to merge.

chrisgrieser:
I think this AI review is not helpful.

1. It points out some correct suggestions, but does not consider that PRs should be atomic (= concerns should be split up into separate PRs.)
  -  Adding tests, runtime validation, and writing docs which, as far as I can see are not present for the current adapter anyway, would be the concern of another PR.
  - The original list of models has neither any chronological nor alphabetical order, so there is no order I could insert the gpt-5 models into. Cleaning up the adapter to sort object keys would be desirable, but that would be another PR.

The AI review is a bit overkill for what is literally a fix consisting of 4 lines of code.

2. Some suggestions are wrong.
  - Ordering the choices logically ("minimal" is lower than "low") makes more sense than to order them alphabetically.
  - `top_p` has not been modified by this PR.

olimorris:
@chrisgrieser presumably this will throw an error if we don't disable streaming? It would probably be helpful to handle the case of no organizational verification in the `on_exit` handler.

Don't worry about the AI review. It's just an automated workflow designed to highlight any obvious flaws in a PR.

chrisgrieser:
Not sure if an extra consideration is needed; if you use streaming without org verification it will fail either way. 

However, I think it might be worth considering to disable streaming by default, though once again that might be an issue for another PR.

olimorris:
> Not sure if an extra consideration is needed;

This PR doesn't currently work if I don't know I need to turn streaming off. I just get a big error message.

> However, I think it might be worth considering to disable streaming by default,

For users who aren't using gpt-5 we're forcing them to change their defaults.

Both of these are two scenarios I can't accept.

olimorris:
Btw...I think it should be as simple as setting `stream = false` for the gpt-5 model choices. This is what we did when o1 came out

chrisgrieser:
Sure thing, disabled streaming by default for the GPT 5 models.

olimorris:
Awesome!! Thanks for PR'ing this. Really appreciated.

PR #2005, opened 2025-08-22
fix: properly render anthropic tool results while restoring chats

ravitemer:
## Description

This is related to the https://github.com/ravitemer/codecompanion-history.nvim/issues/43 issue at the history extension. 

For the anthropic adapter, the content field of a tool result message is a table rather than a plain string e.g:

```lua
{
    content = {
      content = "**`neovim__list_directory` Tool**: Returned the following:\n\n````\n Directory Listing: ````",
      is_error = false,
      tool_use_id = "toolu_012TCg3Xi16Ef8fPdhafFynS",
      type = "tool_result"
    },
    cycle = 2,
    id = 2082665656,
    opts = {
      visible = true
    },
    role = "tool"
  }, 
```

This leads to issues while restoring the chat with messages of this kind while rendering that chat buffer as we assume `msg.content` will always be a string as follows previously:

https://github.com/olimorris/codecompanion.nvim/blob/53b6f5041704edfef3b2d1f6731769d03563bdb9/lua/codecompanion/strategies/chat/ui/init.lua#L282-L285

This PR handles checks for `msg.content.content` field to handle the special anthropic case.

## Related Issue(s)

https://github.com/ravitemer/codecompanion-history.nvim/issues/43

## Checklist


amadeus:
Any movement on this? It essentially means we can't use the history plugin with any anthropic models that utilize a file tool of some kind, which is kind of a bummer.

This is what happens when it tries to resume a saved chat:

CleanShot 2025-08-31 at 00 24 11

ravitemer:
@olimorris Please let me know if you need anything from my side to merge this?

olimorris:
I don't want to have adapter specific renders baked into the UI class.

I can't recall why Anthropic tool output is different but I would propose we look to address that rather than patch in this way.

augustocdias:
@olimorris what would be way to go here? Make the Anthropic adapter set the output as text?

augustocdias:
@olimorris I've tried changing to see what would break and here was the first place where it was expecting a table instead of a string:
https://github.com/olimorris/codecompanion.nvim/blob/03806002044f94f8789b28435efd6d383675ba87/lua/codecompanion/adapters/http/anthropic.lua#L246-L258

igorgue:
Any progress on this? It's been broken for a while.

olimorris:
#2241 should fix this as Anthropic tool output will be stored in the message stack in the same ways as all other tool output

igorgue:
Thank you @olimorris

PR #2014, opened 2025-08-25
fix: default to normal mode after applying inline diff

seblyng:
Problem:

It is annoying to always have to press `<esc>` before being able to either accept or reject the diff with `gda` or `gdr`.

Solution:

Default to normal mode when in applying inline diff to allow users to immediately accept/reject diffs

BTW:

The inline diff PR is a _super_ good improvement. Not only is it very nice that I know don't have to install a different plugin like `mini.diff`, the change in the prompt for how it should insert the patches seems to have helped a lot, making it fail less often now. Previously it would fail _a lot_ for me, but trying to use it now for some hours constantly having it insert code for me, it seems to have gotten _a lot_ better

## Checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,6 +1,6 @@
 ## Description
 
-<!-- Describe the big picture of your changes to communicate to the maintainers why we should accept this pull request. -->
+<!-- If this is a feature request, please describe how it will benefit other CodeCompanion users -->
 
 ## Related Issue(s)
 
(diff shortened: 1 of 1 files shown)


bassamsdata:
I’m not sure why this is happening, I haven’t experienced this myself. I’ll push a PR today (hopefully), and after that, we can see if the issue still occurs before we do the "esc" part.

That said, I’m really glad you like the new inline diff!

On a side note, I’m currently working on the PR that really should have been ready yesterday. But after coming back from vacation, I accidentally deleted a symlink to my local CodeCompanion repo using oil.nvim thinking that Oil will just delete the symlink only, but Oil ended up deleting the entire original directory :( without sending it to trash. So, I lost all of the changes I had made. Bad Oil :(

seblyng:
> So, I lost all of the changes I had made

Oh no, that is not good:(

> I’ll push a PR today (hopefully), and after that, we can see if the issue still occurs before we do the "esc" part.

Sounds good to me👍

seblyng:
@bassamsdata I guess your PR was merged? If so, it didn't fix the issue. Do you always send the request using enter in normal mode? I guess it is still in insert mode for me because I am often using `<C-s>` to send the request

bassamsdata:
Hello @seblyng, thanks for following up.

> I am often using <C-s> to send the request

Ah, that explains it. You’re sending the request from insert mode and staying in insert mode, which is why the chat remains in insert and there’s no issue with the inline diff.

For the diff workflow UX, I think it would be better to switch to normal mode in the diff window even if the user sent the request from insert mode. I’m in favor of this PR because it feels like a more intuitive UX and shouldn’t affect other behavior.

@olimorris, what do you think?

lua/codecompanion/providers/diff/inline.lua
```diff
@@ -48,6 +48,10 @@ function InlineDiff.new(args)
       local winnr = vim.fn.bufwinid(self.bufnr)
       if winnr ~= -1 then
         pcall(api.nvim_win_set_cursor, winnr, { first_diff_line, 0 })
+
+        if vim.api.nvim_get_mode().mode ~= "n" then
```
olimorris: can we change these from `vim.api` to `api`?

olimorris:
Happy to include this. A sensible suggestion

lua/codecompanion/providers/diff/inline.lua
seblyng: Fixed

olimorris:
This is one of those issues where I suspect I wasn't conscious I was even pressing `<esc>`. Good catch @seblyng

PR #2017, opened 2025-08-26
fix: `gpt-5-nano` streaming error

chrisgrieser:
## Description
When using `gpt-5-nano` for the inline strategy, I get this error multiple times:

```
[ERROR] 2025-08-26 12:12:30
Inline output is not supported for non-streaming models
```

However, streaming is explicitly set to `true`. Not sure why that error is occurring, as a temporary fix to get the model to work, disabling `streaming` does appear to work.

## Related Issue(s)
Follow up to #1993

### DIFF
--- a/lua/codecompanion/adapters/http/openai.lua
+++ b/lua/codecompanion/adapters/http/openai.lua
@@ -331,9 +331,10 @@ return {
         -- gpt-5 and gpt-5-mini (but not gpt-5-nano) require organizational
         -- verification with biometric data when streaming is enabled:
         -- https://news.ycombinator.com/item?id=44837367
+        -- (see also #2017)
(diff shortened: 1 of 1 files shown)


olimorris:
I think you can remove `stream` from gpt-5-nano. The inline logic should turn it on/off as required. Unsure why this wouldn't work. We might have to rename this to `can_stream` for other models to avoid any clashes

chrisgrieser:
> I think you can remove stream from gpt-5-nano. 

sure, done.

olimorris:
Sorry I held off merging this owing to `v17.18.0`. Will be happy to merge this. I think it just needs pointing to "adapters/http/openai.lua"

chrisgrieser:
> I think it just needs pointing to "adapters/http/openai.lua"

sure, done

PR #543, opened 2024-12-16
Use vim.treesitter instead of calls to nvim-treesitter.

dsully:
## Description
Use vim.treesitter instead of calls to nvim-treesitter.

Pull in two functions from nvim-treesitter.ts_utils that no longer exist on `main`

## Related Issue(s)
  - Fixes #542 

## Checklist

- [X] I've read the [CONTRIBUTING](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md) guidelines and have adhered to them in this PR
- [X] I've updated the README and ran the `make docs` command

PR #2021, opened 2025-08-26
chore(docs): remove outdated dependency nvim-treesitter

antonk52:
## Description

nvim-treesitter is not used directly since [10.8](https://github.com/olimorris/codecompanion.nvim/blob/3c9224e3a0084d6c251b9f81d5cb3c6ac12878e6/CHANGELOG.md?plain=1#L1757C197-L1757C292). For this reason there should be no requirement for a user to need the nvim-treesitter plugin as a dependency.

it is only used in minimal repro lua files for installing parsers where I left it unchanged. All other references to the plugin were removed.

I've tested the changes and the chat sidebar as well as `:checkhealth` work the same as current main, except no nvim-treesitter dependency in checkhealth

## Related Issue(s)

issue does not exist

## Screenshots

No visual changes to the plugin

## Checklist

~~- [ ] I've updated `CodeCompanion.has` in the [init.lua](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/init.lua#L239) file for my new feature~~ NA
~~- [ ] I've added [test](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md#testing) coverage for this fix/feature~~ NA
~~- [ ] I've run `make all` to ensure docs are generated, tests pass and my formatting is applied~~ I tried but there were multiple unrelated changes in the doc/codecompaion.txt so I left it unpushed for now for an easier review


olimorris:
Huh?! Are we looking at the same plugin?!

antonk52:
Sorry, can you please expand? I am not sure I understand your comment

Since this plugin no longer requires nvim treesitter i.e no `require('nvim-treesitter')` in the code of the plugin, there should be no need to specify "nvim-treesitter" as a dependency. It looks like for the past 6 months codecompanion has been using the neovim's built treesitter methods and not the ones from nvim-treesitter package, ever since this #543 was merged.

olimorris:
If you remove `nvim-treesitter` from the minimal.lua file, will CodeCompanion work?

antonk52:
It will! 

Using `nvim -u minimal.lua`

**Inline changes** by running `:CodeCompantion show me a for loop in lua`

Screenshot 2025-08-26 at 23 00 21

**CodeCompantionChat** not highlighted by default because treesitter is not started.

Screenshot 2025-08-26 at 23 00 43

Since vim has a builtin markdown parser, it can be used to highlight it. All that is needed is to run `vim.treesitter.start()` when opening the buffer

Screenshot 2025-08-26 at 23 00 55

olimorris:
Technically, we require a [yaml](https://codecompanion.olimorris.dev/usage/chat-buffer/#settings) and a markdown_inline parser (the latter for images). However, we could always make that explicitly clear in the documentation. I just feel I owe it to any future users to make sure I've handled situations where there isn't a parser, adequately. I'd hate to brick someones chat because I didn't put a simple check in.

Appreciate you digging into this though. I am hugely in favour of removing dependencies across CodeCompanion (not that we have too many now).

antonk52:
Got it, tbh I used it without a yaml parser on and off for the last few months and I didn't notice breakages. Maybe calling it out explicitly in the documentation would be clearer.

> I just feel I owe it to any future users to make sure I've handled situations where there isn't a parser, adequately. I'd hate to brick someones chat because I didn't put a simple check in.

Sure thing, I definitely did not intend to break anything as I've been able to use it without a yaml parser and figured it was unnecessary. 

In this case it also seems like what is actually required is having yaml and markdown_inline parsers and not the nvim-treesitter plugin. As the parsers can also be installed in other ways such as luarocks which is a lot less popular but still an option.

olimorris:
Appreciated and very solid suggestions.

I just checked chat/init.lua and I do actually check if a yaml parser is installed and I've just pushed a refactor to do the same for the symbol slash commands.

Thanks for this. Happy to merge this now we've both fully tested.

antonk52:
Rebased with no conflicts now, should be good to merge now.

I've also added a line to attach a treesitter to the chat buffer, that required to update some screenshot tests, from the changes you can see that those now properly that the buffer is highlighted using treesitter.

olimorris:
Superb thank you. I'll give this a try later on this week but looking good to me.

olimorris:
...and merged! Great PR. Appreciate you taking the time on this and investigating everything.

PR #2031, opened 2025-08-27
fix(diff): display the correct `sign_text` in floating diffs

bassamsdata:
## Description

This is a small fix to display the correct `sign_text` in floating diffs. The `open_window` function with option `style = "minimal"` prevents signs from showing up properly, so it’s better to leave the `style = "minimal"` option to the module rather than enforcing it inside the general `create_float` function.

## Related Issue(s)

Sign texts don’t appear near deleted lines (virtual text) in inline diffs.

Thank you

## Screenshots

Current behavior: no `sign_text` indicator appears next to the virtual lines highlighted in red.
Screenshot 2025-08-27 at 5 46 37 PM

## Checklist

### DIFF
--- a/lua/codecompanion/utils/ui.lua
+++ b/lua/codecompanion/utils/ui.lua
@@ -33,7 +33,7 @@ M.create_float = function(lines, opts)
     border = (vim.fn.exists("+winborder") == 0 or vim.o.winborder == "") and "single" or nil,
     width = width,
     height = height,
-    style = "minimal",
+    style = opts.style,
     row = row,
     col = col,
     title = opts.title or "Options",
--- a/lua/codecompanion/utils/ui.lua
+++ b/lua/codecompanion/utils/ui.lua
@@ -33,7 +33,7 @@ M.create_float = function(lines, opts)
     border = (vim.fn.exists("+winborder") == 0 or vim.o.winborder == "") and "single" or nil,
     width = width,
     height = height,
-    style = opts.style,
+    style = opts.style or "minimal",
     row = row,
     col = col,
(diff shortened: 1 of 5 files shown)


lua/codecompanion/utils/ui.lua
```diff
@@ -33,7 +33,7 @@ M.create_float = function(lines, opts)
     border = (vim.fn.exists("+winborder") == 0 or vim.o.winborder == "") and "single" or nil,
     width = width,
     height = height,
-    style = "minimal",
+    style = opts.style,
```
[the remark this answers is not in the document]
bassamsdata: Actually, we’re avoiding that issue because the `style` option always takes a single value:`"minimal"`. Falling back to it would cause the very problem this entire PR aims to fix.

olimorris:
That's a fantastic catch! Thank you

lua/codecompanion/utils/ui.lua
```diff
@@ -33,7 +33,7 @@ M.create_float = function(lines, opts)
     border = (vim.fn.exists("+winborder") == 0 or vim.o.winborder == "") and "single" or nil,
     width = width,
     height = height,
-    style = opts.style,
+    style = opts.style or "minimal",
```
bassamsdata, line 36: @olimorris, Copilot is mistaken here. This invalidated the PR and will always add a `style` set to `"minimal"`, even if we don’t specify it in other modules. Falling back to `nil` is the good approach here. 😔

olimorris: Reverting...

PR #2054, opened 2025-08-30
ACP v1.1

olimorris:
## Description

Implement some additional items from the Agent Client Protocol spec.

- [x] Agents can request specific [line numbers](https://agentclientprotocol.com/protocol/schema#fs%2Fread-text-file) when making an `fs/read_text_file` request
- [x] Respect [agent capabilities](https://agentclientprotocol.com/protocol/schema#agentcapabilities) when sending messages
- [x] Split out helper methods into a _helpers.lua_ file
- [x] Files are sent over as [embedded resources](https://agentclientprotocol.com/protocol/content#embedded-resource) rather than text
- [x] Add test coverage for _fs.lua_
- [x] Add lua types

## Checklist


bassamsdata:
This is amazing, thanks for all the hard work!  

One thing I noticed: it seems like `gemini_cli` can’t perform any edits after this PR (at least from what I’ve seen). It always reports the file as empty. I hate to point out an issue without a solution or a clear idea of where the problem lies since I haven’t fully read the whole PR, but I did some debugging to see what’s going on.  

From the debug info, it looks like it’s requesting the correct file, but the content returned is empty. I’m not sure if this is due to an incorrect line or limit in `/acp/fs.lua` with `"line":null,"limit":null` parameter, but maybe we also need to check if the content isn’t `vim.NIL`, though not really sure.

Please feel free to ignore this comment if I’m off base, I haven’t done a full read of the PR and am just going off the debug output. :)  

```json
Received stdout:
{"jsonrpc":"2.0","id":0,"method":"fs/read_text_file","params":{"path":"/Users/bassam/repos/codecompanion.nvim/lua/codecompanion/strategies/chat/slash_commands/quickfix.lua","sessionId":"deb8cc52-41a0-452d-ae72-35fd8bdf0f85","line":null,"limit":null}}
```
```
[acp::_write_to_process] Sending data: {"id":0,"jsonrpc":"2.0","result":{"content":""}}
```


**Edit update:** it looks like if I did it as below makes it work as before
```lua
local has_line = opts.line ~= nil and opts.line ~= vim.NIL
local has_limit = opts.limit ~= nil and opts.limit ~= vim.NIL
```

olimorris:
Thanks for sharing this. It was indeed `vim.NIL` that was catching me out. Fixed that in the latest commit

PR #2090, opened 2025-09-05
chore(copilot): use GH_HOST for Copilot endpoints and token lookup

rmuir:
Update Copilot HTTP requests and token lookup to respect the `GH_HOST` env var, allowing use with GitHub Enterprise instances.

Some networks block the `github.com` endpoints, so they won't work at all. Provide a way for the user to hit different endpoint other than `github.com` (e.g. mycorp.ghe.com) without patching the source code.

The `GH_HOST` environment variable is the same one used to override this for `gh` cli: https://cli.github.com/manual/gh_help_environment

Note: maybe this isn't the best solution, I'm not a lua guy... but I tested it in such a restrictive environment and it works.


rmuir:
Maybe a cleaner solution would be to parse out of the oauth rather than looking for `github.com`, I don't know. `~/.config/github-copilot/apps.json` entry looks like this, the `mycorp.ghe.com` is in there. Could be a better solution than environment variables and "just work" in more cases:

```json
{"mycorp.ghe.com:Xy1.00deadbeef0":{"user":"my-user","oauth_token":"<nope>","githubAppId":"Xy1.00deadbeef0"}}
```

AmadeusK525:
@olimorris Any chance this could be merged? My company just updated copilot support to use Enterprise, so I can't use CodeCompanion anymore without this PR

AmadeusK525:
Well, I don't think this patch is enough... testing it locally I can see that the way the plugin handles URLs is too coupled, as the URL is always fully formed with host + path. The "real" fix would be to separate this

olimorris:
All - apologies for not getting to this sooner. @rmuir presume this is good to be merged?

Aware this isn't enough for @AmadeusK525 but potentially still useful for other users.

rmuir:
@olimorris I think it is no good for current `main`. I tried this PR after the merge to latest in my restrictive environment and PR seems to have problems. When sending an actual chat, I see error implying it is still trying to connect to `github.com` and hitting firewall:

```
[chat::_submit_http] Error: { "curl: (35) Recv failure: Connection reset by peer" }
```

AmadeusK525:
Hey, I re-opened this in #2705, and it does work (I've been using it for a while)

PR #2103, opened 2025-09-10
fix(tools): set strict=true for all chat tools

Groveer:
## Description
Add strict=true to all function tool schemas to ensure consistent strictness across the chat tool catalog.

When I use a third-party model to call tools, I get an error: `All function tools should have the same strictness.` So I think all tools should set strict to true, tested this is correct

## Screenshots
图片

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/tools/catalog/create_file.lua
+++ b/lua/codecompanion/strategies/chat/tools/catalog/create_file.lua
@@ -162,6 +162,7 @@ return {
           "content",
         },
       },
+      strict = true,
     },
   },
   handlers = {
--- a/lua/codecompanion/strategies/chat/tools/catalog/fetch_webpage.lua
+++ b/lua/codecompanion/strategies/chat/tools/catalog/fetch_webpage.lua
@@ -79,6 +79,7 @@ return {
         },
         required = { "url" },
       },
+      strict = true,
     },
   },
   output = {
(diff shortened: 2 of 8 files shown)


Davidyz:
May I ask what provider those errors come from? Also, if this is consistently happening across all tools, maybe we should emphasise this [in the docs](https://codecompanion.olimorris.dev/extending/tools.html#schema) as well.

Groveer:
> May I ask what provider those errors come from? Also, if this is consistently happening across all tools, maybe we should emphasise this [in the docs](https://codecompanion.olimorris.dev/extending/tools.html#schema) as well.

This is our company's model, a third-party Kimi-K2 model, not an official model.

olimorris:
I've never come across, nor have I heard of this being an issue before..

I can't accept this PR if we're not going to test this for every single adapter that supports function calling.

Groveer:
> I've never come across, nor have I heard of this being an issue before..
> 
> I can't accept this PR if we're not going to test this for every single adapter that supports function calling.如果我们不对每个支持函数调用的适配器进行测试，我就不能接受这个 PR。

May I ask why some tools have set strict=true, while others have not?

olimorris:
No idea. We have a lot of contributors to the plugin

Davidyz:
AFAIK `strict=true` decides whether the generated schema will be validated by a json schema validator. This is going to be VERY provider-specific, and I've ran into issues where some providers simply don't support validations of certain schema rules (like `default` or some numerical comparisons).

Groveer:
After my recent testing, I found that some models cannot recognize the `strict` field. I will close this PR, but I suggest making the `strict` field configurable for users.

PR #2104, opened 2025-09-10
Cache Ollama models info

Davidyz:
## Description

In the previous Ollama adapter refactor (#1829), the Ollama adapter sends excessive requests to the server because it doesn't cache any of the results. This blocks the UI whenever the schema is resolved.

This PR implements a basic caching mechanism that stores the model list and model capabilities (thinking and vision), so that we don't have to suffer from the delay when the schema is resolved. The first time the schema is resolved will still be blocking, but the subsequent calls to `get_models` or `check_thinking_capability` will use the cached results and will no longer block the UI.

## Related Issue(s)

#1829 

## Checklist

### DIFF
--- a/lua/codecompanion/adapters/http/ollama/get_models.lua
+++ b/lua/codecompanion/adapters/http/ollama/get_models.lua
@@ -20,8 +20,10 @@ M = {}
 ---@type table<string, table<string, { nice_name: string?, opts: {can_reason: boolean, has_vision: boolean} }>>
 local _cached_models = {}
 
+---@alias OllamaGetModelsOpts {last?: boolean, async?: boolean}
+
 ---@param url string
----@param opts? {last: boolean}
+---@param opts? OllamaGetModelsOpts
 local function get_cached_models(url, opts)
   assert(_cached_models[url] ~= nil, "Model info is not available in the cache.")
   local models = _cached_models[url]
@@ -36,7 +38,8 @@ end
 ---Aborts if there's another fetch job running.
 ---Returns the number of models if the fetches are fired.
 ---@param adapter CodeCompanion.HTTPAdapter Ollama adapter with env var replaced.
-local function fetch_async(adapter)
+---@param opts OllamaGetModelsOpts
+local function fetch_async(adapter, opts)
   assert(adapter ~= nil)
   if running then
     return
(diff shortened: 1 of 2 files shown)


olimorris:
Sorry to be picky, can we conform to the same variable and function naming that we have for the Copilot adapter? The plugin is starting to show that it's been contributed to by a million people and I haven't done a good job of unifying this.

Davidyz:
@olimorris I've done the renaming, but I took a different caching approach that allows caching model lists for multiple ollama endpoints, so the structure won't be exactly the same. While we're at it, maybe I should make a full refactor to the model-fetching and add expiry, etc., similar to [copilot](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/adapters/http/copilot/get_models.lua)?

olimorris:
Feel free to 👍🏼

Davidyz:
Having thought more about it, it's actually not favourable to just throw away the cache after expiry, because doing so means we have to wait for the server to respond when resolving the schema (`think` will be resolved every time the user sends a message, so it'll still occasionally block the UI). I think it might be better to just schedule a background job that refreshes the model information periodically, so that the user will only have to suffer from the blocking once.

Alternatively, we could also refresh on every cache access. This'll reduce the number of background jobs, but if the server state changes, we're more likely to run into outdated information.

Davidyz:
@olimorris just finished the refactoring. It now updates the model information on schema resolution. It should only block on the adapter initialisation, and should be smooth in subsequent runs. Do you think we'd need a timer that periodically updates the model info?

olimorris:
Sorry not had chance to take a look at this until now. I will try and merge later today and update #2181. I'm not sure this PR resolves the issue in #2180. I think I need to add an option when changing adapters that will reset the cache in adapters like Ollama and Copilot to allow for a model refresh.

Davidyz:
I tried setting up a second ollama server, and with this PR the model lists get updated correctly. I'm open to having a switch that disables the caching, but this may cause frozen UI quite frequently because the `think` parameter is bounded by a `condition` function that pulls the model capabilities for every request.

olimorris:
Thanks for testing. I'll test this out when you're ready

Davidyz:
> I'll test this out when you're ready

This is ready in the sense that it already caches the model info for multiple adapters correctly, and refreshes the cache on each cache access. I can incorporate `cache_adapter` from #2181 into this one, if that's what you meant.

Update: the `cache_adapter` option is now part of this PR too.

olimorris:
No didn't intend for you to add that at all - but thank you.

PR #2191, opened 2025-10-01
feat(acp): add session mode command and auto-connect chat buffers

cellfusion:
## Description

  - Add :CodeCompanionChat SetSessionMode so ACP chats can switch session modes.
  - Automatically connect ACP adapters when the chat opens (session/new) to retrieve available modes, showing a
  “Connecting to ACP...” status until the session is ready.

  ## Related Discussion(s)
  
  https://github.com/olimorris/codecompanion.nvim/discussions/2190

  ## Screenshots
  
  
CleanShot 2025-10-02 at 03 20 26@2x

CleanShot 2025-10-02 at 03 20 45@2x

  ## Checklist

  - [ ] I've read the contributing (https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md)
  guidelines and have adhered to them in this PR
  - [ ] I've added test (https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md#testing) coverage for this fix/feature 
  - [ ] I've run make all to ensure docs are generated, tests pass and my formatting is applied
  - [x] (optional) I've updated CodeCompanion.has in the init.lua 
  - [ ] (optional) I've updated the README and/or relevant docs pages

### DIFF
--- a/lua/codecompanion/commands.lua
+++ b/lua/codecompanion/commands.lua
@@ -46,6 +46,7 @@ local chat_subcommands = vim.deepcopy(adapters)
 table.insert(chat_subcommands, "Toggle")
 table.insert(chat_subcommands, "Add")
 table.insert(chat_subcommands, "RefreshCache")
+table.insert(chat_subcommands, "SetSessionMode")
 
 ---@type CodeCompanion.Command[]
 return {
--- a/lua/codecompanion/init.lua
+++ b/lua/codecompanion/init.lua
@@ -140,6 +140,8 @@ CodeCompanion.chat = function(args)
         return CodeCompanion.toggle(args)
       elseif prompt == "refreshcache" then
         return CodeCompanion.refresh_cache()
+      elseif prompt == "setsessionmode" then
+        return CodeCompanion.set_session_mode()
       else
         table.insert(messages, {
           role = config.constants.USER_ROLE,
@@ -254,6 +256,17 @@ CodeCompanion.restore = function(bufnr)
     return
   end
   chat.ui:open()
+end
+
+---Set Session Mode
+---@return nil
+CodeCompanion.set_session_mode = function()
+  local chat = CodeCompanion.last_chat()
+  if not chat then
+    return
+  end
+
+  chat.set_session_mode()
 end
 
 ---Return a chat buffer
--- a/lua/codecompanion/strategies/chat/acp/set_session_mode.lua
+++ b/lua/codecompanion/strategies/chat/acp/set_session_mode.lua
@@ -1,0 +1,77 @@
+local util = require("codecompanion.utils")
+
+local M = {}
+
+local CONSTANTS = {
+  PROMPT_TITLE = "Set Session Mode",
+  PROMPT_KIND = "session mode",
+}
+
+---Build a human readable label for a mode entry
+---@param mode table
+---@param current_mode_id string|nil
+---@return string
+local function mode_label(mode, current_mode_id)
+  local name = mode and (mode.name or mode.displayName or mode.id)
+  if type(name) ~= "string" or name == "" then
+    name = tostring(mode and mode.id or "Unknown")
+  end
+
+  if current_mode_id and mode and mode.id == current_mode_id then
+    name = string.format("%s (current)", name)
+  end
+
+  return name
+end
+
+---Build prompt and mappings for vim.fn.confirm
+---@param modes table[]
+---@param current_mode_id string|nil
+---@return string, string[], table<number, string>, table<string, number>
+local function build_choices(modes, current_mode_id)
+  local prompt = string.format("%s: %s?", util.capitalize(CONSTANTS.PROMPT_KIND), CONSTANTS.PROMPT_TITLE)
+  local choices, index_to_mode, mode_to_index = {}, {}, {}
+
+  for index, mode in ipairs(modes or {}) do
+    local choice_label = mode_label(mode, current_mode_id)
+    table.insert(choices, string.format("&%d %s", index, choice_label))
+    if mode and mode.id then
+      index_to_mode[index] = mode.id
+      mode_to_index[mode.id] = index
+    end
+  end
+
+  return prompt, choices, index_to_mode, mode_to_index
+end
+
+---Display the available modes and return the selected mode id (if any)
+---@param chat CodeCompanion.Chat|nil
+---@param opts { available_modes: table[], current_mode_id?: string|nil }
+---@return string|nil
+function M.show(chat, opts)
+  opts = opts or {}
+  local _ = chat
+  local modes = opts.available_modes or {}
+  if vim.tbl_isempty(modes) then
+    return nil
+  end
+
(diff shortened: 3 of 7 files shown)


lua/codecompanion/init.lua
```diff
@@ -256,6 +258,17 @@ CodeCompanion.restore = function(bufnr)
   chat.ui:open()
 end
 
+---Set Session Mode
+---@return nil
+CodeCompanion.set_session_mode = function()
```
olimorris: I think we should be explicit and call this `set_acp_session_mode`. You're correct to name it this, but if a user doesn't use ACP adapters, this might read as a feature for the http ones

lua/codecompanion/commands.lua
```diff
@@ -46,6 +46,7 @@ local chat_subcommands = vim.deepcopy(adapters)
 table.insert(chat_subcommands, "Toggle")
 table.insert(chat_subcommands, "Add")
 table.insert(chat_subcommands, "RefreshCache")
+table.insert(chat_subcommands, "SetSessionMode")
```
olimorris, line 49: I have some reservations about the command. I typically start an ACP session with `:CodeCompanionChat claude_code`. I don't believe `:CodeCompanionChat claude_code SetSessionMode` will work. Now that's through no fault of your own, that's just how I built the command. But we need to account for that.

lua/codecompanion/strategies/chat/init.lua
```diff
@@ -599,6 +628,163 @@ function Chat:apply_settings(settings)
...
+  if self.ui and should_show_intro(self) then
+    self:_set_acp_status_message(ACP_MESSAGES.CONNECTING)
+  end
+
+  self._acp_connecting = true
+  local connection = self.acp_connection
+  if not connection then
+    local client = get_client(self.adapter)
+    if not client or type(client.new) ~= "function" then
+      util.notify("ACP client not available", vim.log.levels.ERROR)
```
olimorris: This will lead to a double notification as `log:error` also creates a notification

lua/codecompanion/strategies/chat/init.lua
```diff
@@ -435,7 +451,8 @@ function Chat.new(args)
     { bufnr = self.bufnr, id = self.id, model = self.adapter.schema and self.adapter.schema.model.default }
   )
 
-  if self.adapter.type == "http" then
+  local needs_acp_connect = self.adapter.type == "acp"
```
olimorris: Rename this to be `need_acp_connection` so it's a little more readable as `if not need_acp_connection then`

lua/codecompanion/strategies/chat/init.lua
```diff
@@ -599,6 +628,163 @@ function Chat:apply_settings(settings)
...
+  if self.status == CONSTANTS.STATUS_CONNECTING then
+    self.status = ""
+  end
+  self:update_metadata()
+  return true
+end
+
+---Schedule establishing the ACP connection so the UI can render first.
+---@param show_intro boolean
+function Chat:_schedule_acp_connection(show_intro)
```
olimorris, line 715: We are technically scheduling the session but it would be clearer to call it `create_acp_connection` as it's execution will be deemed `needs_acp_connection`.

lua/codecompanion/strategies/chat/init.lua
```diff
@@ -599,6 +628,163 @@ function Chat:apply_settings(settings)
...
+      self.status = ""
+    end
+    return true
+  end
+
+  if self._acp_connecting then
+    if self.ui and should_show_intro(self) then
+      self:_set_acp_status_message(ACP_MESSAGES.CONNECTING)
+    end
+    vim.wait(20000, function()
```
olimorris: We can use `config.strategies.chat.opts.wait_timeout` in place of hard coding this.

lua/codecompanion/strategies/chat/acp/set_session_mode.lua
```diff
@@ -0,0 +1,77 @@
...
+  if #choices == 0 then
+    return nil
+  end
+
+  local default_choice = 1
+  if opts.current_mode_id and mode_to_index[opts.current_mode_id] then
+    default_choice = mode_to_index[opts.current_mode_id]
+  end
+
+  local picked = vim.fn.confirm(prompt, table.concat(choices, "\n"), default_choice, "Question")
```
olimorris: Where possible, I prefer to use `vim.ui.select` as it looks better and doesn't block the editor. Does it need to be a `vim.fn.confirm`

cellfusion: Indeed, `vim.ui.select` seems better in this case.
I’ll try modifying it.

lua/codecompanion/commands.lua
cellfusion: Indeed, it’s a command that can’t be used right at startup…
So I was thinking of creating `:CodeCompanionACP` for future ACP-related operations, and then having subcommands like `:CodeCompanionACP SetSessionMode`. But adding it just for this one command feels a bit awkward…
I’ll give it some more thought.

lua/codecompanion/init.lua
cellfusion: Done

lua/codecompanion/strategies/chat/init.lua
cellfusion: Done

lua/codecompanion/strategies/chat/init.lua
cellfusion: Done

lua/codecompanion/strategies/chat/init.lua
cellfusion: Done

lua/codecompanion/strategies/chat/acp/set_session_mode.lua
cellfusion: Done 😄

lua/codecompanion/commands.lua
olimorris: The other option is that we keep the first word as the adapter (keeps things backwards compatible) and have a second word that looks for additional commands

lua/codecompanion/strategies/chat/init.lua
cellfusion: Done

olimorris:
Hey @cellfusion. I've added this in #2331 and I've chosen to add this via a slash command - `/mode`. Whilst it would be great to add this via the cmd line, I try and limit how many commands CodeCompanion has. It would be great to parametize it via something like `:CodeCompanionChat adapter=claude_code mode=ask` but it's a little too complicated right now. I value just adding this feature into the plugin.

Thanks for all your hard work. Really appreciate your contribution.

cellfusion:
@olimorris 

Hey! Thanks so much!
I was really struggling with how to solve this and time was just slipping away without finding a good solution.
I'm so relieved that at least the /mode command has been incorporated.

PR #2249, opened 2025-10-13
Tools: Easier to YOLO and send rejection message

olimorris:
## Description

1. Make it easier to use yolo mode:

https://github.com/user-attachments/assets/235afb1c-6b00-4862-b60d-773cecc30e55

2. When you reject a tool call, specify a rejection reason:

https://github.com/user-attachments/assets/c2f8dd2a-9da6-4c4d-8c99-2e8f5a6fb07a

## Related Issue(s)

#2217 

## Checklist

### DIFF
--- a/tests/strategies/chat/helpers/test_wait.lua
+++ b/tests/strategies/chat/helpers/test_wait.lua
@@ -54,7 +54,6 @@ end
 
 T["for_decision()"]["waits for matching event with correct id"] = function()
   child.lua([[
-    vim.g.codecompanion_yolo_mode = false
     wait.for_decision("test_123", {"CodeCompanionAccept", "CodeCompanionReject"}, _G.test_callback)
   ]])
 
@@ -74,7 +73,6 @@ end
 
 T["for_decision()"]["waits for reject event with correct id"] = function()
   child.lua([[
-    vim.g.codecompanion_yolo_mode = false
     wait.for_decision("test_456", {"CodeCompanionAccept", "CodeCompanionReject"}, _G.test_callback)
   ]])
 
@@ -94,7 +92,6 @@ end
 
 T["for_decision()"]["ignores events with wrong id"] = function()
   child.lua([[
-    vim.g.codecompanion_yolo_mode = false
     wait.for_decision("test_789", {"CodeCompanionAccept", "CodeCompanionReject"}, _G.test_callback)
   ]])
 
@@ -125,7 +122,6 @@ end
 
 T["for_decision()"]["handles notification option"] = function()
   child.lua([[
-    vim.g.codecompanion_yolo_mode = false
     wait.for_decision("test_notify", {"Accept", "Reject"}, _G.test_callback, {
       notify = "Please make a decision..."
     })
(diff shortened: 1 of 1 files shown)


shyam-king:
This feature does not work for edit files tool? (`insert_edit_into_file`)

olimorris:
Yep. It's global

shyam-king:
No, I mean it is not working when I do `gdr`. Should I raise a bug report?

lua/codecompanion/strategies/chat/tools/catalog/insert_edit_into_file.lua
```diff
@@ -183,7 +190,7 @@ local function edit_file(action, chat, output_handler, opts)
       -- NOTE: This is required to ensure folding works for chat buffers that aren't visible
       codecompanion.restore(chat.bufnr)
       return output_handler(response)
-    end, wait_opts)
+    end, wait_opts(chat))
```
shyam-king, line 193: We are not doing anything on rejected here ? Shouldn't we call `helpers.rejected` or something (I do not have context of codebase at all so I apologise if I am completely off , basing my reason on the rest of PR patch)

olimorris: Rejection is handled via the orchestrator, not at a tool level.

olimorris:
Pretty sure yolo mode is gdy. I'm confused by what you're trying to do. Reject or yolo mode? Might be worth reading the docs again.

shyam-king:
> Pretty sure yolo mode is gdy. I'm confused by what you're trying to do. Reject or yolo mode? Might be worth reading the docs again.

I actually landed here from https://github.com/olimorris/codecompanion.nvim/discussions/2217

I didn't realise this (PR) was YOLO only. I was expecting to provide feedback whenever I do `gdr` (reject change).


EDIT: In fact, now I'm more confused about the use case of the PR. If you are going YOLO, how will you provide rejection reason? 

EDIT 2: I have realised how to better articulate my issue (and looks like this is a special case for tool `insert_edit_into_file`). In non-YOLO mode (where every edit goes through human: `gda`/`gdr`): there is no way to provide feedback to LLM on why we are rejecting the edit. This is different from rejecting the tool call execution because the user confirmation of the edit happens **during tool call execution** at: https://github.com/olimorris/codecompanion.nvim/blob/9f894a08ed706f3ec74362023f1f0da10729f29b/lua/codecompanion/strategies/chat/tools/catalog/insert_edit_into_file.lua#L170

Current behaviour:
1. LLM attempts to execute `insert_edit_into_file`
2. No user confirmation for tool execution (because of `requires_approval` being `false`)
3. Diff is rendered
4. User rejects the edit (`gdr`)
5. LLM gets "User rejected the edits" as tool result

Expected behaviour:
1. LLM attempts to execute `insert_edit_into_file`
2. No user confirmation for tool execution (because of `requires_approval` being `false`)
3. Diff is rendered
4. User rejects the edit (`gdr`)
5. We prompt the user for reason on rejection
6. LLM gets the reason as result

olimorris:
> If you are going YOLO, how will you provide rejection reason?

You're mistaken by what YOLO means. It removes the need for approvals and rejections and so automatically applies an LLM's suggestions.

> ...and looks like this is a special case for tool `insert_edit_into_file`

Yes, correct. This was never added into that tool. I don't know why. `v18.0.0` will include this 👍🏼 .

PR #2251, opened 2025-10-15
Fix: Allow `after/ftplugin/codecompanion.lua` settings to override plugin window options

bassamsdata:
## Description

Users couldn't override the plugin's default window options (like `wrap`, `number`, etc.) using their `after/ftplugin/codecompanion.lua` file. The plugin's defaults would always take precedence.

### Root Cause

The timing of when filetype was set vs. when window options were applied:

### CodeCompanion Chat Buffer (Before Fix)
1. Buffer created
2. **Filetype set immediately** (buffer not in window yet)
3. `FileType` autocmd fires → `after/ftplugin` runs
4. User tries `vim.wo.wrap = false` but **fails silently** (no window exists)
5. Much later, window created and buffer shown (in `UI:open()`)
6. Plugin sets its default window options
7. User's settings are lost

### The Fix

Set filetype **after** showing buffer in window and applying plugin defaults:

```lua
-- In UI:open() for all layouts (float, vertical, horizontal, buffer)
local function finalize_window_setup(winnr, bufnr, opts)
  ui.set_win_options(winnr, opts)  -- Plugin defaults first
  api.nvim_set_option_value("filetype", "codecompanion", { buf = bufnr })  -- Triggers ftplugin last
end
```

Now the sequence is:
1. Window created and buffer shown
2. Plugin sets its default window options
3. Filetype set → `FileType` fires → `after/ftplugin` runs
4. User's `vim.wo.wrap = false` works (window exists)
5. **User settings override plugin defaults**

### Additional Fix

For syntax highlighting, there’s no longer a need to wait for the `BufEnter` event because highlighting is applied immediately in the `FileType` autocmd. I can keep the `BufEnter` autocmd as a safety net if you prefer?

## Related Issue(s)

https://github.com/olimorris/codecompanion.nvim/discussions/1607#discussioncomment-14672615
#2164
#526

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/ui/init.lua
+++ b/lua/codecompanion/strategies/chat/ui/init.lua
@@ -21,8 +21,8 @@ local CONSTANTS = {
 }
 
 ---Finalize window setup by applying the options and setting the filetype. This
----This ensures that any codecompanion specific ftplugin can run after the
----window options are set, allowing the plugin defaults to be overriden
+---This ensures that any codecompanion specific ftplugins can run after the
+---window options are set, allowing the plugin defaults to be overridden
 ---@param winnr number
 ---@param bufnr number
 ---@param opts table Window options to apply


olimorris:
Wow. This is amazing! There's going to be some very happy users 👏🏼

PR #2260, opened 2025-10-17
Add support for ACP command completion in chat buffers

iamthebot:
## Description

  Adds completion support for ACP commands in chat buffers. ACP adapters like claude-code expose commands (e.g., `/compact`,
  `/context`) via the `available_commands_update` notification, and this PR makes them available through all 4 completion
  providers (blink, CMP, COC, default omnifunc).

  Commands are triggered with backslash (`\`) by default to avoid conflicting with CodeCompanion's built-in slash commands. When
   sending messages, the plugin automatically transforms `\command` to `/command` for the agent.

  **How it works:**
  - Commands are discovered dynamically from the ACP agent after session initialization (typically 1-5 seconds)
  - Storage uses a two-level mapping: `session_id → commands` and `bufnr → session_id`
  - Cleanup happens on `ChatClosed` rather than `BufDelete` (which fires spuriously during completion)
  - All providers mark commands as incomplete/dynamic so new commands appear immediately after discovery

  **Benefits for users:**
  - Access agent-specific commands directly from the chat buffer with completion
  - Configurable trigger character via `strategies.chat.acp_commands.opts.trigger`
  - Works seamlessly across all completion providers
  - Automatic command transformation means no manual editing needed

  ## Related Issue(s)
See the discussion [here](https://github.com/olimorris/codecompanion.nvim/discussions/2258)

  ## Screenshots

Screenshot 2025-10-16 at 6 31 27 PM

  ## Checklist

  - [x] I've read the [contributing](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md) guidelines and
  have adhered to them in this PR
  - [x] I've added [test](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md#testing) coverage for this
  fix/feature (19 new tests covering command storage, transformation, and integration)
  - [ ] _(optional)_ I've updated `CodeCompanion.has` in the
  [init.lua](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/init.lua#L239) file for my new feature
  - [x] _(optional)_ I've updated the README and/or relevant docs pages (added documentation to
  `doc/usage/chat-buffer/index.md`)

## AI Disclosure
I used claude code via ACP in codecompanion (of course! I had to test this very functionality) to write large portions of this. However, the completion code was very hairy to get correct so much of that is manually written + debugged.

### DIFF
--- a/doc/usage/chat-buffer/index.md
+++ b/doc/usage/chat-buffer/index.md
@@ -100,7 +100,7 @@ When using an ACP adapter (such as claude-code), you can also type `\` (backslas
 > [!NOTE]
 > It typically takes 1-5 seconds after opening a chat buffer for ACP commands to become available. The agent needs to initialize and scan for both built-in and custom commands. If you define a new custom command mid-session, the same delay applies before it appears in the completion list.
 
-The backslash trigger is used to avoid conflicts with CodeCompanion's built-in [Slash Commands](/usage/chat-buffer/slash-commands). When you send a message, `\command` is automatically transformed to `/command` for the agent. The trigger character can be customized via `strategies.chat.acp_commands.opts.trigger` in your config.
+The backslash trigger is used to avoid conflicts with CodeCompanion's built-in [Slash Commands](/usage/chat-buffer/slash-commands). When you send a message, `\command` is automatically transformed to `/command` for the agent. The trigger character can be customized via `strategies.chat.slash_commands.opts.acp.trigger` in your config.
 
 It's worth noting that not all commands available in ACP CLI tools are exposed via the SDK. Only a subset of built-in commands are supported, though this is constantly evolving as the underlying SDKs mature.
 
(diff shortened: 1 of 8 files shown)


olimorris:
Firstly, this is an immense PR. Super high quality. Thank you 🙏🏼.

Few things:
- If I opened up a chat buffer and did `ga` to change to an ACP adapter, the slash commands didn't appear. I've patched that
- I moved the config to `strategies.chat.slash_commands.opts.acp` as that's more inline with where I would expect to find it
- What was your thinking for having `\` in place of `/` for the agent slash commands? Was it purely to isolate the agent's from CodeCompanions? I'm torn because I like the idea of easily being able to see what agent specific commands there are and not lose them amongst the other slash commands. Alternatively, it feels less clean than being able to do `/` and we have to do some post-processing to replace `\` with `/`.

olimorris:
Awesome thanks for sharing your thought process. There's trade-offs either way and I think having the separation of ACP commands and CodeCompanion ones is reasonable.

PR #2281, opened 2025-10-22
New File System Change Tracking, Enhanced Super Diff

bassamsdata:
## Description

2 new/refactored changes in this PR:
 1. New FS Monitor to track all chnages made by LLM to the file system.
 2. Completely New UI for SuperDiff.

### Details:
This replaces the old snapshot-based `edit_tracker` with FSMonitor, an event-driven file system monitor. 

FSMonitor uses `vim.uv.new_fs_event` and async I/O to reliably detect created, modified, deleted, and renamed files in the `cwd`

Benefits: 
- More reliable, works with both http and ACP and any future enhancements.
- Detect any file creation, deletion, and renames.
- Fully non-blocking Async watcher.
- Easier to extend (e.g., restore checkpoints)
- Cleaner API.

> Note:
> This is working version but still work in progress.

Thank you

## Screenshots
New SuperDiff:

https://github.com/user-attachments/assets/a09acc61-405a-400c-a040-49a2d2509aa0

## Checklist


jinzhongjia:
Is there any progress about this pr?

bassamsdata:
@jinzhongjia, thank you for following up. This PR is nearly complete, but there are still several items that would need to be addressed after v18. Because of that, I’m considering opening a new PR for it.

@olimorris, I’d really appreciate your thoughts on this PR, do you think it has the potential to be merged?

At a high level, here’s a brief overview:

1. It replaces `edit_tracking` with a new file system monitor that is more reliable. The new approach detects not only file edits, but also file deletions and creations, and it does not depend on specific tools. As a result, it should benefit all tool integrations, whether HTTP, ACP, or others. I’d also say it’s more maintainable than the current `edit_tracker` implementation, which I originally wrote as well.
2. It introduces a new `super diff` (the name can certainly be changed if you prefer) that is more closely aligned with Neovim and easier for users to navigate and maintain.

   * It includes several features, with one of the most important being checkpoints. These allow users to review changes at specific points in time and revert back to a chosen checkpoint if needed.

There’s an initial video of the super diff in the OP. I’ve made a few UI improvements since recording it, but it still gives a good idea of what it looks like.

Apologies for mentioning you twice in the same day, and thank you in advance for your time. I wanted to get your perspective before moving forward.

olimorris:
> I’d really appreciate your thoughts on this PR, do you think it has the potential to be merged?

I've been reflecting on my philosophy for CodeCompanion moving forward into 2026 and beyond and I will share this at some point in the near future. Ultimately, I want CodeCompanion to solely focus on enabling developers to write great code with AI as efficiently as possible.

This PR sits in a grey area. Does it directly help devs write great code?! Perhaps. I could accept an argument either way.  However, from my perspective, having a file/edit monitoring system directly in the plugin feels too much. I also have doubts about how much it would be used by the user base.

As such, I think this is best suited to being an external plugin/extension and we can add additional events throughout CodeCompanion, allowing this to happen.

bassamsdata:
Totally fair. I agree this would work well as a standalone plugin/extension. Since it’s nearly finished, I’ll go ahead and make it a plugin.

> we can add additional events throughout CodeCompanion, allowing this to happen.

Thank you

PR #2287, opened 2025-10-22
Part of v18.0.0 - Improve user commands

olimorris:
## Description

Update the plugin's commands, allowing for:

- `:CodeCompanionChat adapter=claude_code are you working?`

## Checklist

### DIFF
--- a/doc/codecompanion.txt
+++ b/doc/codecompanion.txt
@@ -1,4 +1,4 @@
-*codecompanion.txt*         For NVIM v0.11        Last change: 2025 October 21
+*codecompanion.txt*         For NVIM v0.11        Last change: 2025 October 22
 
 ==============================================================================
 Table of Contents                            *codecompanion-table-of-contents*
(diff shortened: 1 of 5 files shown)


bassamsdata:
This is a great addition, it’s definitely going to make my Neovim workflow smoother :)

Would you be open to something like this?

```markdown
:CodeCompanionChat adapter=copilot/gpt-5-mini are you working?
```

Basically, if we detect a `/` separator, we could treat the part before it as the adapter and the part after as the model name.

If that sounds good to you, I’d be happy to take this on and send an update to the PR!

olimorris:
That's a great idea. I've built the commands pipeline such that we could do something like:

```
:CodeCompanionChat adapter=copilot model=gpt-5-mini are you working?
```

That would be my preferred way of doing it as it behaves like you're in the CLI. I'll need to do some resolving of the models in the background for completion purposes but I'll put this on the list.

bassamsdata:
> I've built the commands pipeline such that we could do something like:
> 
> ```
> :CodeCompanionChat adapter=copilot model=gpt-5-mini are you working?
> ```

Ah this is cleaner, I didn't know that. Thanks

PR #2300, opened 2025-10-23
Fix Race Condition During ACP Initialization w/ Command Updates

iamthebot:
## Description

When a new ACP session is created, the agent responds with both:
  1. A session response containing the session ID
  2. A session/update notification with available_commands_update

  These messages often arrive in the same I/O batch and are processed during the RPC wait loop (while waiting for the session creation response). The notification handler was checking self.session_id before it had been set, causing commands to be ignored with the log message:

```
[acp::handle_available_commands_update] No session ID; ignoring commands update
```

This resulted in slash command completions (as I added in #2260) being empty for the entire session. But only sometimes (since it's a race condition).

## Related Issue(s)

N/A

## Screenshots

N/A

## Checklist

(This is kind of hard to test given how it arises; however, existing tests do pass)


olimorris:
I had this happen once and I put it down to being claude-code-acp itself. But amazing catch.

PR #2306, opened 2025-10-25
Add Gemini reasoning support

Davidyz:
## Description

1. Properly add reasoning parameters in the requests.
2. handle reasoning output.
3. [x] use #2359

Reference: https://ai.google.dev/gemini-api/docs/openai#thinking

## Screenshots

image

## Checklist

### DIFF
--- a/lua/codecompanion/adapters/http/gemini.lua
+++ b/lua/codecompanion/adapters/http/gemini.lua
@@ -160,34 +160,6 @@ return {
       return data
     end,
 
-    ---@param self CodeCompanion.HTTPAdapter.Gemini
-    ---@param data table The reasoning output from the LLM
-    ---@return nil|{ content: string, _data: table }
-    form_reasoning = function(self, data)
-      if data == nil or not vim.iter(data):any(function(item)
-        return item ~= nil
-      end) then
-        return
-      end
-
-      local content = vim
-        .iter(data)
-        :map(function(item)
-          local val = item.content or item
-          if type(val) == "string" then
-            return val
-          end
-        end)
-        :filter(function(content)
-          return content ~= nil
-        end)
-        :join("")
-
-      return {
-        content = content,
-      }
-    end,
-
     tools = {
       format_tool_calls = function(self, tools)
         return openai.handlers.tools.format_tool_calls(self, tools)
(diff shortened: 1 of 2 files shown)


lua/codecompanion/adapters/http/gemini.lua
```diff
@@ -151,14 +218,9 @@ return {
...
-      default = "medium",
-      desc = "Constrains effort on reasoning for reasoning models. Reducing reasoning effort can result in faster responses and fewer tokens used on reasoning in a response.",
-      choices = {
-        "high",
-        "medium",
-        "low",
-        "none",
-      },
+      default = -1, -- dynamic reasoning
+      -- TODO: validate requires having `self` in the params.
```
Davidyz, line 273: Since different models have different [thinking budget](https://ai.google.dev/gemini-api/docs/thinking#set-budget), we might need to change `validate` so that it also accepts `self` in the parameters (something like `validate(self, n)` instead of the current `validate(n)`. @olimorris would that be alright?

Davidyz:
Btw, this may also be the culprit behind the Gemini adapter having occasional time out error: it defaults to use dynamic thinking, but the cc adapter didn't request for the reasoning summary and thought it timed out.

Maybe we should increase the default timeout for Gemini adapter or make it configurable?

Davidyz:
The `thinking_budget` schema currently doesn't validate because the valid value depends on the model, and the current schema implementation doesn't pass `self` to the `validate` function.

Davidyz:
A few things after the gemini-3-pro release:
1. Gemini-3-pro uses [`thinking_level` (`"low"` or `"high"`)](https://ai.google.dev/gemini-api/docs/thinking#thinking-levels) instead of `thinking_budget` (`integer`). I've added the relevant schema, but haven't tested that on a live API.
2. They seem to have added support for the standard [`reasoning_effort` field](https://ai.google.dev/gemini-api/docs/openai#thinking) from the OpenAI API. It'll be mapped to the appropriate `thinking_level` or `thinking_budget` parameter on the server. This doesn't matter for the Gemini 3 model because it's also a string enum, but for the Gemini 2.5 series, we lose the fine control over the number of tokens used for reasoning.

PR #2311, opened 2025-10-27
feat(ACP): support all plugin diff providers

bassamsdata:
## Description

Enable ACP adapter to work with all diff providers available in the plugin:
- inline diff (float and buffer layouts)
- Neovim builtin split mode
- mini.diff plugin

Previously only inline floating layout was supported. Aligns ACP adapter feature parity with HTTP adapter.

> Note: Diff buffers in ACP will include a suffix like `_diff_${diff_id}`, as shown in the screenshot below. This helps avoid errors that occur when two buffers share the same name while the real buffer is already loaded.

Thank you

## Related Issue(s)

https://github.com/olimorris/codecompanion.nvim/pull/2305#issuecomment-3449081768
#2136

## Screenshots

diff using Claude Code with `provider = "split"`

Screenshot 2025-10-28 at 10 58 40 PM

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/acp/request_permission.lua
+++ b/lua/codecompanion/strategies/chat/acp/request_permission.lua
@@ -224,15 +224,15 @@ local function on_user_response(request, diff, opts)
         vim.wo[opts.winnr].winbar = ""
       end)
 
-      if not opts.keep_win_open then
-        pcall(api.nvim_win_close, opts.winnr, true)
-      else
+      if opts.keep_win_open then
         -- Prevents the window from closing when we delete the temp buffer
         pcall(function()
           api.nvim_win_call(opts.winnr, function()
             vim.cmd("buffer #")
           end)
         end)
+      else
+        pcall(api.nvim_win_close, opts.winnr, true)
       end
     end
 
@@ -388,7 +388,7 @@ local function show_diff(chat, request)
   local provider_config = config.display.diff.provider_opts[provider] or {}
   local layout = provider_config.layout
   local is_floating = provider == "inline" and layout == "float"
-  local keep_win_open = provider == "inline" and layout == "buffer"
+  local keep_win_open = provider == "inline" and layout == "buffer" or provider == "split"
 
   local diff = diff_module.new({
     bufnr = bufnr,


amadeus:
Unfortunately it seems to get into quite a broken state:

CleanShot 2025-10-28 at 13 36 19

like it was showing a mix of the overlay diff view, but side by side. i couldn't immediately get the hot keys to work, and when I was able to finally accept them, it had closed my file buffer completely and I was left with only the code companion chat buffer

vs what it should look like (with sonnet for example):
CleanShot 2025-10-28 at 13 38 28

bassamsdata:
Thank you for testing and for the feedback! I’ve actually never used the Vim diff with CodeCompanion, so I’m not entirely sure how it’s supposed to look. When testing with `gemini_cli`, it appears like below screenshots and video which seems identical to GitHub Copilot to me, though I might be wrong. If you notice any issues in the screenshots or videos I shared, please let me know. If both look correct to you, then `claude_code` might be handling diffs differently (though I doubt it).

That said, I installed both Claude Code and `claude_code_acp`, but I’m running into an issue where CodeCompanion doesn’t detect it. I’ll sort that out later.

The only difference compared to the HTTP adapter is that this diff is created with a temp buffer, whereas the HTTP one isn’t. I also found an issue where there might be a mismatch because one buffer has `keymaps winbar,` while the other does not.


<details>

<summary>GitHub Copilot (HTTP):</summary>

Screenshot 2025-10-28 at 5 08 20 PM


https://github.com/user-attachments/assets/48d1b600-91a6-4bb4-8093-88e7d55cf5dc


</details>

<details>

<summary>Gemini_cli (ACP):</summary>

Screenshot 2025-10-28 at 5 08 58 PM



https://github.com/user-attachments/assets/f585e792-7ef8-4a25-a375-31f99ae48c6a


</details>

Thank you

bassamsdata:
@amadeus, I tested it with Claude Code and think I see what you mean, Claude tends to generate more concise diffs since it’s generally smarter about what to include. We don’t have direct control over Claude’s internal tools, so we can only display the code it sends back, which makes it behave a bit differently from the HTTP adapters.

For the keymaps, I tried running a minimal script and it worked fine with (g1, g2, etc.), so tbh, I’m not entirely sure about the issues you're having.

The code should now align properly in Split View (winbar should be on both windows), and I replaced the scratch buffer with a real one for a smoother UX. Maybe give it another try when you can?

Also, the monitor feature for ACP should be landing soon, it’ll show the full diff of all file system changes made by the LLM, which should make things much easier to follow.

Screenshot 2025-10-28 at 10 58 40 PM

RithwikVaidun:
I was actually working on this feature as well but for inline buffer diffs. You can can checkout the fork here: https://github.com/RithwikVaidun/codecompanion.nvim/tree/cursor-like
The code is a little messy and it's still a work in progress.  I implemented a small testing framework to test out the UI that could use some work.

amadeus:
So yeah, in the latest version, it definitely seems less buggy than before, however in general I don't think the experience works very well in that I'd prefer a proper vim diff to have context (allowing me to expand portions of the file etc. (Also not sure if there's some technical reason that the diff can't be applied to the buffer without saving, and then show a classic vim diff, but that would be ideal for me I think) 

Here's videos of doing essentially the same thing with sonnet vs claude code. claude code also suffers from a weird issue where you end up without the file anymore after the change which also isn't great.

Ideal Example using Sonnet (sorry for the low quality, was trying to get it under the 10mb limit for github):

https://github.com/user-attachments/assets/6d6ddfa0-6f5f-46f2-9b5e-beaef83499e7

Claude Code example:

https://github.com/user-attachments/assets/65477f35-eb2a-473c-b2cf-6bb005bad847

bassamsdata:
Hello @amadeus, thanks for the update, and sorry for the delay, I've had a bunch of stuff going on lately . I checked out how Zed handles ACP, and it looks like they're doing the same thing: showing the diff for what the LLM sends through ACP. 

Doing a full diff of the entire file would take a lot of work utilizing `insert_edit_into_file` and extra code, so honestly, I don't plan to implement that part.

Here's a screenshot from Zed showing what ACP sends and what the diff looks like. 
Screenshot 2025-11-13 at 3 36 00 PM


Also, the issue with the window close should be fixed now.

So I'd consider this PR ready to go, unless someone spot any issues.

Thank you

olimorris:
Thanks once again @bassamsdata. I think the QOL for most people using ACP is high so I'll merge this now.

olimorris:
@bassamsdata are you noticing that the winbar from the ACP diff window isn't being cleared properly?

2025-11-17 14_50_15 - WezTerm@2x

I'm unsure if this is because of the changes in `request_permission.lua`

bassamsdata:
I don’t think I’m experiencing this issue. Could you share which layout you’re using, is it the inline float layout?
I remember running into something similar once, but it should be fixed now. I’m trying to reproduce it on my end. Is the problem that the colors change when a winbar is already present? I’m asking because I don’t use a winbar at all.

Edit: Even though I’m not seeing the issue myself, I suspect there might be a bug where the winbar isn’t being cleared in the other window only when `provider = split`. Not sure if that’s the same problem you’re encountering, though.

edit2: @olimorris, so if the issue in colors, then yes we don't clear the `winhighlight` at all, and maybe we should add
`vim.wo[opts.winnr].winhighlight = ""` to the cleanup fcuntion. if that's the issue, I'll make a PR.

olimorris:
You're right that the winhighlight does fix the coloring. I didn't test this PR before merging (that's on me), and I've noticed that now I don't even seem to get a float at all, something which I quite like for my ACP workflow:

https://github.com/user-attachments/assets/9021ccac-abc7-4059-9edd-b4d0309e49c1

Is there a config option for that? I can't find that documented.

bassamsdata:
As for the floating window, that’s actually the default behavior. It should use the floating layout unless you explicitly set `layout = "buffer"` in config. Here’s the relevant option:

https://github.com/olimorris/codecompanion.nvim/blob/dbc54cb86bc169ab49a510786ef8cc4d10293c3b/lua/codecompanion/config.lua#L1405

Regarding `winhighlight`: this PR didn’t actually change anything in that area, since we weren’t clearing `winhighlight` to begin with.. If the issue is only related to winhighlight, I have a [fix](https://github.com/bassamsdata/codecompanion.nvim/tree/fix_winbar_highlight) prepared for it so I can PR it now

edit: typos: fix, not a PR

edit2:  the option is [documented](https://codecompanion.olimorris.dev/configuration/chat-buffer#diff):

bassamsdata:
> the winhighlight does fix the coloring

After reading more about the winbar, I probably should have added this `winhighlight` from the start. Apologies for that.

olimorris:
> After reading more about the winbar, I probably should have added this winhighlight from the start. Apologies for that.

No not at all. These UI elements in Neovim are cumbersome to test especially when you're relying on an LLM to trigger the behavior that causes the UI to change.

I was being a dumbass...I run two Neovim configs, for v0.11 and nightly. I forgot that with nightly I had the layout set to buffer than the default float 🤦🏼.

PR #2327, opened 2025-11-01
refactor(images): introduce image utils and refactor image handling

Davidyz:
## Description

Refactor the `/images` slash command and take the image preprocessing out of the slash command.

The ultimate goal (TBD in a future PR) is to make it easier for tools to make use of images. For example, the existing `@search_web` tool which uses tavily API supports [searching for images](https://docs.tavily.com/documentation/api-reference/endpoint/search#body-include-images). ~This may eventually require modifications to `CodeCompanion.Chat` or even some adapters, but for this PR, I think it's better to keep the scope small.~ The OpenAI API doesn't seem to support multimodal tool call results. Tools that return images might need to attach the images (base64) as a regular user message. In this case, the existing `helper.add_image` might suffice.

Verified pickers: 
- [x] default
- [x] telescope (no preview)
- [x] snacks

## Checklist

### DIFF
--- a/lua/codecompanion/utils/images.lua
+++ b/lua/codecompanion/utils/images.lua
@@ -78,7 +78,6 @@ function M.from_url(url, ctx, cb)
   local result = string.format("Could not get the image from %s.", url)
 
   local extmark_id = nil
-
   local ns = nil
   if ctx.chat_bufnr then
     ns = "codecompanion_fetch_image_" .. tostring(ctx.chat_bufnr)
@@ -94,6 +93,12 @@ function M.from_url(url, ctx, cb)
     proxy = config.adapters.http.opts.proxy,
     output = loc,
     callback = function(response)
+      if extmark_id then
+        extmark_id = nil
+        vim.schedule(function()
+          ui_utils.clear_notification(ctx.chat_bufnr, { namespace = ns })
+        end)
+      end
       if response.status ~= 200 then
         result = string.format(
           "Could not get the image from %s.\nHTTP Status: %d\nError: %s",
@@ -122,13 +127,6 @@ function M.from_url(url, ctx, cb)
             cb(result)
           end)
         end
-      end
-
-      if extmark_id then
-        vim.schedule(function()
-          extmark_id = nil
-          ui_utils.clear_notification(ctx.chat_bufnr, { namespace = ns })
-        end)
       end
     end,
   })


Davidyz:
@olimorris, this PR should be functional as-is. I've tested loading an image from a URL or from the snacks picker, and they both work as intended. But before I dive into this too far, I'd like to know whether you're okay with where this is going?

Davidyz:
I also noticed that `vim.base64` can actually handle images. This is probably because [you can store any binary data into a string](https://www.lua.org/pil/2.4.html) in lua. Thus, we could probably ditch the CLI-based base64 implementation and just use `vim.base64.encode`

olimorris:
Yes, absolutely happy with these suggested changes.

lua/codecompanion/utils/base64.lua
olimorris, line 1: Not sure we need the base64 utility file anymore

Davidyz: I guess that depends on whether you're comfortable with categorising pdf as images lol

When I was playing around with Gemini API, I was a bit surprised to see that it handled a URL to a pdf on arxiv just like an image. My local llama.cpp instance doesn't work with pdf though. I'll try that with different providers and decide what to do with it. Maybe it's just Google quietly deployed a RAG behind its openai-compatible endpoint. 

That said, it's probably a good idea to add pdf to the hard coded mime type mappings anyway, because we already know a provider that works with it. I can see this being helpful for people who are trying to reproduce the results from a paper.

olimorris: I meant, we could either put the base64 logic in image.lua itself or just add an `encode` to utils/init.lua. And the mimetype logic should move over to image.lua anyway.

Davidyz: Yep sure. That'll mean one less item in checkhealth

Davidyz: I ended up putting the base64 encode into `utils.files` because it seems to contain more file IO stuff. I hope that's ok.

Davidyz:
Moving the `add_image` function from helpers to `Chat`. I personally find this more intuitive, and when we bring image support to the `search_web` tool, we won't have to `require` the helper.

Davidyz:
Regarding the use of images in tools, I plan to implement the following in a separate PR: 
1. Add an option and a parameter to the existing search tool to allow the LLM to request image results from the searches. In the search tool, the images won't be added yet. This is to give the LLMs an opportunity to review the descriptions, etc. to decide whether they want a particular image.
2. Introduce a new tool, `fetch_images`, which accepts an array of URLs to images. This tool will then call the functions implemented in this PR to add the requested images to the chat.

Ofc these are just preliminary design that haven't been tested, and are subjected to changes.

PR #2348, opened 2025-11-05
Current Editing (Buffer) Context Tool

dv-blk:
## Description

Hello 👋 

I was having some fun learning to use your plugin this weekend and tried adding some functionality I've seen in other plugins. At first trying to shim the #{buffer} variable into all my prompts.

This is generated code, a crudely initialized MRU helper and hijacking a line in the default tool system prompt, just a proof of concept to gauge interest, I can likely contribute more time to improving this code if so.

In other chat plugins, edit prompts, without specifying additional variables have the ability to infer where to make an edit.

This adds a 'most recently used file editing buffer' helper and a read-only `get_editing_context` tool so users can issue edit prompts without needing to specify where they are and what lines they are referencing.

This can be called several times in the same conversation, to provide users current editing position when they move during it - not very well at the moment, it is hit and miss with remembering to check on consecutive edits and will use the previous information, perhaps this is a limitation with gpt/llms I have not fully realized. 

## Related Issue(s)

This resolves my own limitation (not knowing how yet) to toggle a chat with a custom user prompt that adds the #{buffer} variable:
https://github.com/olimorris/codecompanion.nvim/discussions/2339

https://github.com/olimorris/codecompanion.nvim/discussions/2232
https://github.com/olimorris/codecompanion.nvim/discussions/1976
## Screenshots

image

## Checklist

### DIFF
--- a/lua/codecompanion/config.lua
+++ b/lua/codecompanion/config.lua
@@ -250,7 +250,7 @@ Never use a tool that does not exist. Use tools using the proper procedure, DO N
 Never say the name of a tool to a user. For example, instead of saying that you'll use the insert_edit_into_file tool, say "I'll edit the file".
 If you think running multiple tools can answer the user's question, prefer calling them in parallel whenever possible.
 When invoking a tool that takes a file path, always use the file path you have been given by the user or by the output of a tool.
-When user makes a request to edit code, always use get_editing_context to understand the current file and cursor position before making any edits.
+Each time a user makes another request to edit code, use get_editing_context everytime to understand the current file and cursor position before making any edits.
 </toolUseInstructions>
 <outputFormatting>
 Use proper Markdown formatting in your answers. When referring to a filename or symbol in the user's workspace, wrap it in backticks.


olimorris:
Thanks for taking the time to contribute.

I'm not clear on what the use case for something like this would be. If I use an LLM to generate any code, I give it access to the `@{files}` tool. So it can determine which files to edit and where to make those edits and it has the autonomy to figure that out, saving me sharing a ton of context.

Your example in the image above is very trivial, something that you'd do with a few keystrokes yourself in Vim. It's for this reason that I'm not understanding where the value in something like this would be compared to what we have already.

dv-blk:
In VS Code, chat is aware of what window is currently active and it's file, It is not necessary to make any reference to it while continuing to chat and making edits among various files.

Here's is an example use between files, and full_stack_dev featuring get_editor_context, for the same prompt.

Files taking a general approach, removing all the comments from my vim config:
image

Prompting with a little bit of context about most recently active file editor:
image

(I can try with only get_editor_context, but I may have to move around what I added as a prompt first)

It has added awareness that I am likely referring to this file and the lines appearing on the screen.

olimorris:
I'm still not clear what the value add is. Why wouldn't you use the inline assistant? Or, why aren't the buffer and viewport variables sufficient? Couldn't the latter be updated to incorporate this?

dv-blk:
The value add would be for those that, when preferring the use of chat over the inline assistant (e.g.. refactoring code while GPT maintains the context of previous edits in that chat), not needing to specify variables at all - Omitting those variables in the inline assistant as well, assuming it has access to the same context.
  
  In chat the context added from variables can't be removed either, I'm guessing that is only visually relevant, though.

olimorris:
I'll ponder this. I'm not opposed to it conceptually, I think giving an LLM context of what you see and what you're doing in Neovim is helpful. If you can provide some more examples of what prompts you're sending to the LLM and help me better understand how people would use it, that would be good.

A tool is definitely not the right way to do accomplish this. A user should be "giving" context to an LLM rather than an LLM "requesting" it. Again. sharing more examples of how this would be used will help me understand how we could implement this.

dv-blk:
Okay that makes sense, sounds good, thanks for the feedback! 

I will noodle on how to integrate this better when I get time, I think that will help demonstrate.

PR #2354, opened 2025-11-06
feat(diff): Add Diff Hunk Navigation

bassamsdata:
## Description

Adds navigation between diff hunks in inline diffs with `]h` (next) and `[h` (previous) keymaps.

### Changes

**Navigation**
- Added `jump_to_next_hunk()` and `jump_to_prev_hunk()` methods to InlineDiff
- Wraps around when reaching first/last hunk
- works in both `HTTP` and `ACP`

**Winbar**
- Now default for all inline layouts (buffer and floating)
- Shows keymaps with colors and hunk count
- Automatically truncates for narrow windows
- Replaces previous extmark hints approach

### Config

Navigation keymaps are configurable via `strategies.inline.keymaps`:
```lua
next_hunk = { modes = { n = "]h" } }
prev_hunk = { modes = { n = "[h" } }
```

**Note:** Please let me know if there’s anything you don’t like in the PR, especially regarding the winbar. I tested the diff with many themes and everything worked correctly. I feel that colored keymaps are easier to spot.

Thank you

## Related Issue(s)

 #2152

## Screenshots

Screenshot 2025-11-05 at 9 51 44 PM

Screenshot 2025-11-05 at 9 50 52 PM

https://github.com/user-attachments/assets/0fc866c0-0792-4210-b266-0db0473a9a93

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/tools/catalog/insert_edit_into_file/init.lua
+++ b/lua/codecompanion/strategies/chat/tools/catalog/insert_edit_into_file/init.lua
@@ -713,16 +713,17 @@ local function edit_file(action, chat_bufnr, output_handler, opts)
     })
   end
 
+  -- Write to DISK before creating diff
+  local write_ok, write_err = write_file_content(path, dry_run_result.final_content, file_info)
+  if not write_ok then
+    return output_handler({
+      status = "error",
+      data = fmt("Error writing to `%s`: %s", action.filepath, write_err),
+    })
+  end
+
   -- Auto-apply in YOLO mode
   if vim.g.codecompanion_yolo_mode then
-    local write_ok, write_err = write_file_content(path, dry_run_result.final_content, file_info)
-    if not write_ok then
-      return output_handler({
-        status = "error",
-        data = fmt("Error writing to `%s`: %s", action.filepath, write_err),
-      })
-    end
-
     return output_handler({
       status = "success",
       data = fmt("Edited `%s` file%s", action.filepath, extract_explanation(action)),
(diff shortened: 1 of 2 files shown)


bassamsdata:
his would require significant changes after v18. If we agree to introduce hunk-based navigation for diffs, it would be better to open a new PR targeting v18, so I’m closing this for now. If anyone is interested in picking this up, please feel free to do so.

olimorris:
Still related to diffs, I'd love to get your views on expanding the inline diff as originally discussed in #2508. I was taking a look at how Sidekick.nvim works and it looks like folke's employing a word and character level diff, potentially with Tree-sitter becoming involved.

I've got some time over the holiday season so was thinking of having a play.

bassamsdata:
That would be great. I’m not very familiar with how Folke integrated Tree-sitter, but it’s an interesting approach to include it. Is it mainly used for indentation, or is it also involved in calculating diffs like smart semantic diffs?

I’m currently working on implementing inline diffs as a standalone plugin, and it will support word and character-level diffs (but not based on treesitter and it's based on `iskeyword`) . Adding word/character diffing is fairly straightforward in CodeCompanion and largely similar to how we currently use `vim.text.diff`, just with different options and applied only to the lines where it’s needed.

I really like how Zed handles this recently [#43269](https://github.com/zed-industries/zed/pull/43269). They enable word diffing when a hunk has changes with a similar number of lines (lines deleted equal lines added) and when there are five or fewer changed lines. This makes the diff much easier to read and less cluttered.

One thing worth noting is that I used NES in `sidekick.nvim` for a while, but I wasn’t happy with how it displays diffs, especially when it shows a one-line change as a single line. I think our approach is better. It also doesn’t always suggest changes in the right places, and pressing tab to complete sometimes results in incorrect final text. This doesn’t seem to be a Copilot issue, as I used `copilot-nes.nvim` for some time without encountering this problem. I’m just sharing my opinion based on that experience.

Regarding correct indentation, I’m not entirely sure how Tree-sitter would be difficult to implement. In my experience, it’s quite difficult for an LLM to consistently determine the correct indentation, as it depends heavily on the surrounding text and context. Locally, I modified the inline prompt to strongly emphasize preserving indentation, and it seems to work about 99% of the time.

> `- **CRITICAL**: ENSURE YOU PRESERVE THE EXACT INDENTATION (TABS/SPACES) as it appears in the provided code. Failure to maintain the original indentation will BREAK the user's code in the buffer, potentially causing syntax errors or unexpected behavior. This requirement is non-negotiable: always preserve indentation exactly as given.`

It would also be really nice to have the ability to toggle word diffing on and off directly within the diff view using a `vim.b` variable.

Sorry for the long comment :)

olimorris:
Oh wow that Zed diff is 🔥. That is absolutely the gold standard that I would love to have.

I think he uses Tree-sitter for the highlighting only. I don't think there's anything indentation wise going on. Also, I still don't think indentation is a CodeCompanion problem to fix if the LLM is outputting it incorrectly. But that's perhaps where myself and the person in the other issue disagree.

Your new plugin sounds very exciting!!!

PR #2359, opened 2025-11-07
feat(adapters): support extra fields in chat-completion response in OpenAI adapter

Davidyz:
## Description

Add an optional `parse_extra` handler that parses extra (non-standard) fields in the OpenAI `chat/completions` responses. This can be used by downstream adapters (openrouter, gemini, deepseek) to render their in-house reasoning format. This handler will be defined in either the downstream adapter (deepseek, gemini, etc.) or the user config (when they use `extend` to customise the adapter), and will only be called when there's a non-`nil` extra field.

## Related Issue(s)

Compared to #1938, this is a more flexible and less invasive design that can be made to work with more OpenAI-based APIs.
If this PR is accepted, I'll update #2306 to use this too.

## Checklist

### DIFF
--- a/doc/codecompanion.txt
+++ b/doc/codecompanion.txt
@@ -1,4 +1,4 @@
-*codecompanion.txt*        For NVIM v0.11        Last change: 2025 November 17
+*codecompanion.txt*        For NVIM v0.11        Last change: 2025 November 19
 
 ==============================================================================
 Table of Contents                            *codecompanion-table-of-contents*
@@ -643,7 +643,7 @@ CodeCompanion, you simply need to follow their Getting Started
 <https://docs.augmentcode.com/cli/overview#getting-started> guide.
 
 
-SETUP: DOCKER CAGENT ~
+SETUP: CAGENT ~
 
 To use Docker’s Cagent <https://github.com/docker/cagent> within
 CodeCompanion, you need to follow these steps:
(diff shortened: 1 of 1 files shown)


Davidyz:
This is still WIP (needs tests, typing and better docs), but @olimorris would you accept this as an alternative to #1938? 

@SDGLBL, since you wrote the original PR, I'd love to know your thoughts on this too (there's an example snippet in the docs about how to configure this for the openrouter format).

olimorris:
Totally down for this. Be a welcome addition.

lua/codecompanion/adapters/http/init.lua
```diff
@@ -50,6 +50,7 @@ local function get_handler(adapter, name)
     parse_chat = "chat_output",
     parse_inline = "inline_output",
     parse_tokens = "tokens",
+    parse_extra = "parse_extra",
```
olimorris: Propose we call this `parse_meta` as this will be the extension point for anything else in the adapter

Davidyz:
Nice. I'll also try to refactor the deepseek adapter to use this design. If that works, we'd automatically have unit tests for `parse_extra`.

On the deepseek documentation, they're using the openai python SDK with the `/chat/completion` endpoint, which is not the one that they claim to be openai-compatible (`/v1/chat/completion`). I think we can reasonably assume that the former endpoint (same as the one used by the cc deespeek adapter) is also openai compatible, just with the extra `reasoning_content` field.

Davidyz:
The tests are failing because [the openai `chat_output` handler didn't use `tool.index` from the response, but rather the 1-based iteration variable](https://github.com/olimorris/codecompanion.nvim/blob/991dd81ac37b56b6d13529a08e86a42d183d79dc/lua/codecompanion/adapters/http/openai.lua#L232) when non-streaming. When I change this to `_index = tool_index`, many other tests failed (that is, tests for other openai-based adapters). I looked up the openai docs and it seems like [`index` isn't part of the non-stream `tool_calls` response](https://platform.openai.com/docs/api-reference/chat/object#chat-object-choices-message-tool_calls), so I'll just change the index in the mocked response here.

lua/codecompanion/adapters/http/init.lua
Davidyz: I'm naming it `parse_meta` in the new adapter API. As for the legacy API, since the handlers aren't grouped into `response`, `request`, etc. I'll call it `parse_message_meta` so that it's clearer that this handler applies to the message/delta.

olimorris:
Awesome stuff, this looks good to me. Could you run `make docs`? Also, does the `test_json.lua` file do anything?

Davidyz:
> Also, does the `test_json.lua` file do anything?

No. That was unintended. Just removed it.

PR #2365, opened 2025-11-08
fix(mistral tools): Ensure we enqueue user messages after tool calls

flovilmart:
## Description

Mistral strictly requires that tool are followed by an assistant message. 
In certain scenarios, a user message may get interleaved in the conversation breaking the complete flow:

STR: 
- Start a session with `@full_stack_dev`
- Create a `hello` file, with some basic content
- Add the `hello` file with `/buffer` command to the converation
- Ask the agent to update the hello file with some new content
- Notice the session will break either immediately or on the next turn with the following error.

Root cause:
- I can trace the root cause of that particular use case to the file watcher attached to the buffer which syncs back the buffer to the LLM every times the buffer is updated.
- Given the buffer is updated by the tool call, the LLM don't have the time to get the assistant response to the tool call, which is required by Mistal's API. the correct order is `tool` -> `assistant` -> `user`. If we have `tool` -> `user` -> `assistant` the message will break.

The fix:
- if we see a `tool` followed by `user` we push the message in a pending stack. 
- As long as we have `tool`  followed by a  `user` call we keep popping them from the messages. 
- If we don' have any other messages, we keep them for the next round of interactions - they will be flushed
- if we see a message that is not `tool` nor `user` (likely assistant), we pop back all messages.

Ex:
-  [t, u1, u2, a] -> [t, a, u1, u2]
- [t, u1, u2] -> [t] (pending = [u1,u2]

```
Output data:
{"object":"error","message":"Unexpected role 'user' after role 'tool'","type":"invalid_request_message_order","param":null,"code":"3230"}
[ERROR] 2025-11-07 19:08:30
Error: {"object":"error","message":"Unexpected role 'user' after role 'tool'","type":"invalid_request_message_order","param":null,"code":"3230"}
```

When this happens, the session is fully broken

## Checklist

- [x] _(optional)_ I've updated `CodeCompanion.has` in the [init.lua](https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/init.lua#L239) file for my new feature

### DIFF
--- a/lua/codecompanion/adapters/http/mistral.lua
+++ b/lua/codecompanion/adapters/http/mistral.lua
@@ -61,20 +61,33 @@ return {
       return openai.handlers.form_parameters(self, params, messages)
     end,
     form_messages = function(self, messages)
+      local pending_messages = self.pending_messages or {}
       local is_previous_tool = false
-      local messages = vim.iter(messages):filter(function(msg)
+      for k, msg in ipairs(messages) do
         local is_tool = msg.role == "tool"
         local is_user = msg.role == "user"
-        local keep = true
         -- Mistral does not like user after tool messages, those should always be assistant
-        -- Worst case we drop 1 user message, but the assistant will respond to the tool
         if is_previous_tool and is_user then
-          log:debug("Dropping user after tool message: %s", msg.content)
-          keep = false
+          table.insert(pending_messages, msg)
+          messages[k] = nil
+          -- message was dropped, so for the next message, the previous one
+          -- is still a tool
+          is_tool = true
+        else
+          if not is_previous_tool then
+            -- Flush pending messages whenever we can
+            for i, m in ipairs(pending_messages) do
+              table.insert(messages, m)
+            end
+            pending_messages = {}
+          end
         end
         is_previous_tool = is_tool
-        return keep
-      end)
+      end
+
+      -- Keep the pending messages for next round
+      self.pending_messages = pending_messages
+
       return openai.handlers.form_messages(self, messages)
     end,
     chat_output = function(self, data, tools)


flovilmart:
@olimorris  I went with a very naive implementation with the filter and a boolean keeping the side effect from the previous message, not my best code, but that works. 
Happy to revisit implementation with your input 😉

olimorris:
Should we discard the user's message? Or, store it and insert it into the message stack at an appropriate time in the future? My concern would be if I'm a user, my tool fails, I respond via the chat buffer and the LLM responds without ever seeing what I've typed. That would be unexpected and depending on what I've responded with, a potential waste of my time.

EDIT: To make the latter possible, I can make a PR that would allow `form_messages` to receive the chat buffer number or the ID. That way you could store "pending" messages in the adapter itself and clear them out when they are consumed. Doubt many people would have multiple Mistral chats going on at the same time, using tools, but just in case...

flovilmart:
> My concern would be if I'm a user, my tool fails, I respond via the chat buffer and the LLM responds without ever seeing what I've typed. 

This happens when we got a tool call response that hasn’t been answered by the LLM yet and the user has typed something into the buffer and sent it already. Not because the tool call failed, simply that it hasn’t been processed by the assistant yet

> Or, store it and insert it into the message stack at an appropriate time in the future?

That a good suggestion, but probably overkill. 

. I saw it happen multiple times with and empty string, I’d have to check more failures on my side and see the content of that buffer. Now with the debug log I’ll be able not to crash my sessions and continue. Worst, case this is a simple copy paste from the previous chat message.
Personally I think this workaround is plenty enough.

flovilmart:
Another example after using an edit tool and accepting the change:

```
    {
      "role": "assistant",
      "content": "I'll edit the `tests/adapters/http/test_mistral.lua` file to add the new test case. Here's the code to add:\n\n````lua\n// filepath: tests/adapters/http/test_mistral.lua\n// ...existing code...\n{ role = \"tool\", content = \"ok\" },\n{ role = \"assistant\", content = \"that works\" },\n// ...existing code...\nT[\"Mistral adapter\"][\"form_messages\"][\"it skips messages when user follows tools\"] = function()\n// ...existing code...\n{ role = \"tool\", content = \"ok\" },\n// ...existing code...\nT[\"Mistral adapter\"][\"form_messages\"][\"it can form messages with assistant after tool\"] = function()\n  local input = {\n    { role = \"user\", content = \"User1\" },\n    {\n      role = \"llm\",\n      tools = {\n        calls = {\n          {\n            [\"function\"] = {\n              arguments = '{\"location\":\"London, UK\",\"units\":\"fahrenheit\"}',\n              name = \"weather\",\n            },\n            id = \"call_1_a460d461-60a7-468c-a699-ef9e2dced125\",\n            type = \"function\",\n          },\n        },\n      },\n    },\n    { role = \"tool\", content = \"ok\" },\n    { role = \"assistant\", content = \"Assistant response\" },\n  }\n\n  local expected = {\n    messages = {\n      {\n        content = \"User1\",\n        role = \"user\",\n      },\n      {\n        role = \"llm\",\n        tool_calls = {\n          {\n            [\"function\"] = {\n              arguments = '{\"location\":\"London, UK\",\"units\":\"fahrenheit\"}',\n              name = \"weather\",\n            },\n            id = \"call_1_a460d461-60a7-468c-a699-ef9e2dced125\",\n            type = \"function\",\n          },\n        },\n      },\n      { role = \"tool\", content = \"ok\" },\n      { role = \"assistant\", content = \"Assistant response\" },\n    },\n  }\n\n  h.eq(expected, adapter.handlers.form_messages(adapter, input))\nend\n// ...existing code...\n````\n\nThis test case verifies that the mistral adapter can properly form messages when there's an assistant message following a tool message. The test checks that the assistant message is included in the output without any modifications."
    },
    { "role": "user", "content": "EDIT THE FILE" },
    {
      "role": "assistant",
      "tool_calls": [
        {
          "id": "5xJ4yNuGq",
          "function": {
            "arguments": "{\"filepath\": \"tests/adapters/http/test_mistral.lua\", \"mode\": \"append\", \"dryRun\": false, \"explanation\": \"Adds test for the mistral adapter for a case where there is an assistant message after the tool. it forms messages properly\", \"edits\": [{\"newText\": \"T[\\\"Mistral adapter\\\"][\\\"form_messages\\\"][\\\"it can form messages with assistant after tool\\\"] = function()\\n  local input = {\\n    { role = \\\"user\\\", content = \\\"User1\\\" },\\n    {\\n      role = \\\"llm\\\",\\n      tools = {\\n        calls = {\\n          {\\n            [\\\"function\\\"] = {\\n              arguments = '{\\\"location\\\":\\\"London, UK\\\",\\\"units\\\":\\\"fahrenheit\\\"}',\\n              name = \\\"weather\\\",\\n            },\\n            id = \\\"call_1_a460d461-60a7-468c-a699-ef9e2dced125\\\",\\n            type = \\\"function\\\",\\n          },\\n        },\\n      },\\n    },\\n    { role = \\\"tool\\\", content = \\\"ok\\\" },\\n    { role = \\\"assistant\\\", content = \\\"Assistant response\\\" },\\n  }\\n\\n  local expected = {\\n    messages = {\\n      {\\n        content = \\\"User1\\\",\\n        role = \\\"user\\\",\\n      },\\n      {\\n        role = \\\"llm\\\",\\n        tool_calls = {\\n          {\\n            [\\\"function\\\"] = {\\n              arguments = '{\\\"location\\\":\\\"London, UK\\\",\\\"units\\\":\\\"fahrenheit\\\"}',\\n              name = \\\"weather\\\",\\n            },\\n            id = \\\"call_1_a460d461-60a7-468c-a699-ef9e2dced125\\\",\\n            type = \\\"function\\\",\\n          },\\n        },\\n      },\\n      { role = \\\"tool\\\", content = \\\"ok\\\" },\\n      { role = \\\"assistant\\\", content = \\\"Assistant response\\\" },\\n    },\\n  }\\n\\n  h.eq(expected, adapter.handlers.form_messages(adapter, input))\\nend\\n\", \"replaceAll\": false, \"oldText\": \"{ role = \\\"tool\\\", content = \\\"ok\\\" },\\n    { role = \\\"assistant\\\", content = \\\"that works\\\" },\"}]}",
            "name": "insert_edit_into_file"
          }
        }
      ]
    },
    {
      "role": "tool",
      "tool_call_id": "5xJ4yNuGq",
      "content": "Edited `tests/adapters/http/test_mistral.lua` buffer\nAdds test for the mistral adapter for a case where there is an assistant message after the tool. it forms messages properly"
    },
    {
      "role": "user",
      "content": "<attachment filepath=\"tests/adapters/http/test_mistral.lua\" buffer_number=\"56\">The file `tests/adapters/http/test_mistral.lua`, has been modified. Here are the changes:\n```diff\n@@ -50,8 +50,55 @@\n         },\n       },\n     },\n+T[\"Mistral adapter\"][\"form_messages\"][\"it can form messages with assistant after tool\"] = function()\n+  local input = {\n+    { role = \"user\", content = \"User1\" },\n+    {\n+      role = \"llm\",\n+      tools = {\n+        calls = {\n+          {\n+            [\"function\"] = {\n+              arguments = '{\"location\":\"London, UK\",\"units\":\"fahrenheit\"}',\n+              name = \"weather\",\n+            },\n+            id = \"call_1_a460d461-60a7-468c-a699-ef9e2dced125\",\n+            type = \"function\",\n+          },\n+        },\n+      },\n+    },\n     { role = \"tool\", content = \"ok\" },\n-    { role = \"assistant\", content = \"that works\" },\n+    { role = \"assistant\", content = \"Assistant response\" },\n+  }\n+\n+  local expected = {\n+    messages = {\n+      {\n+        content = \"User1\",\n+        role = \"user\",\n+      },\n+      {\n+        role = \"llm\",\n+        tool_calls = {\n+          {\n+            [\"function\"] = {\n+              arguments = '{\"location\":\"London, UK\",\"units\":\"fahrenheit\"}',\n+              name = \"weather\",\n+            },\n+            id = \"call_1_a460d461-60a7-468c-a699-ef9e2dced125\",\n+            type = \"function\",\n+          },\n+        },\n+      },\n+      { role = \"tool\", content = \"ok\" },\n+      { role = \"assistant\", content = \"Assistant response\" },\n+    },\n+  }\n+\n+  h.eq(expected, adapter.handlers.form_messages(adapter, input))\n+end\n+\n   }\n \n   local expected = {\n```</attachment>"
    }
```

flovilmart:
actually it comes from [here](https://github.com/flovilmart/codecompanion.nvim/blob/cb260b45f589165c6ff5748ae1c2c134f22b0b75/lua/codecompanion/strategies/chat/watchers.lua#L143), when the buffer gets updated after accepting a change 🤔

flovilmart:
@olimorris  any idea? Can we disable the file watcher for mistral otherwise if the file is in the buffer?

olimorris:
I can't accept the PR in its current form as I think it could negatively impact the UX for a Mistral user, as I outlined above. Dropping a user's message fixes the issue but introduces another one further down the line.

> Can we disable the file watcher for mistral otherwise if the file is in the buffer?

No. I appreciate this fixes this problem. But again, we're impacting the user experience. I.e. what a user expects versus what is happening. If I've shared a watched buffer with the LLM, then I expect CodeCompanion to hold up its end of the bargain, and share that buffer with the LLM.

This should be worked around in the Mistral adapter, with test coverage.

flovilmart:
ok so the proper expectation is that the `invisible` messages would be enqueued, before we flush them in the next round? is that right @olimorris ?

olimorris:
The expectation would be that Mistral doesn't throw an error and that no user messages are discarded.

flovilmart:
I have reworked the implementation to make sure we don't drop any messages but keep them in a pending queue on `self` They are re-injected after the 1st opportunity to make sure they're ~almost~ in order. 

One of the key challenges is that I can't mutate the underlying `messages` array to reorder them fully. So I have pop and push them every cycle.
Let me know what you think of this approach and if you see any caveats.

flovilmart:
@olimorris  any chance you can review / play with the code here? I would love to get this merged

yyxi:
I'm experiencing the same issue. Thanks @flovilmart for the fix and @olimorris for CodeCompanion — it would be great to have this resolved.

olimorris:
Not forgotten about this PR. Will get around to testing it sometime after version 18 is released. Quick scan of the test and coverage doesn't look complete. I can't see where we're testing that pending messages are added to the chat buffer.

flovilmart:
I need to find some time to add more tests 😅

PR #2381, opened 2025-11-14
feat(mistral): Override env when using codestral

flovilmart:
When using codestral-latest, I wanna use the codestral API endpoint rather than the mistral one as it is free to use.

### DIFF
--- a/lua/codecompanion/adapters/http/mistral.lua
+++ b/lua/codecompanion/adapters/http/mistral.lua
@@ -46,6 +46,13 @@ return {
           self.opts.tools = false
         end
       end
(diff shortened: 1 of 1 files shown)


olimorris:
Firstly, can you respect the PR template. I try and keep standards and consistency up across the project.

Secondly, can you share a reference to this in Mistrals documentation? Otherwise this just looks like your preference rather than something Mistral ask us to do.

flovilmart:
You're right, that my personal preference to use the free to use model. 
I'll close the PR and let others figure out how to set it up.

https://mistral.ai/news/codestral

PR #2387, opened 2025-11-15
feat: implement JSON argument validation for insert_edit_into_file

jinzhongjia:
- Added functions to check for non-serializable types and validate JSON arguments.
- Enhanced error handling for invalid edits, ensuring required fields are present and correctly typed.
- Introduced comprehensive tests for various JSON validation scenarios, including edge cases and malformed JSON inputs.

## Description

When I user claude sonnet 4.5 and opus 4.1, I meet this problem:

The llm ouputs incorrect tool call (contains unencoded characters)

## Screenshots

image

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/tools/catalog/insert_edit_into_file/init.lua
+++ b/lua/codecompanion/strategies/chat/tools/catalog/insert_edit_into_file/init.lua
@@ -199,8 +199,7 @@ local function validate_json_arguments(args)
       error_path
     )
     return false,
-      fmt(
-        [[JSON argument validation failed: Invalid value type detected
+      fmt([[JSON argument validation failed: Invalid value type detected
 
 Your JSON arguments contain invalid values. Please ensure all values are valid JSON types:
 - Strings (text)
@@ -210,8 +209,7 @@ Your JSON arguments contain invalid values. Please ensure all values are valid J
 - Objects {}
 - null
 
-Do not include any code references, variables, or computed values - only literal JSON values.]]
-      )
+Do not include any code references, variables, or computed values - only literal JSON values.]])
   end
 
   -- Try to serialize to catch other JSON encoding issues (like circular references)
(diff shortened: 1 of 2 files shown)


jinzhongjia:
I am currently unable to consistently reproduce this issue. Occasionally, when making a significant modification, LLM will not output the correct content as prompted.

bassamsdata:
I hit a similar error once with Claude Sonnet 4.5 while using the Copilot adapter, and, weirdly enough, I never saw it again. Looking at your error though, I really don't think it's coming from the `insert_edit_into_file` tool. It doesn't look like the JSON even made it that far. The function-tooling JSON itself seems broken, so the real failure is probably coming from `tools/init.lua:128`.

The JSON in the error doesn't seem to have the final closing bits - it's missing the `]}}` at the end.

If you don't mind checking a couple things for me please:

* after the error showed up, did the chat get corrupted? Like, were you unable to send anything afterward?
* and if you still have that chat open, could you hit `gd` and share the Arguments it shows there?
* or, if you had debug mode on, maybe search for "Args" and share that chunk?

I've got a small fix locally for this, just haven't had a moment to reproduce it cleanly and send a PR.

jinzhongjia:
When I meet this error,I can't send anything

jinzhongjia:
No, no, no, actually, the purpose of this PR is to verify whether the input of the edit tool can be parsed into a valid JSON.

bassamsdata:
> When I meet this error, I can't send anything

Ah, got it - thanks for confirming that. That's exactly what happened on my side too, so yeah, it lines up with the same issue. It basically blows up *before* the tool even gets a chance to run.

I'll put together a PR for this. It's a weird edge case, but the fix is small, and it should also stop the chat from getting stuck afterward.

> actually, the purpose of this PR is to verify whether the input of the edit tool can be parsed into a valid JSON.

Appreciated!,  just to clarify though, that means this PR isn't really addressing the error you were running into. It's more about checking that the edit-tool input is valid JSON in the first place.

We already have validation + proper error messages on the edit-file tool side, so the issue you hit is happening *before* that layer even gets a chance to run.

jinzhongjia:
OK, I get this, so I think I can close this pr

olimorris:
I can confirm I've had this once before with `sonnet-4.5`. The actual issue is with Anthropic sending over incorrect JSON. But that is not within CodeCompanion's purview.

But, this shouldn't be leaving the chat buffer broken and unrecoverable. If that is the case, then happy to have a PR to tools/init.lua to fix that.

bassamsdata:
> The actual issue is with Anthropic sending over incorrect JSON.

Yep, exactly, this one isn't on CodeCompanion.

> then happy to have a PR to tools/init.lua to fix that.

Sounds good. I'll put one today.

PR #2390, opened 2025-11-16
Prevent chat from getting stuck when the LLM sends malformed JSON

bassamsdata:
## Description

This one just handles malformed JSON tools schema from the LLM a bit more gracefully, so the chat doesn't get corrupted and stuck afterward. That way we avoid the whole HTTP 400 mess.

**What changed:** (`tools/init.lua`)

- catch the JSON error and give the LLM a clearer, more useful message
- drop the broken message from the chat stack so the chat keeps functioning normally

Thank you

## Related Issue(s)

#2387

## Checklist

### DIFF
--- a/lua/codecompanion/strategies/chat/tools/init.lua
+++ b/lua/codecompanion/strategies/chat/tools/init.lua
@@ -86,12 +86,61 @@ function Tools:_handle_tool_error(tool, error_message)
   return utils.fire("ToolsFinished", { bufnr = self.bufnr })
 end
 
----Resolve and prepare a tool for execution
----@param tool table The tool call from the LLM
+---When an LLM sends a tool call with malformed structure or JSON, the assistant message
+---containing tool_calls gets added to chat.messages by Chat:done() before tools:execute().
+---If we don't remove this orphan message (one with tool_calls but no results), subsequent
+---API requests will fail with HTTP 400 errors, making the chat unusable.
+
+---Remove a malformed assistant message from chat history to prevent corruption.
+---@param tool table The malformed tool with an id field used for verification
+---@return nil
+function Tools:_remove_malformed_message(tool)
+  if #self.chat.messages == 0 then
+    return
+  end
+
+  local last_msg = self.chat.messages[#self.chat.messages]
+
+  -- Verify this is the message containing our malformed tool call
+  if last_msg.role == config.constants.LLM_ROLE and last_msg.tools and last_msg.tools.calls then
+    -- Check if THIS specific tool call is in the message
+    local found_malformed_call = false
+    for _, call in ipairs(last_msg.tools.calls) do
+      if call.id == tool.id then
+        found_malformed_call = true
+        break
+      end
+    end
+
+    if found_malformed_call then
+      log:debug("Removing malformed assistant message with tool call %s from history", tool.id)
+      table.remove(self.chat.messages, #self.chat.messages)
+    else
+      log:warn("Expected to find tool call %s in last message but didn't - not removing", tool.id)
+    end
+  else
+    log:debug(
+      "Expected last message to be assistant with tool_calls for tool %s, but found role=%s with tools=%s",
+      tool.id,
+      last_msg.role or "unknown",
+      last_msg.tools and "yes" or "no"
+    )
+  end
+end
+
+---@param tool table The tool to resolve
 ---@return table|nil resolved_tool The resolved tool or nil if failed
 ---@return string|nil error_msg Error message if resolution failed
 ---@return boolean|nil is_json_error Whether this is a JSON parsing error that needs special handling
 function Tools:_resolve_and_prepare_tool(tool)
+  -- CRITICAL: Validate tool structure FIRST before accessing any fields
+  -- In rare cases, LLMs send malformed tool calls where tool["function"] is a string
+  if type(tool["function"]) ~= "table" then
+    log:error("Malformed tool call: 'function' field is not a table, got %s", type(tool["function"]))
+    self:_remove_malformed_message(tool)
+    return nil, "Malformed tool structure", true
+  end
+
   local name = tool["function"].name
   local tool_config = self.tools_config[name]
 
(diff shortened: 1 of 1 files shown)


lua/codecompanion/strategies/chat/tools/init.lua
```diff
@@ -109,40 +110,65 @@ function Tools:_resolve_and_prepare_tool(tool)
...
+        -- Remove the malformed assistant message from history BEFORE adding tool output
+        -- The assistant message with tool_calls was added by Chat:done() before tools:execute()
+        if
+          #self.chat.messages > 0
+          and self.chat.messages[#self.chat.messages].role == config.constants.LLM_ROLE
+          and self.chat.messages[#self.chat.messages].tools
+          and self.chat.messages[#self.chat.messages].tools.calls
+        then
+          log:debug("Removing malformed assistant message with tool calls from history")
+          table.remove(self.chat.messages, #self.chat.messages)
```
olimorris: Check: We want to remove the orphan message containing `tool_calls` which will have no corresponding result. My concern here is that we're just removing the last message in the chat. We should probably have some checks here.

olimorris:
Thanks for this. Very robust fix indeed and very much like the feedback to the LLM.

bassamsdata:
I’m delaying this a bit while I verify whether the fix will _actually_ resolve the stuck chat issue and output the error message.

jinzhongjia:
Is there any progress on this PR?

bassamsdata:
This PR is nearly complete I just want to confirm it works in real use. It behaves correctly in mocking.
The issue seems to occur only with Sonnet 4.5. I'll test it over the next two days with Sonnet 4.5 and update the PR accordingly.

olimorris:
Other users are reporting issues with sonnet and how it handles JSON, as an FYI, https://github.com/olimorris/codecompanion.nvim/discussions/2127#discussioncomment-15061503

bassamsdata:
Alright, this PR behaves as intended, which is both good and not-so-good. The good news is that it correctly removes the malformed JSON from the message stack. The downside is that when we add a new error message `to inform the LLM` to the chat stack, it ends up creating an orphaned `tool_result`. 

Here s what happens when the LLM sends malformed JSON in a tool call with the PR in its current state:

1. `_remove_malformed_message()` correctly removes the LLM message containing the `tool_use` block
2. BUT `add_tool_output()` was still being called, which adds a `tool_result` message referencing that `tool_use_id`
3. This creates an **orphaned `tool_result`** without a corresponding `tool_use`
4. On the next API call, LLM rejects with: `"unexpected tool_use_id found in tool_result blocks"
```lua
[ERROR] 2025-11-25 23:04:12
Error: {"error":{"message":"messages.214.content.1: unexpected `tool_use_id` found in `tool_result` blocks: toolu_01JkxZdLzY9rdXPR7JMddk5L. Each `tool_result` block must have a corresponding `tool_use` block in the previous message.","code":"invalid_request_body"}}
[ERROR] 2025-11-25 23:04:12
[chat::_submit_http] Error: {
  body = '{"error":{"message":"messages.214.content.1: unexpected `tool_use_id` found in `tool_result` blocks: toolu_01JkxZdLzY9rdXPR7JMddk5L. Each `tool_result` block must have a corresponding `tool_use` block in the previous message.","code":"invalid_request_body"}}',
  exit = 0,
  headers = { "cache-control: no-cache", "content-security-policy: default-src 'none'; sandbox", "content-type: text/plain; charset=utf-8", "strict-transport-security: max-age=31536000", "x-content-type-options: nosniff", "x-request-id: 00000-7390b49f-0087-4b74-a881-88277a2b472e", "date: Wed, 26 Nov 2025 04:04:12 GMT", "x-github-backend: Kubernetes", "x-github-request-id: C805:122F72:2A332:35BB2:69267C3C" },
  status = 400
}
```

I'll update the PR tomorrow. 

Thank you

olimorris:
Firstly, thanks for all your hard work on this. It's really appreciated as always.

I'm 50/50 on whether I want to merge this PR or not. Ultimately, we're fighting against Anthropic and putting forward a programatic fix to an issue we don't own. I also question how prevalent this issue really is for the wider community.

My preference would be to fully explore some prompt engineering techniques to coax the LLM in always giving correct output, perhaps asking it to self-reflect on its tool calls. Or, we use the new background strategy and create a structured output parser, sending the erroneous JSON to another low-cost LLM to fix.

This is also where I'll bring in my trusty thought matrix of effort versus impact. This PR has been a huge amount of effort to implement (and likely to maintain) versus an issue which has only happened to me once in all my time using Anthropic for tool calls (everyday for however many months). I appreciate this isn't the case for @jinzhongjia and there experience is different to mine.

jinzhongjia:
This problem has occurred several times, especially when LLM deems it necessary to make a large change all at once, such as modifying 200 lines of code.

jinzhongjia:
I now prefer LLM modifying the code gradually, rather than modifying large chunks of code all at once.

bassamsdata:
> large change all at once, such as modifying 200 lines of code.

Interesting information. The issue occurred for me this time while using Opus 4.5 with the Copilot adapter. I think we can simply remove the faulty JSON and everything should function correctly, though that's more of a defensive approach. I'll continue investigating this problem.

> I'm 50/50 on whether I want to merge this PR or not.

It's completely fine not to merge it, but I'll keep looking into the issue. I'll leave the PR open for tracking purposes, but I'll switch it to a draft.
I'll likely move the tool to TOON as your comment pointed it out and I hadn't known about it before, and it seems like an interesting alternative approach for me to try locally. I m also preparing a small PR to update some of the tool prompts, since I noticed a few incorrect patterns with certain LLMs now that I m running a fully autonomous agent with CodeCompanion.

olimorris:
Worth noting that Anthropic have strict-tool-use:  https://platform.claude.com/docs/en/build-with-claude/structured-outputs#strict-tool-use

bassamsdata:
Fine-tuning the prompt in v18 has made the issue obsolete, as it effectively restrains the LLM (primarily Claude) from aggressively modifying large files in a single pass.

PR #2391, opened 2025-11-16
Fix Keymap Handling for `list_code_usages` + Properly Attribute Async Tool Responses

bassamsdata:
### **Fix 1:**
Diff keymaps Not firing when using `list_code_usages` tool with `insert_edit_into_file` due to window and buffer manipulations.

Made the whole thing properly headless so keymaps actually work again. No more window jumping or buffer/cursor manipulation - everything runs off filepath/line/col now and loads files quietly in the background.

### **Fix 2:**
Async Tool Responses Going to the Wrong Tool
The issue was async tools stepping on each other. If accepted a diff at the wrong moment, the callback could end up tied to a different tool because `self.orchestrator.tool` had already changed.

So each Runner now grabs its own tool reference when it's created and sticks to it the whole way through. During success/error callbacks, we temporarily restore that tool so the response always lands on the right `tool_call_id`.
Makes the whole async flow way more predictable when multiple edits are running at once especially if we have an error.

Thank you

## Related Issue(s)

issue 1: Using the 2 tools together and force the LLM to use edit tool after the list code usages.

issue 2: the same but force the LLM to edit multiple files after the list_code_usages.

## Checklist

### DIFF
--- a/tests/strategies/chat/tools/catalog/test_list_code_usages.lua
+++ b/tests/strategies/chat/tools/catalog/test_list_code_usages.lua
@@ -74,7 +74,8 @@ local T = new_set({
       child.lua([[
         local LH = require("codecompanion.strategies.chat.tools.catalog.list_code_usages.lsp_handler")
 
-        LH.execute_request_async = function(_, method, cb)
+        -- mock signature for headless implementation: (filepath, line, col, method, callback)
+        LH.execute_request_async = function(filepath, line, col, method, cb)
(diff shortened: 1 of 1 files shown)


olimorris:
My advice would be not to spend too much time on `list_code_usages`. I've used it in the last month and I'm not a fan of the UX and how it behaves - to the point where I'm thinking of moving it out of the plugin. It doesn't "feel" CodeCompanion like and I observe that by choice most LLMs prefer `grep` anyway.

PR #2402, opened 2025-11-18
fix(chat buffer): Fix duplicated context when CodeCompanionChat Add with no opened chat

chanserv:
## Description

When I run a keymap to simulate `CodeCompanionChat Add` in visual mode and with no opened chat buffer, a new buffer is opened and the context (the visual selection) will get added twice to the newly opened buffer (screenshot 1)
This is mainly because the add function will open a new buffer and the routine that open a new buffer will add the context, after that the add function will resume and it will also append the context to the new buffer.

This fix will pass an empty context to the underlying method when opening a new buffer, so no context will be added. After that the context will be handled by add method
## Related Issue(s)

N/A

## Screenshots

Before

before

After

after

## Checklist


olimorris:
Thanks for the PR. Unfortunately this is causing the tests to fail. I think you forgot the brackets

PR #2472, opened 2025-11-30
fix(chat): Handle string model lists and add tests

Davidyz:
## Description

Handle the special case where `models` is a list of strings in `get_models_list`.

1. [`vim.iter` treats tables and lists differently](https://github.com/neovim/neovim/blob/8c1327a62245c0b5ecc0c5f3795911b75144d5a0/runtime/lua/vim/iter.lua#L1092). When the parameter is a list, the signature of `map` would be `fun(item)` instead of `fun(key, value)`. 
2. When `model.choices` is a list of strings (happened when I was using `openai_compatible` adapter), the existing code essentially convert everything to `nil` because [`value` here](https://github.com/olimorris/codecompanion.nvim/blob/8ad65eef735b31bb47d76f59d878ee1bac4bdc85/lua/codecompanion/strategies/chat/keymaps/change_adapter.lua#L99) would be `nil`.
3. This caused the model list in the `vim.ui.select` picker to only contain the default model.

## Related Issue(s)

`git bisect` pointed me to #2427 

## Screenshot

I have a few (~10) LLMs on this endpoint, but without this PR, I can only see one.
image

## Checklist

### DIFF
--- a/lua/codecompanion/interactions/chat/keymaps/change_adapter.lua
+++ b/lua/codecompanion/interactions/chat/keymaps/change_adapter.lua
@@ -92,6 +92,10 @@ function M.get_models_list(adapter)
   local models_list = vim
     .iter(models)
     :map(function(key, value)
+      if type(key) == "string" and value == nil then
+        -- `models` is already a list
+        return key
+      end
       if type(value) == "table" and not value.id then
         value.id = key
       end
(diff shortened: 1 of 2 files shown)


olimorris:
@Davidyz happy for this to be merged?

Davidyz:
Yep, I was just rebasing from `main` and verified that this still works with all the changes in v18. Curiously, my local `make test` failed (not just for this branch, but `main` too). The CI passed, and it's been a while since I did it last time, so it's probably something wrong/outdated with my local setup.

Issue #2525, opened 2025-12-15
[Bug]: `context.text` (and `shared.code`) do not properly escape selection


chrisgrieser:
### Pre-submission checklist

- [x] I have read the [documentation](https://codecompanion.olimorris.dev)
- [x] I have updated the plugin and all dependencies to the latest versions
- [x] I have searched for existing [issues](https://github.com/olimorris/codecompanion.nvim/issues) and [discussions](https://github.com/olimorris/codecompanion.nvim/discussions)
- [x] My issue is not a minor or cosmetic quirk (e.g., formatting, spacing, or other non-functional details)

### Neovim version (nvim -v)

NVIM v0.11.5

### Operating system/version

macOS 26.1 (arm)

### Adapter and model

openai, gpt-5-mini

### Describe the bug

`context.text` (and also when using the lua function [shared.code](https://github.com/olimorris/codecompanion.nvim/blob/8e813517c537007c784b2702609b244712433e1f/lua/codecompanion/actions/builtins/shared.lua), for that matter) do not properly escape the selection if the selection includes lua patterns like `%w` or `%s`.

Thus, the issue not only occurs with your own prompts, but also when using a builtin prompt like `explain`.

### Steps to reproduce

1. Write a line with a lua pattern, e.g. `local pattern = "%w"`
2. Select the line
3. Run `require("codecompanion").prompt("explain")`

### Expected behavior

the context code should not be altered.

### Screenshots or recordings (optional)

<img width="1007" height="315" alt="Image" src="https://github.com/user-attachments/assets/7a0dbb7f-2f02-4c9c-8a0f-3a01796b315c" />

### minimal.lua file

```lua
vim.env.LAZY_STDPATH = "/tmp/nvim-debug"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

local plugins = {
	{
		"olimorris/codecompanion.nvim",
		dependencies = "nvim-lua/plenary.nvim",
		opts = {
			interactions = {
				chat = { adapter = { name = "openai_responses", model = "gpt-5-mini" } },
			},
			opts = { log_level = "DEBUG" },
			adapters = {
				http = {
					openai_responses = function()
						return require("codecompanion.adapters").extend("openai_responses", {
							env = { api_key = "..." },
							schema = {
								model = {
									choices = {
										["gpt-5-mini"] = { opts = { can_reason = true } },
									},
								},
[...]

PR #2532, opened 2025-12-15
fix: pattern escape context

olimorris:
## Description

Ensure that any code that is received from the context object is pattern escaped.

## Related Issue(s)

#2525

## Checklist


chrisgrieser:
Thank you!

Looking at the fix, I think this issue might not be fully resolved, though. The fix only handles `${context.code}`, however if I am using lua functions in the prompt library, there are still problems:
- `context.code` is escaped while `context.lines` is not, which is somewhat inconsistent at the very least.
- Any result of a lua function used, e.g., the original `${shared.code}` still has the escaping issue.

olimorris:
If you're wanting to use `context.lines` then I would advise using Lua prompts over markdown or, handling the escaping in your custom functions.

Issue #1850, opened 2025-07-16
[Bug]: `Bad Request` after partial tool call failures with Github Copilot


cairijun:
### Pre-submission checklist

- [x] I have read the [documentation](https://codecompanion.olimorris.dev)
- [x] I have updated the plugin and all dependencies to the latest versions
- [x] I have searched for existing [issues](https://github.com/olimorris/codecompanion.nvim/issues) and [discussions](https://github.com/olimorris/codecompanion.nvim/discussions)
- [x] My issue is not a minor or cosmetic quirk (e.g., formatting, spacing, or other non-functional details)

### Neovim version (nvim -v)

v0.12.0-dev-768+gf576b59a09

### Operating system/version

Linux

### Adapter and model

Github Copilot with gpt-4.1

### Describe the bug

When Copilot tries to do multiple tool calls in one response but some of them fail, it may stuck in `Bad Request` errors for all subsequent requests.

My guess is that Copilot expects results of ALL tool calls, whether succeeded, failed or being skipped. But CodeCompanion would stop tool execution when any of them failed or rejected, and not generate results for the skipped tool calls.

### Steps to reproduce

1. `nvim --clean -u minimal.lua`
2. `:CodeCompanionChat`
3. Enter:
    ```
    @{files}
    Read the following files and make a summary.
    * `/tmp/xxx/sample/error1.txt`
    * `/tmp/xxx/sample/error2.txt`
    * `/tmp/xxx/sample/error3.txt`
    ```
4. Wait for the file read error.
5. Enter: `Ignore this error.`
6. Copilot responses `Bad Request`, and we are stuck here unless we manually fix the chat with the debug panel.
7. `gd`, remove the tool calls that were skipped.
8. The chat can continue now.

### Expected behavior

We should be able to continue this chat, either let the LLM ignore the errors or instruct it revise its tool call requests.

CodeCompanion should either keep running other tool calls, or generate dummy results for them to let the LLM know these tool calls were skipped.

### Screenshots or recordings (optional)

```lua
{ {
[...]

PR #2545, opened 2025-12-17
fix(tools): correct handling multi-tool in queue

bassamsdata:
## Description

This fixes an issue with multiple tools being called one after another. If one tool gets canceled or runs async, it could end up canceling the remaining tool calls in the queue. Since `error()` already calls `setup()`, which closes the failed tool and starts the next one, nothing else needs to be done.

## Related Issue(s)

**Reasoning:** There’s no need to call `close()` at this point, as it would end up closing the newly started tool rather than the one that encountered the error. The errored tool has already been properly closed during the setup invoked by `error()`.

to reproduce: we can share all the tools with LLM and ask it to test them all at once. 

## Checklist

### DIFF
--- a/lua/codecompanion/interactions/chat/tools/orchestrator.lua
+++ b/lua/codecompanion/interactions/chat/tools/orchestrator.lua
@@ -101,6 +101,7 @@ end
 ---@field tool CodeCompanion.Tools.Tool
 ---@field queue CodeCompanion.Tools.Orchestrator.Queue
 ---@field status string
+---@field current_tool_stdout table Output accumulator for the current tool only (reset per-tool)
 local Orchestrator = {}
 
 ---@param tools CodeCompanion.Tools


bassamsdata:
It seems that the issue is more complex than I initially thought. I'm encountering problems with user confirmation and the approval/rejection process when multiple tools are in the queue. For example, the 'insert_edit_into_file' action is being rejected, but it still appears as if it's been accepted in the chat buffer.

olimorris:
I've never encountered this before. What steps are you taking to see this?

cairijun:
It seems #1852 has never thoroughly resolved #1850 . Simply change the `read_file` to async can partially reproduce the issue:

```patch
diff --git a/lua/codecompanion/interactions/chat/tools/builtin/read_file.lua b/lua/codecompanion/interactions/chat/tools/builtin/read_file.lua
index e444d8390..57273a974 100644
--- a/lua/codecompanion/interactions/chat/tools/builtin/read_file.lua
+++ b/lua/codecompanion/interactions/chat/tools/builtin/read_file.lua
@@ -125,8 +125,10 @@ return {
     ---@param args table The arguments from the LLM's tool call
     ---@param input? any The output from the previous function call
     ---@return { status: "success"|"error", data: string }
-    function(self, args, input)
-      return read(args)
+    function(self, args, input, output_handler)
+      vim.schedule(function()
+        output_handler(read(args))
+      end)
     end,
   },
   schema = {
```

With async cmds, all 3 tools are executed, but only the first call's output handler is called:


With sync cmds, all three errors are reported:
image

cairijun:
I also found another issue with multi-tool invocations.

The `Orchestrator` accumulate all tools' output in `self.tools.stdout` and simply pass it to each tool's `output.success`:

https://github.com/olimorris/codecompanion.nvim/blob/dfc85fe9299c36a6054ca2318ca0e9ee48f5446b/lua/codecompanion/interactions/chat/tools/orchestrator.lua#L385-L390

https://github.com/olimorris/codecompanion.nvim/blob/dfc85fe9299c36a6054ca2318ca0e9ee48f5446b/lua/codecompanion/interactions/chat/tools/orchestrator.lua#L208-L218

So tools invoked in the later receive ALL the previous output in their `success` handlers, and most of them simply join the `stdout` into long lines:
https://github.com/olimorris/codecompanion.nvim/blob/dfc85fe9299c36a6054ca2318ca0e9ee48f5446b/lua/codecompanion/interactions/chat/tools/builtin/read_file.lua#L183-L190

This causes `[read_file(a), read_file(b), read_file(c)]` generates an output sequence like:
* `content_of_a`
* `content_of_a` `\n` `content_of_b`
* `content_of_a` `\n` `content_of_b` `\n` `content_of_c`

olimorris:
@cairijun that's an interesting discovery, thanks for sharing. I'll start work on a PR for that.

bassamsdata:
@olimorris, to make these errors easier to reproduce, I found a straightforward approach using Copilot (GPT-4.1). Simply navigate to the `codecompanion` directory and use the following prompt (please adjust the file path as needed).

```markdown
hello

@{files}
@{cmd_runner}

please test all tools all together since I'm doing a testing for those tools.
for insert_edit_into_file you can add small comment to this file
~/repos/codecompanion.nvim/lua/codecompanion/providers/diff/inline.lua
set oldTEXT to "^" so no need to know file conents.
for some tools i need them in this order:
insert_edit_into_file -> cmd_runner -> create_file -> delete_file
but test them all, all togther in one go please.
NO NEED FOR PLAN. JUST DO IT
```

The LLM will generate a plan regardless; however, after responds with “yes,” several issues begin to surface.

1. While waiting for the diff from `insert_edit_into_file`, all subsequent tools in the queue are canceled (appearing as if the user canceled the tool calls). This seems to occur because the async tool sends a `nil` response to the orchestrator while it is still waiting.
2. There are also multiple issues related to the confirmation prompt that become apparent during this process.
3. Additionally, when attempting to reject certain async tools such as `cmd_runner` or `insert_edit_into_file`, the tool is rejected correctly even though it appears as a successful execution in the chat buffer.

bassamsdata:
One more thing I wanted to mention, I’m not entirely sure about this part, so I might be missing something. My understanding is that if a tool is rejected, the other tools in the queue should still be allowed to run. Is that correct?

```lua
-- NOTE: Cancel current tool, then cancel all queued tools
self.output.cancelled(cmd)
self:close()
self:cancel_pending_tools()
```

I’ve made a few new changes in commit [859f9a6](https://github.com/olimorris/codecompanion.nvim/pull/2545/commits/859f9a6059c938a0a6645d1f5d8a35cab1509d3d) and added some comments to make the behavior clearer to my intention.

Thank you

olimorris:
> My understanding is that if a tool is rejected, the other tools in the queue should still be allowed to run. Is that correct?

Yes, if a tool is rejected, then that tool is not executed and we move on to the next queue. However, if the user opts to cancel then they're cancelling every tool in the queue.

bassamsdata:
The current state of this PR addresses all of these issues, although I’m not yet certain that the solutions are the optimal ones.

edit: this PR is just me trying to fix the issues. it's draft and for tacking only.

olimorris:
Added test coverage for @cairijun's point. When we merge this I'm going to do some light refactoring on the orchestrator. @bassamsdata let me know when you're happy

bassamsdata:
I’ve tested it across several scenarios and I’m happy with the current behavior. At this point, I’ll leave it up to you if  you’d like to change or adjust anything.

olimorris:
This looks great. I'll merge in the morning

PR #2548, opened 2025-12-18
feat: add Gemini 3 Flash

Davidyz:
## Description

Add [Gemini 3 Flash](https://blog.google/products/gemini/gemini-3-flash/) to the gemini adapter.

## Related Issue(s)

Possible followup: do we want to use the [Gemini API](https://ai.google.dev/api/models) for dynamically fetching and parsing available models (similar to copilot and ollama)?
```bash
curl "https://generativelanguage.googleapis.com/v1beta/models?key=$GEMINI_API_KEY" \
-H "Content-Type: application/json"
```
This returns all kinds of models (including image generations, embedding, etc.), but we can filter them by checking for `"generateContent"` in the `"supportedGenerationMethods"` field.

The response also seems to contain a `thinking` key for each model, so we can actually generate accurate capabilities from the fetched data. The only caveat is that this API also returns some gemma-3 models, which may or may not be too verbose.

## Checklist

### DIFF
--- a/lua/codecompanion/adapters/http/gemini.lua
+++ b/lua/codecompanion/adapters/http/gemini.lua
@@ -126,6 +126,10 @@ return {
       default = "gemini-3-pro-preview",
       choices = {
         ["gemini-3-pro-preview"] = { formatted_name = "Gemini 3 Pro", opts = { can_reason = true, has_vision = true } },
+        ["gemini-3-flash-preview"] = {
+          formatted_name = "Gemini 3 Flash",
+          opts = { can_reason = true, has_vision = true },
+        },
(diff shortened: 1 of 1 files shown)


olimorris:
> Possible followup: do we want to use the [Gemini API](https://ai.google.dev/api/models) for dynamically fetching and parsing available models (similar to copilot and ollama)?

As with OpenAI and Anthropic, I'm going to keep the models static for now. There have been numerous cases where new models require modifications to the adapter. `gemini-3` being the most recent one with its thought signatures and of course OpenAI with the new `responses` endpoint.

I'm hoping 2026 will bring a new protocol to make this less of an issue...

PR #2549, opened 2025-12-18
feat: add MCP client

cairijun:
## Description

As discussed in https://github.com/olimorris/codecompanion.nvim/discussions/2506, we want a basic MCP client implemented in CodeCompanion.

Core design:
* `MCP.Client`s are started on first chat creation.
* Tools from MCP servers are loaded as CodeCompanion tools.
* Tools from the same MCP server are grouped as a tool group.
* User can use `@{mcpServerName}` to grant tool access to the LLM.

Features:

| Category     | Feature                                | Supported     | Notes                                                                 |
|--------------|----------------------------------------|-------------|-----------------------------------------------------------------------|
| Transport    | Stdio                        | Yes   |                                                                       |
| Transport    | Streamable HTTP              | No | No plan; can use `mcp-remote`                                         |
| Basic    | Cancellation                           | Partial     | Only triggered on timeout                                             |
| Basic    | Progress                               | No | Possible future improvement                                           |
| Basic    | Task                                   | No | No plan; too complex and marked `experimental` in the Spec           |
| Client   | Roots                                  | Yes   | Disabled by default                                                    |
| Client   | Sampling                               | No | No plan; needs concrete examples to justify                           |
| Client   | Elicitation                            | No | No plan; needs concrete examples                                      |
| Server   | Tools                                  | Yes   | Currently only supports Text Content                                  |
| Server   | Tool list changed notification         | No | No plan; needs concrete examples to demonstrate necessarity  |
| Server   | Prompts                                | No | Planned; needs concrete examples to help design UX                     |
| Server   | Resources                              | No | Planned                                                                |
| Server   | Completion                             | No | No plan; depends on Prompts UX                                         |
| Server   | Logging                                | Partial     | Currently implemented by forwarding JSON-RPC messages to `log:info`   |
| Server   | Pagination                             | Yes   |                                                                       |

Config example:
```lua
local config = { ... }
config.mcp.servers = {
  -- a simple case
  tavilyMcp = {
    cmd = { "npx", "-y", "tavily-mcp@latest" },
    env = {
      TAVILY_API_KEY = "cmd:cat ~/.config/tavily_api_key",
    },
  },
  -- workspace roots reporting
  filesystem = {
    cmd = { "npx", "-y", "@modelcontextprotocol/server-filesystem" },
    roots = function()
      local roots =
        { { name = "Current Working Directory", uri = vim.uri_from_fname(vim.fn.getcwd()) } }
      vim
        .iter(vim.lsp.get_clients())
        :map(function(client)
          return client.workspace_folders
        end)
        :flatten()
        :each(function(folder)
          table.insert(roots, { name = "LSP Workspace", uri = folder.uri })
        end)
      return roots
    end,
    register_roots_list_changed = function(notify)
      vim.api.nvim_create_autocmd("LspAttach", {
        callback = function()
          notify()
        end,
      })
    end,
  },
  -- override output handlers
  sequentialThinking = {
    cmd = { "npx", "-y", "@modelcontextprotocol/server-sequential-thinking" },
    tool_overrides = {
      sequentialthinking = {
        output = {
          success = function(self, tools, cmd, stdout)
            local output = stdout and stdout[#stdout]
            local msg = "Sequential thinking: " .. self.args.thought
            tools.chat:add_tool_output(self, output, msg)
          end,
        },
      },
    },
  },
  -- a more complex example
  runPython = {
    cmd = { "uvx", "mcp-run-python", "--deps", "numpy,pandas", "stdio" },
    -- override server-level prompt
    server_instructions = function(orig)
      return orig
        .. "\nThe Python environment has the following packages pre-installed: numpy, pandas."
        .. " You can only use these packages and standard library modules."
    end,
    -- set default tool opts
    default_tool_opts = {
      require_approval_before = true,
    },
    tool_overrides = {
      run_python_code = {
        timeout_ms = 5 * 60 * 1000, -- 5 minutes
        output = {
          prompt = function(self)
            local args = self.args
            return string.format(
              "Confirm to execute the following Python code:\n```python\n%s\n```\nGlobal Variables:\n%s",
              args.python_code,
              vim.inspect(args.global_variables)
            )
          end,
        },
      },
    },
  },
}

require("codecompanion").setup(config)
```

## Related Issue(s)

https://github.com/olimorris/codecompanion.nvim/discussions/2506

## Screenshots

image
image

## Checklist

### DIFF
--- a/doc/codecompanion.txt
+++ b/doc/codecompanion.txt
@@ -1,4 +1,4 @@
-*codecompanion.txt*         For NVIM v0.11        Last change: 2026 January 31
+*codecompanion.txt*        For NVIM v0.11        Last change: 2026 February 01
 
 ==============================================================================
 Table of Contents                            *codecompanion-table-of-contents*
@@ -34,6 +34,7 @@ Table of Contents                            *codecompanion-table-of-contents*
   - Current Limitations                |codecompanion-acp-current-limitations|
   - See Also                                      |codecompanion-acp-see-also|
 6. MCP                                                     |codecompanion-mcp|
+  - Usage                                            |codecompanion-mcp-usage|
   - Implementation                          |codecompanion-mcp-implementation|
   - Protocol Version                      |codecompanion-mcp-protocol-version|
   - See Also                                      |codecompanion-mcp-see-also|
@@ -837,6 +838,18 @@ CodeCompanion implements the Model Context Protocol (MCP)
 external systems and applications. The plugin only implements a subset of the
 full MCP specification, focusing on the features that enable developers to
 enhance their coding experience.
+
+
+USAGE                                                *codecompanion-mcp-usage*
+
+To use MCP servers within CodeCompanion, refer to the
+|codecompanion-usage-chat-buffer-tools-mcp| section in the chat buffer usage
+section of the documentation.
+
+If |codecompanion-configuration-mcp-enabling-servers|, the servers will be
+started when you open a chat buffer for the first time. However, you can use
+the |codecompanion-usage-chat-buffer-slash-commands-mcp| to start or stop
+servers manually.
 
 
 IMPLEMENTATION                              *codecompanion-mcp-implementation*
@@ -2114,7 +2127,7 @@ applications to external systems.
 
 You can find out which parts of the protocol CodeCompanion has implemented on
 the |codecompanion-model-context-protocol| page. Currently, you can leverage
-MCP servers with |codecompanion-usage-chat-buffer|.
+MCP servers with |codecompanion-usage-chat-buffer-index|.
 
 
 MCP SERVERS ~
@@ -2136,6 +2149,10 @@ to fetch the value from any source you like.
 
 ROOTS
 
+
+  [!IMPORTANT] The `roots` feature is a hint to MCP servers. Compliant servers
+  use it to limit file system access, but CodeCompanion cannot enforce this. For
+  untrusted servers, use isolation mechanisms like containers.
 Roots <https://modelcontextprotocol.io/specification/2025-11-25/client/roots>
 allow you to specify directories that the MCP server can access. By default,
 roots are disabled for security reasons. You can enable them by adding a
@@ -2166,7 +2183,34 @@ The `tool_overrides` field is a table where keys are the **MCP tool names**
 
 
 
-AVAILABLE OVERRIDE OPTIONS
+TOOL DEFAULTS
+
+You can set default options for all tools by setting the `tool_defaults`
+option. However, note that `tool_overrides` take precedence over them:
+
+>lua
+    require("codecompanion").setup({
+      mcp = {
+        ["math-server"] = {
+          cmd = { "npx", "-y", "math-mcp-server" },
+          tool_defaults = {
+            require_approval_before = true,
+          },
+          -- Per-tool overrides take precedence over tool_defaults
+          tool_overrides = {
+            add = {
+              opts = {
+                require_approval_before = false,
+              },
+            },
+          },
+        },
+      },
+    })
+<
+
+
+OVERRIDE OPTIONS
 
 Each tool override can include:
 
@@ -2187,36 +2231,8 @@ Each tool override can include:
   timeout               number          Custom timeout in milliseconds for
                                         this tool
 
-  enabled               boolean or      Whether the tool is enabled
-                        function        
+  enabled               boolean         Whether the tool is enabled
   ------------------------------------------------------------------------
-
-DEFAULT TOOL OPTIONS
-
-If you want to apply the same options to all tools from a server, use
-`tool_defaults`:
-
->lua
-    require("codecompanion").setup({
-      mcp = {
-        ["math-server"] = {
-          cmd = { "npx", "-y", "math-mcp-server" },
-          tool_defaults = {
-            require_approval_before = true,
-          },
-          -- Per-tool overrides take precedence over tool_defaults
-          tool_overrides = {
-            add = {
-              opts = {
-                require_approval_before = false,  -- Override the default for this tool
-              },
-            },
-          },
-        },
-      },
-    })
-<
-
 
 RULES                                      *codecompanion-configuration-rules*
 
@@ -3902,6 +3918,14 @@ In the `openai_responses` adapter, the following tools are available:
 - `web_search` - Allow models to search the web for the latest information before generating a response.
 
 
+MCP ~
+
+The MCP servers you’ve |codecompanion-configuration-mcp| in CodeCompanion
+expose their own set of tools that you can use in the chat buffer. Once a
+server has been started, the tools will be available to you and appear in the
+completion menu, by typing `@`. They are prefixed with `mcp:`.
+
+
 SECURITY ~
 
 CodeCompanion takes security very seriously, especially in a world of agentic
@@ -4081,6 +4105,15 @@ image picker is only available with `snacks.nvim` and the `vim.ui.select`.
 
 The `rules` slash command allows you to add
 |codecompanion-usage-chat-buffer-rules| groups to the chat buffer.
+
+
+/MCP ~
+
+The `mcp` slash command allows you to start and stop
+|codecompanion-configuration-mcp| servers manually from within a chat buffer.
+This is applied at a global level, so starting/stopping servers in one chat
+buffer will affect all other chat buffers. A `snacks.nvim` and `vim.ui.select`
+provider is available for selecting which MCP servers to start/stop.
 
 
 /MODE ~
--- a/doc/configuration/mcp.md
+++ b/doc/configuration/mcp.md
@@ -6,7 +6,7 @@ description: Learn how to configure MCP servers within CodeCompanion.nvim
 
 In [#2549](https://github.com/olimorris/codecompanion.nvim/pull/2549), CodeCompanion added support for the [Model Context Protocol (MCP)](https://modelcontextprotocol.io), an open-source standard for connecting AI applications to external systems.
 
-You can find out which parts of the protocol CodeCompanion has implemented on the [MCP](/model-context-protocol) page. Currently, you can leverage MCP servers with [chat interactions](/usage/chat-buffer).
+You can find out which parts of the protocol CodeCompanion has implemented on the [MCP](/model-context-protocol) page. Currently, you can leverage MCP servers with [chat interactions](/usage/chat-buffer/index).
 
 ## Configuring MCP Servers
 
@@ -26,7 +26,7 @@ require("codecompanion").setup({
 })
 ```
 
-```lua [Env Vars] {5-7}
+```lua [Environment Variables] {5-7}
 require("codecompanion").setup({
   mcp = {
     ["tavily-mcp"] = {
@@ -46,7 +46,7 @@ In the example above, we're using [1Password CLI](https://developer.1password.co
 ### Roots
 
 > [!IMPORTANT]
-> The `roots` feature is only a **hint** to MCP servers. A **compliant** server can use this information to prevent unexpected file system access from the LLM. However, CodeCompanion cannot enforce this upon the servers. If you are using uncompliant or even untrusted MCP servers (and you shouldn't), consider using a real isolation mechanism such as containers.
+> The `roots` feature is a hint to MCP servers. Compliant servers use it to limit file system access, but CodeCompanion cannot enforce this. For untrusted servers, use isolation mechanisms like containers.
 
 [Roots](https://modelcontextprotocol.io/specification/2025-11-25/client/roots) allow you to specify directories that the MCP server can access. By default, roots are disabled for security reasons. You can enable them by adding a `roots` field to your server configuration:
 
@@ -88,7 +88,7 @@ By default, all MCP servers are enabled in CodeCompanion. This results in server
 
 ::: code-group
 
-```lua [Global] {3}
+```lua [Globally] {3}
 require("codecompanion").setup({
   mcp = {
     enabled = false,
@@ -99,7 +99,7 @@ require("codecompanion").setup({
 })
 ```
 
-```lua [Individual Server] {3,6-8}
+```lua [Per Server] {3,6-8}
 require("codecompanion").setup({
   mcp = {
     enabled = true,
@@ -180,21 +180,9 @@ require("codecompanion").setup({
 
 :::
 
-### Available Override Options
+### Tool Defaults
 
-Each tool override can include:
-
-| Option | Type | Description |
-|--------|------|-------------|
-| `opts` | `table` | Tool options like `require_approval_before`, `require_approval_after` |
-| `output` | `table` | Custom output handlers (`success`, `error`, `prompt`, `rejected`, `cancelled`) |
-| `system_prompt` | `string` | Additional system prompt text for this tool |
-| `timeout` | `number` | Custom timeout in milliseconds for this tool |
-| `enabled` | `boolean` or `function` | Whether the tool is enabled |
-
-### Default Tool Options
-
-If you want to apply the same options to all tools from a server, use `tool_defaults`:
+You can set default options for all tools by setting the `tool_defaults` option. However, note that `tool_overrides` take precedence over them:
 
 ```lua
 require("codecompanion").setup({
@@ -208,7 +196,7 @@ require("codecompanion").setup({
       tool_overrides = {
         add = {
           opts = {
-            require_approval_before = false,  -- Override the default for this tool
+            require_approval_before = false,
           },
         },
       },
@@ -216,3 +204,18 @@ require("codecompanion").setup({
   },
 })
 ```
+
+
+### Override Options
+
+Each tool override can include:
+
+| Option | Type | Description |
+|--------|------|-------------|
+| `opts` | `table` | Tool options like `require_approval_before`, `require_approval_after` |
+| `output` | `table` | Custom output handlers (`success`, `error`, `prompt`, `rejected`, `cancelled`) |
+| `system_prompt` | `string` | Additional system prompt text for this tool |
+| `timeout` | `number` | Custom timeout in milliseconds for this tool |
+| `enabled` | `boolean`  | Whether the tool is enabled |
+
+
--- a/doc/model-context-protocol.md
+++ b/doc/model-context-protocol.md
@@ -6,6 +6,12 @@ description: How to leverage Model Context Protocol (MCP) servers within CodeCom
 
 CodeCompanion implements the [Model Context Protocol (MCP)](https://modelcontextprotocol.io) to enable you to connect the plugin to external systems and applications. The plugin only implements a subset of the full MCP specification, focusing on the features that enable developers to enhance their coding experience.
 
+## Usage
+
+To use MCP servers within CodeCompanion, refer to the [tools](/usage/chat-buffer/tools#mcp) section in the chat buffer usage section of the documentation.
+
+If [enabled](/configuration/mcp#enabling-servers), the servers will be started when you open a chat buffer for the first time. However, you can use the [MCP slash command](/usage/chat-buffer/slash-commands#mcp) to start or stop servers manually.
+
 ## Implementation
 
 
--- a/doc/usage/chat-buffer/slash-commands.md
+++ b/doc/usage/chat-buffer/slash-commands.md
@@ -63,6 +63,10 @@ The _image_ slash command allows you to add images into a chat buffer via remote
 
 The _rules_ slash command allows you to add [rules](/usage/chat-buffer/rules) groups to the chat buffer.
 
+## /mcp
+
+The _mcp_ slash command allows you to start and stop [Model Context Protocol (MCP)](/configuration/mcp) servers manually from within a chat buffer. This is applied at a global level, so starting/stopping servers in one chat buffer will affect all other chat buffers. A _snacks.nvim_ and `vim.ui.select` provider is available for selecting which MCP servers to start/stop.
+
 ## /mode
 
 The _mode_ slash command is specific to [ACP](/configuration/adapters-acp) adapters and allows users to switch between different agent operating modes, as per the [protocol](https://agentclientprotocol.com/protocol/session-modes) docs.
--- a/doc/usage/chat-buffer/tools.md
+++ b/doc/usage/chat-buffer/tools.md
@@ -309,6 +309,10 @@ In the `openai_responses` adapter, the following tools are available:
 
 - `web_search` - Allow models to search the web for the latest information before generating a response.
 
+## MCP
+
+The MCP servers you've [configured](/configuration/mcp) in CodeCompanion expose their own set of tools that you can use in the chat buffer. Once a server has been started, the tools will be available to you and appear in the completion menu, by typing `@`. They are prefixed with `mcp:`.
+
 ## Security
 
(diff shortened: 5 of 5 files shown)


olimorris:
Fantastic, thank you @cairijun. The holidays start from Monday for me, so I will get around to properly testing this out, then. But this will be priority 1 for me.

lua/codecompanion/mcp/init.lua
```diff
@@ -0,0 +1,46 @@
...
+---@class CodeCompanion.MCPConfig
+---@field servers table<string, CodeCompanion.MCP.ServerConfig>?
+
+---@type table<string, CodeCompanion.MCP.Client>
+local clients = {}
+
+---Start all configured MCP servers if not already started
+function M.start_all_if_not_started()
+  local mcp_cfg = require("codecompanion.config").mcp
+  if not mcp_cfg then
```
olimorris: We can remove this conditional because it can never occur based on how you've set out the config.

lua/codecompanion/mcp/tool_bridge.lua
```diff
@@ -0,0 +1,162 @@
...
+  end
+
+  if #installed_tools == 0 then
+    log:info("[MCP.%s] has no valid tools to configure", client.name)
+    return {}
+  end
+
+  local server_prompts = {
+    string.format(
+      "I'm giving you access to tools from MCP server '%s': %s.",
```
olimorris: This should be a constant at the top of the file so it's easy to find and edit.

```lua
local constants = {
  PROMPT = "I'm giving you ...."
}

-- constants.PROMPT
```

lua/codecompanion/interactions/chat/init.lua
```diff
@@ -525,6 +525,8 @@ function Chat.new(args)
 
   self:update_metadata()
 
+  require("codecompanion.mcp").start_all_if_not_started()
```
olimorris: `start_all_if_not_started` is verbose and repetitive. I think `start` will suffice. The chat class shouldn't care whether it's started or not. Let the conditional in the function determine if it needs to be started or not

lua/codecompanion/mcp/tool_bridge.lua
```diff
@@ -0,0 +1,162 @@
+local log = require("codecompanion.utils.log")
+
+local M = {}
+
+---Default tool output callbacks
+local DefaultOutputCallbacks = {}
```
olimorris: This should be snake case.

Also, using `Default` implies that there is an alternative for these callbacks. Something other than the defaults`.

lua/codecompanion/mcp/tool_bridge.lua
```diff
@@ -0,0 +1,162 @@
...
+
+function DefaultOutputCallbacks.success(self, tools, cmd, stdout)
```
olimorris: Throughout CodeCompanion, I use the following pattern.

```lua
local output = {
  success = function(self, tools, cmd, stdout)
    -- something
  end,
}
```

I propose doing the same for consistency. It also reads better as `output.success` rather than `DefaultOutputCallbacks.success`

lua/codecompanion/mcp/tool_bridge.lua
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  return prefixed_name, tool_cfg
+end
+
+---Install tools from an MCP server into CodeCompanion
+---@param client CodeCompanion.MCP.Client
+---@param mcp_tools MCP.Tool[]
+---@return string[] tools
+function M.install_tools(client, mcp_tools)
```
olimorris: We're not installing tools so I'd be minded to rename this to `share_tools` or something along those lines.

lua/codecompanion/mcp/tool_bridge.lua
```diff
@@ -0,0 +1,162 @@
...
+  local server_prompts = {
+    string.format(
+      "I'm giving you access to tools from MCP server '%s': %s.",
+      client.name,
+      table.concat(installed_tools, ", ")
+    ),
+  }
+
+  local cfg_server_instr = client.cfg.server_instructions
+  local final_server_instructions
```
olimorris: The whole section from L130 to L160 reads very verbosely to me.

The ultimate aim is to send prompts to the MCP server. The journey to get there is complicated. We have `server_prompts` to start off with. We then have `client.cfg.server_instructions`. And then `final_server_instructions`. Then these are merged into `server_prompts` before being concatenated into the chat.

Surely we can just have `prompts` as a table and we insert into that table depending on whether `client.cfg.server_instructions` is a function or a string etc? The more variables we introduce (like `final_server_instructions`) the more complicated it reads.

lua/codecompanion/config.lua
```diff
@@ -997,6 +997,10 @@ The user is working on a %s machine. Please respond with system specific command
       warning = " ",
     },
   },
+  -- MCP SERVERS -----------------------------------------------------------
+  mcp = {
```
olimorris: This should sit in `interactions.chat`.

lua/codecompanion/mcp/client.lua
```diff
@@ -0,0 +1,492 @@
...
+    connection = self._conn_factory(name, cfg),
+    resp_handlers = {},
+    server_request_handlers = {
+      ["ping"] = self._handle_server_ping,
+      ["roots/list"] = self._handler_server_roots_list,
+    },
+  }, self)
+end
+
+function Client:start_if_not_started()
```
olimorris: Add LuaLS annotations and types to these methods. Makes debugging and reading so much easier

lua/codecompanion/mcp/client.lua
```diff
@@ -0,0 +1,492 @@
...
+    self:_on_conn_line_read(line)
+  end, function(err)
+    self:_on_conn_close(err)
+  end)
+  utils.fire("MCPServerStart", { name = self.name })
+
+  self:_start_init_mcp()
+end
+
+function Client:_start_init_mcp()
```
olimorris: `start_init_mcp` feels repetitive. What's the difference between start and init?

lua/codecompanion/mcp/client.lua
```diff
@@ -0,0 +1,492 @@
...
+  if not self.server_capabilities.tools then
+    log:warn("[MCP.%s] does not support tools", self.name)
+    return
+  end
+
+  local all_tools = {} ---@type MCP.Tool[]
+  local function load_tools(cursor)
+    self:request("tools/list", { cursor = cursor }, function(resp)
+      if resp.error then
+        log:error("[MCP.%s] tools/list request failed: %s", self.name, vim.inspect(resp))
```
olimorris: FYI you don't `vim.inspect`. This is a common mistake made by LLMs when traversing the CodeCompanion codebase. You can just use `resp` and the log class will expand any tables for you.

lua/codecompanion/mcp/init.lua
olimorris, line 1: What's the logic around having `client.lua` separate from `init.lua`? For consistency, would be nice to follow what we've done with ACP and just have `init.lua`

lua/codecompanion/mcp/client.lua
```diff
@@ -0,0 +1,492 @@
...
+    log:warn("[MCP.%s] exited with error: %s", self.name, err)
+  end
+  utils.fire("MCPServerExit", { name = self.name, err = err })
+end
+
+function Client:_on_conn_line_read(line)
+  if not line or line == "" then
+    return
+  end
+  local ok, msg = pcall(vim.json.decode, line, { luanil = { object = true } })
```
olimorris: In `acp/init.lua`, you'll see I have defined:

```lua
Connection.static.methods = {
  decode = { default = vim.json.decode },
  encode = { default = vim.json.encode },
  job = { default = vim.system },
  schedule = { default = vim.schedule },
  schedule_wrap = { default = vim.schedule_wrap },
}
```

Can you do the same in here? Makes mocking that much easier when it comes to testing and is consistent with the ACP and HTTP clients.

lua/codecompanion/mcp/client.lua
```diff
@@ -0,0 +1,492 @@
...
+---Default Connection implementation backed by vim.system
+---@class CodeCompanion.MCP.StdioConnection : CodeCompanion.MCP.Connection
+---@field name string
+---@field cmd string[]
+---@field env table?
+---@field _proc vim.SystemObj?
+---@field _last_tail string?
+---@field _on_line_read fun(line: string)?
+---@field _on_close fun(err: string?)?
+local StdioConnection = {}
```
olimorris: Should we call this `StdioTransport`? Feel like it ties in better with the [docs](https://modelcontextprotocol.io/specification/2025-11-25/basic/transports)

lua/codecompanion/mcp/tool_bridge.lua
cairijun: I think we have a cohesive logic of building a prompt here. Extracing a string template to a constant doesn't add too much value IMO.

Or should we extract this whole block of code to a `build_server_prompt(tools)` method?

lua/codecompanion/mcp/tool_bridge.lua
cairijun: We indeed have alternatives. The user can override some of the callbacks via `tool_overrides.{TOOL_NAME}.output` config table. I changed it to `local default_output = { ... }`.

lua/codecompanion/mcp/tool_bridge.lua
cairijun: I think `setup_tools` reads more clear to me. What do you think?

lua/codecompanion/mcp/tool_bridge.lua
cairijun: `server_instructions` might be `nil` (neither the server nor the user provides one) and we don't want a `Detailed instructions ...` heading without body. So an extra variable seems inevitable.

I moved the config overriding logic to `Client:get_server_instructions()` to make this piece of code cleaner.

lua/codecompanion/mcp/client.lua
cairijun: `start` implies this is asynchronous and `init` refers to the MCP initialization procedure.

Renamed to `_start_initialization` to avoid confusion.

lua/codecompanion/mcp/init.lua
cairijun: We may have some management APIs that do not directly tie to a specific client, e.g. `RefreshMCPTools`. The difference is that we generally have only one ACP instance active, but multiple MCP servers. So I prefer making `client.lua` focus on the logic of a single MCP client, while let the other (potentially public to the users) APIs go to `init.lua`.

lua/codecompanion/mcp/client.lua
cairijun: Sure. But there's a difference that the method overriding table must be an extra optional argument to the constructors as it shouldn't be exposed to the user config.

The transport factory function was also move to `methods` for consistence.

cairijun:
Hi @olimorris , thanks for your detailed and constructive comments! I've been a bit busy on some personal matters recently so I'm replying slower than usual.

I've push some commits addressing most of the comments.

The generated tool groups are now named `mcp.{SERVER_NAME}`, while the tools are named `mcp_{SERVER_NAME}_{ORIG_TOOL_NAME}` without the use of `.`, as it is not supported by many LLM providers. This inconsistency is not visible to the users as they always use the groups and the tools are actually hidden.

There's also a major change that I moved the output formatting logic from `cmds` to `success`/`error`. Now the user provided output handlers receive the original MCP tool result content objects. It makes the handlers a bit more verbose for simple cases, but must more flexible for complex tool results. I think this is acceptable since you normally don't need to override them for simple cases.

olimorris:
No worries @cairijun and no rush. I too have been in the same boat. We'll definitely get this PR into CodeCompanion 👍🏼. I will be picking this back up next week too.

olimorris:
I've got some time planned for this weekend to take a look at this. Looking forward to it.

lua/codecompanion/mcp/client.lua
olimorris, line 1: There's a couple of LSP warnings in this file that are worth addressing.

lua/codecompanion/mcp/tool_bridge.lua
```diff
@@ -0,0 +1,194 @@
+local log = require("codecompanion.utils.log")
+
+local CONSTANTS = {
```
olimorris, line 3: I've extracted some messaging to the top of the file as I'm likely to tweak this over time. Also, preference is to have `@{mcp:tavily}` rather than `@{mcp.tavily}` as I think it visually clarifies tavily is part of the MCP group (super minor nitpick).

lua/codecompanion/mcp/tool_bridge.lua
```diff
@@ -0,0 +1,194 @@
...
+  }
+
+  -- The prompt should also contain instructions from the server, if any.
+  local server_instructions = client:get_server_instructions()
+  if server_instructions and server_instructions ~= "" then
+    table.insert(server_prompt, "Detailed instructions for this MCP server:")
+    table.insert(server_prompt, server_instructions)
+  end
+
+  chat_tools.groups[fmt("%s%s", CONSTANTS.TOOL_PREFIX, client.name)] = {
```
olimorris: I've noticed that MCP tools appearing in the completion menu in the chat buffer to be temperamental. I'm assuming there is a timing difference between when these are resolved in the chat buffer and when the `providers.completion.init` file resolves them. I do think that logic should be moved to the latter file

olimorris (changes requested):
Initial reaction:



This is awesome. Really, really awesome. The fact all I needed to do to get this working, was:

```lua
      mcp = {
        servers = {
          ["tavily-mcp"] = {
            cmd = { "npx", "-y", "tavily-mcp@latest" },
            env = {
              TAVILY_API_KEY = "cmd:op read op://personal/Tavily_API/credential --no-newline",
            },
          },
        },
      },
```

...is tremendous.

I've put some inline comments. Pretty much minor stuff now. I will write the docs and add some more test coverage (think we might need a bit of coverage around mcp servers appearing in the completion list).

There's a couple of (hopefully) small hurdles I've noticed:

- In my example above, I'm being prompted by 1Password everytime I create a new chat buffer. I may not want to add the Tavily MCP server to this chat buffer so think this prompt should be when I've pressed submit to send the response to the LLM.
- It can be hit-and-miss if the MCP server appears in the completion list in the chat buffer. I notice when I open a fresh Neovim, create a chat buffer, I often don't see the MCP servers. If I open a second chat buffer, I do see them. I think we may need to shift the logic of creating the MCP groups to the `providers.completion.init` file like we do for all other tools and groups.

Otherwise, this is amazing and I can't wait to merge this into CodeCompanion. No rush and take your time. We all have lives outside of open-source. And, if you need me to pickup any bits, just say.

cairijun:
>* In my example above, I'm being prompted by 1Password everytime I create a new chat buffer. I may not want to add the Tavily MCP server to this chat buffer so think this prompt should be when I've pressed submit to send the response to the LLM.
>* It can be hit-and-miss if the MCP server appears in the completion list in the chat buffer. I notice when I open a fresh Neovim, create a chat buffer, I often don't see the MCP servers. If I open a second chat buffer, I do see them. I think we may need to shift the logic of creating the MCP groups to the providers.completion.init file like we do for all other tools and groups.

They might come from a same root cause --- MCP servers are started **asynchronously** on chat creation.

You're prompted by 1Pass because `mcp.start_servers()` is triggerred on `Chat.new()`. But that should only happen on the **first** chat buffer, even if the server is extremely slow to start --- unless it **exited** before you start another chat. I do not use 1Pass and didn't realise their could be a user interaction involved.

And because the servers are started (and loaded) **asynchronously**, they will be missing from the completion list until actually available. We can add a stub tool group to the config then resolve it once the server is ready, but that may cause missing tools if the chat is submitted before the tool group is resolved. Making the procedure synchronous also seems inappropriate to me, as `npx` might need to install a whole universe as deps.

I ran over the `Chat` and `ToolRegistry` implementation and failed to find an appropriate mechanism allowing me to inject a potentially long run logic **before** the chat is actually submitted. Another approach is to use an nvim command or slash command to **explicitly** start (maybe some) MCP servers, to avoid starting all servers on chat creation.

I personally use the `MCPServerStart`/`MCPToolsLoaded`/`MCPServerExit` events with `fidget.nvim` to setup a progress notification to make the startup progress visible to me.

图片

To sum up, we might have the following ways to address this:
1. Create a stub tool group without any tools, and resolve it once the server is ready. The server can be started on demand, but we still need to find a way to block chat submittion before all mentioned servers are ready.
2. Allow disabling auto-start for some servers (`auto_start = false`), and the users need to use some command to explictly start them.
3. Keep the currently behavior but give more clear status feedback to the users, so they know their config is ok just the servers need some time to start.

olimorris:
Appreciate this thoughtful post and your investigation.

+1 on status updates to users. 

I will also add in an option for users to start a chat without their MCP servers

olimorris:
Happy to pick this PR up now and get it to ready to merge. I'm going to merge it into `develop` ahead of a `v19.0.0` release as I will be making some breaking changes to how tools are called. Also gives me time to work with it properly and test it out in the real world.

cairijun:
I've been dog-feeding myself for ~1 month on daily work, with several MCP servers (`brave-search-mcp-server`, `mcp-server-fetch`, `server-sequential-thinking`, and a few of our internal ones). I can say it works pretty well and meets my daily needs.

There are a few things might worth improving, though.
1. We are somewhat strict when parsing msg lines from the servers, but there are non-compliant implementations in the wild. E.g. `mcp-server-fetch` will call `npm` to install some node deps, producing garbage data on its stdout. This generally does not break the normal flow (malformed lines are ignored), but the user will be interrupted and confused by a red error notification. I think we can add a "lenient" mode for those servers, being more quiet on non-fatal errors.
2. The OpenAI Responses API supports only a very limited subset of JSON schema in strict mode, which is not compatible with some servers. E.g. `mcp-run-python` uses `"additionalProperties" = true`, `mcp-server-fetch` uses `"format" = "uri"`.
3. (For the Responses API, again) We transform optional arguments to `null|XXX` in the tool schemas, allowing the LLM to use `null`s in the tool calls. But most MCP servers are rather strict on argument validation, not treating `null`s as "missing". This causes "expected XXX, received null" errors occasionally.

Thanks for you work @olimorris and I can't wait to see it launch!

olimorris:
Awesome thanks for the heads up. Your tests look really good and I will do a thorough look prior to merging to see how easily I can create something that will allow us to mimic the output from other MCP servers.

As a heads up, I've moved the MCP config to `config.mcp` from `config.interactions.chat.mcp`...anticipating that one day we may have MCP servers that are not restricted to just the chat interaction 😉.

I want to add a slash command to allow users to start/stop servers as well as update the config to specify servers that will autostart with a chat buffer. Gonna try and get all of that done this week.

olimorris:
- Added slash command so users can manually stop and start servers
- Servers can be turned off at the config level and enabled in the chat
- Completion engine automatically refreshes when mcp servers are added, keeping config immutable

cairijun:
The newly added 10s default timeout seems a bit too small. It affects the `initialize` request, which waits for the server to startup. 10s might not be adequate for `npx`.

Also `CodeCompanion.MCP.ToolOverride` is currently using `timeout_ms` as the config key, while `timeout` is used elsewhere. I think we better make it consistent.

olimorris:
Good spot. Will amend

olimorris:
Will aim to merge this over the weekend. Just the docs to go.

I will add a subsequent PR next week to allow users to specify a JSON file that contains their servers.

cairijun:
I have several little patches that might improve user experience:

1. Add output truncation to `success` callback to prevent Neovim freezing due to treesitter's poor performance with large texts.

`````diff
diff --git a/lua/codecompanion/mcp/tool_bridge.lua b/lua/codecompanion/mcp/tool_bridge.lua
index 3d8a3794f..315ae2298 100644
--- a/lua/codecompanion/mcp/tool_bridge.lua
+++ b/lua/codecompanion/mcp/tool_bridge.lua
@@ -34,13 +34,19 @@ local output = {
   success = function(self, tools, cmd, stdout)
     local chat = tools.chat
     local output = M.format_tool_result_content(stdout and stdout[#stdout])
+    local output_for_user = output
+    local DISPLAY_LIMIT_BYTES = 1000
+    if #output_for_user > DISPLAY_LIMIT_BYTES then
+      local utf_offset = vim.str_utf_start(output_for_user, 1 + DISPLAY_LIMIT_BYTES)
+      output_for_user = output_for_user:sub(1, DISPLAY_LIMIT_BYTES + utf_offset) .. "\n\n...[truncated]"
+    end
     local for_user = fmt(
       [[MCP: %s executed successfully:
 ````
 %s
 ````]],
       self.name,
-      output
+      output_for_user
     )
     chat:add_tool_output(self, output, for_user)
   end,
`````

2. Fix `error` callback to properly display error messages and tool arguments.

`````diff
diff --git a/lua/codecompanion/mcp/tool_bridge.lua b/lua/codecompanion/mcp/tool_bridge.lua
index 3d8a3794f..315ae2298 100644
--- a/lua/codecompanion/mcp/tool_bridge.lua
+++ b/lua/codecompanion/mcp/tool_bridge.lua
@@ -53,13 +59,16 @@ local output = {
     local chat = tools.chat
     local err_msg = M.format_tool_result_content(stderr and stderr[#stderr] or "<NO ERROR MESSAGE>")
     local for_user = fmt(
-      [[MCP: %s failed
+      [[MCP: %s failed:
 ````
 %s
+````
+Arguments:
+````%s
 ````]],
       self.name,
-      vim.inspect(self.args),
-      err_msg
+      err_msg,
+      vim.inspect(self.args)
     )
     chat:add_tool_output(self, "MCP Tool execution failed:\n" .. err_msg, for_user)
   end,
`````

3. Make `prompt` callback to include tool invocation arguments in the prompt message. Showing only the tool name seems less useful. Users can override the function if they find the prompt too verbose.

```diff
diff --git a/lua/codecompanion/mcp/tool_bridge.lua b/lua/codecompanion/mcp/tool_bridge.lua
index 3d8a3794f..315ae2298 100644
--- a/lua/codecompanion/mcp/tool_bridge.lua
+++ b/lua/codecompanion/mcp/tool_bridge.lua
@@ -69,7 +78,7 @@ local output = {
   ---@param tools CodeCompanion.Tools
   ---@return nil|string
   prompt = function(self, tools)
-    return fmt("Execute the `%s` MCP tool?", self.name)
+    return fmt("Execute the `%s` MCP tool?\nArguments:\n%s", self.name, vim.inspect(self.args))
   end,
 }
```

4. Support environment variable substitution in cmd configuration to enable MCP servers requiring API_KEY or other secrets passed via command-line arguments (e.g., `mcp-remote`).

```diff
diff --git a/lua/codecompanion/mcp/client.lua b/lua/codecompanion/mcp/client.lua
index d6405ea67..2ae8a1a01 100644
--- a/lua/codecompanion/mcp/client.lua
+++ b/lua/codecompanion/mcp/client.lua
@@ -99,8 +99,9 @@ function StdioTransport:start(on_line_read, on_close)
   self._on_close = on_close
 
   adapter_utils.get_env_vars(self)
+  local cmd = adapter_utils.set_env_vars(self, self.cmd)
   self._sysobj = self.methods.job(
-    self.cmd,
+    cmd,
     {
       env = self.env_replaced or self.env,
       text = true,
```

olimorris:
@cairijun great patches. Please push them

cairijun:
Pushed.

I also added a security warning about the `roots` feature in the doc to prevent potential misuse.

olimorris:
Fantastic work. So excited to have this in CodeCompanion.

There's some minor features I'll aim to add in before I release `v19.0.0`. Namely, allowing users to specify MCP servers in an external JSON file which will make adding servers without restarting Neovim that much easier. Potentially allow people to migrate from MCPHub.nvim to this as well.

I will also think about how we can better debug user's MCP issues. I'm expecting there will be a handful of edge cases with adapters (like as you said with the `openai_responses`).

Once again, thanks for your hard work @cairijun.

PR #2562, opened 2025-12-19
fix(inline): leave visual mode upon triggering inline interaction

chrisgrieser:
## Description
It is the idiomatic behavior of vim commands to leave visual mode upon triggering if you act upon the selection. Using the inline interaction, this is not the case, which this PR addresses.

## Related Issue(s)
  - Fixes #2538

## Checklist

- [ ] I've added [test](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md#testing) coverage for this fix/feature --> not relevant for this kind of change

### DIFF
--- a/lua/codecompanion/interactions/inline/init.lua
+++ b/lua/codecompanion/interactions/inline/init.lua
@@ -420,7 +420,7 @@ function Inline:submit(prompt)
   log:info("[Inline] Request started")
 
   if self.buffer_context.is_visual then
-    vim.cmd.normal({ vim.fn.mode(), bang = true })
+    pcall(vim.cmd.normal, { vim.fn.mode(), bang = true })
   end
 
   -- Inline editing only works with streaming off - We should remember the current status


olimorris:
To be a nuisance, is there an authority that references this that we could include in this PR?

chrisgrieser:
Cannot think of one, since I assume it's too fundamental to be spelled out in the vim docs? Like, I cannot think of any vim command (or any plugin command for that matter) that leaves you in visual mode after acting on it. You can try with `gU` on a visual selection, which leaves visual mode.

I mean it makes sense from a UX perspective: you do the visual selection to give the command, in this case the AI, some input. There is no other reason you entered visual mode, and thus no reason to stay in visual mode after having fulfilled that purpose.

Pardon the maybe dumb analogy, but it's like keeping the spoon in your hand after you finished eating. There is no reason for it.

olimorris:
Haha no all good. Just wanted to make sure there wasn't somethingthat Bram had written many years ago that we were flouting.

lua/codecompanion/interactions/inline/init.lua
```diff
@@ -419,6 +419,13 @@ local _streaming = true
 function Inline:submit(prompt)
   log:info("[Inline] Request started")
 
+  -- it's the idiomatic behavior for visual mode commands in vim to leave visual
```
olimorris: We don't need these comments. It's totally clear what this is doing

lua/codecompanion/interactions/inline/init.lua
```diff
@@ -419,6 +419,13 @@ local _streaming = true
...
+  -- mode upon triggering them (#2538)
+  local mode = vim.fn.mode()
+  if mode:find("[Vv]") then
```
olimorris: We can actually use `self.buffer_context.is_visual` on the inline class. That captures the context of the buffer at the time the inline request is made.

Unlikely anything will change in those milliseconds but is more robust and reuses functionality we've already defined.

Not that you'd have spotted that because the LuaLS annotations have not been updated.

lua/codecompanion/interactions/inline/init.lua
chrisgrieser: The comment is not describing what the line does, but rather why it does so (because it's idiomatic in vim). But yeah, It's probably not necessary, removed it.

lua/codecompanion/interactions/inline/init.lua
chrisgrieser: fixed

chrisgrieser:
Addressed both issues in latest commit.

> Just wanted to make sure there wasn't somethingthat Bram had written many years ago that we were flouting.

Agree, I am also not a fan of cargo cults, haha.

chrisgrieser:
> FAIL in tests/interactions/inline/test_inline.lua | Inline | forms correct prompts: ...nion.nvim/lua/codecompanion/interactions/inline/init.lua:423: Vim:E35: No previous regular expression
  Traceback:
    ...nion.nvim/lua/codecompanion/interactions/inline/init.lua:423
    tests/interactions/inline/test_inline.lua:104

and the line: https://github.com/chrisgrieser/codecompanion.nvim/blob/498f0e8cc414143593ed6846c4206ce90ccf7a72/tests/interactions/inline/test_inline.lua#L104

Hmm, I think do not know enough about the text suite to tell what the issue here is. 🤔

olimorris:
FYI, I reverted this PR as in #2605 it was noted that this was causing issues

chrisgrieser:
From the bug description, it looks like the normal command ran `n` to trigger a search, which would suggest that `buffer_context.is_visual` is not reliable. 

The original version of this PR that checked for the mode directly would avoid the bug, if that is indeed the cause.

PR #2600, opened 2025-12-27
New diff

olimorris:
## Description

2026-01-14 18_51_38 - Ghostty@2x

Ref: https://github.com/zed-industries/zed/pull/43269

This PR will enable a more thorough diff experience in CodeCompanion, allowing for:

- Syntax highlighting for additions and deletions (current implementation allows for just additions)
- Dynamic diff - for inline interactions, can display the diff inline or show a large one in a floating window
- Hunk navigation
- Word level highlighting (similar to Zed's diff)
- Keymap instructions that update when you navigate hunks
- Unified diff across the builtin tools and ACP tools
- Integration testing with screenshots - making it easier to debug diff output

This PR will also see the removal of support for other diff providers as I seek to streamline CodeCompanion into 2026, keeping it simpler and leaner. Unfortunately this means a number of destructive changes:

1. No support for mini.diff and the builtin split provider
2. Replacement of the current inline provider
3. Removal of the superdiff feature and edit tracker

## AI Usage

Mixture of LLMs were used throughout this PR - it was a long process. I initially asked Claude Code to summarize the approach taken by [sidekick.nvim](https://github.com/folke/sidekick.nvim) and the diff in Zed. This gave me a basis to determine how I'd like the CodeCompanion diff to be. Then in-conjunction with Claude Code over a month (it was Xmas...) we riffed on this final solution. Commands and tests were all my own as per my policy of "an LLM can create a feature or the tests but cannot do both".

## Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## [18.5.1](https://github.com/olimorris/codecompanion.nvim/compare/v18.5.0...v18.5.1) (2026-01-24)
+
+
+### Bug Fixes
+
+* **adapters:** check for nil in `openai_responses` ([#2662](https://github.com/olimorris/codecompanion.nvim/issues/2662)) ([e22c043](https://github.com/olimorris/codecompanion.nvim/commit/e22c04336b5c56e3839bfebd6d88944f014f4b30))
+* **adapters:** copilot supported endpoints ([#2691](https://github.com/olimorris/codecompanion.nvim/issues/2691)) ([dd98466](https://github.com/olimorris/codecompanion.nvim/commit/dd98466a893abf499fbd69ab9526b2da7c094fb8))
+* **tools:** remove globals for better concurrent tool usage ([#2680](https://github.com/olimorris/codecompanion.nvim/issues/2680)) ([c9d74dd](https://github.com/olimorris/codecompanion.nvim/commit/c9d74dd667cf609b4f2064ae7f5471285b5356cb))
+
 ## [18.5.0](https://github.com/olimorris/codecompanion.nvim/compare/v18.4.1...v18.5.0) (2026-01-21)
 
 
--- a/doc/codecompanion.txt
+++ b/doc/codecompanion.txt
@@ -5280,8 +5280,8 @@ required.
                 role = "user",
                 opts = { auto_submit = true },
                 -- Scope this prompt to the cmd_runner tool
-                condition = function()
-                  return _G.codecompanion_current_tool == "cmd_runner"
+                condition = function(chat)
+                  return chat.tools.tool and chat.tools.tool.name == "cmd_runner"
                 end,
                 -- Repeat until the tests pass, as indicated by the testing flag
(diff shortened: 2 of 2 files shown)


olimorris:
Hey @bassamsdata, hope you're well!

I wanted to give you the head up regarding this PR that I've merged into `develop`. I've made the decision to streamline a whole host of CodeCompanion features in 2026, which will form part of some breaking changes for `v19.0.0`.

I'm removing support for all other diff providers/types (mini.diff and the split) alongside the Super Diff. I want to keep the feature set for CodeCompanion as tight as possible to continue to allow me to continue adding new features at a decent pace.

I went through every feature over Christmas and asked myself _"is this what core CodeCompanion should be providing for users? Or, should it be its own extension?"_. Regarding the Super Diff, it felt to me that it should be its own extension but CodeCompanion should provide the necessary events to enable it. So wanted to give you the heads up should you choose to build it out into its own plugin. I don't expect to be merging `v19.0.0` for a couple of weeks btw.

Don't want this to come across as me not appreciating your work. Far from it!

bassamsdata:
Hey @olimorris, Hope you're doing well too, all good on my side, and thanks for the heads up.

Totally agree with you, tbh. I really understand the burden of maintaining a core, big plugin, especially around the diffs module, that area alone has raised a lot of issues and discussions from users over time. Streamlining makes complete sense to me, and I have no issues with that decision at all.

Regarding Super Diff, I completely get your reasoning. I also noticed that some users were getting confused between Super Diff and the approval diffs themselves, so your call here is very reasonable. 

For what it's worth, I actually built the tracking + Super Diff logic with a different UI (for both ACP and HTTP) as part of an extension called [fs-monitor.nvim](https://github.com/bassamsdata/fs-monitor.nvim). It's fully working and I use it daily, so I think your instinct to make it a separate extension is spot on. I just haven't had the time yet to properly announce it.

I'll try out the `develop` branch as well, honestly, having a clear branch for future-breaking releases is really nice to see.

Please don't worry about how it comes across! Not at all,  I completely respect the decision. I use CodeCompanion daily for almost everything, so appreciation definitely goes both ways. 

I actually wanted to discuss a few enhancements/fixes with you (mostly around tools, which are very integral to my day-to-day), but the start of the new year was a bit hectic. I'll probably open a discussion about that soon.

Thanks again for the heads up, and really appreciate the transparency here 🙏

dhruvinsh:
@olimorris , since the events like `CodeCompanionDiffAttached` is not available with this changes, what is the best alternative right now?

I had some autocommand working for the commit message diff acceptance as below,

```lua
-- codecompanion: gitcommit accept
vim.api.nvim_create_autocmd({ "User" }, {
  pattern = "CodeCompanionDiffAttached",
  group = utils.augroup("gitcommit_accept", true),
  callback = function(request)
    local bufnr = request.data.bufnr
    if vim.bo[bufnr].filetype == "gitcommit" then
      vim.defer_fn(function()
        vim.api.nvim_feedkeys("gda", "t", false)
      end, 200)
    end
  end,
})
```

PR #2624, opened 2026-01-07
Anthropic token count fixes

slg95:
## Description

The Anthropic token count had two issues:

- The token count was susceptible to a race condition if multiple Anthropic chat windows were communicating with an LLM at the same time
- The token count was not taking into account cached input tokens

The second issue I could understand if you don't want changed because if the primary concern is token cost, then the cached tokens should not be shown. However, this would be misleading anyway since what is shown is only the token count for the last request/response. My primary concern is the context window size, so I believe the accurate number needs to be shown in the chat window for all tokens currently in context.  

## Checklist

- [ ] I've run `make all` to ensure docs are generated, tests pass and [StyLua](https://github.com/JohnnyMorganz/StyLua) has formatted my code

### DIFF
--- a/lua/codecompanion/adapters/http/anthropic.lua
+++ b/lua/codecompanion/adapters/http/anthropic.lua
@@ -415,6 +415,7 @@ return {
           if json.type == "message_start" then
             self.temp.input_tokens = (json.message.usage.input_tokens or 0)
               + (json.message.usage.cache_creation_input_tokens or 0)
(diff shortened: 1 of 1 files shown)


olimorris:
Super, thanks for this! I hadn't ever imagined anyone would have multiple windows open at the same time so left it like that to hear if anyone was doing anything cool with parallel executions.

Issue #2620, opened 2026-01-07
[Bug]: Tool call breaks chat buffer for concatenated JSON response `{}{}`


bitromortac:
### Pre-submission checklist

- [x] I have read the [documentation](https://codecompanion.olimorris.dev)
- [x] I have updated the plugin and all dependencies to the latest versions
- [x] I have searched for existing [issues](https://github.com/olimorris/codecompanion.nvim/issues) and [discussions](https://github.com/olimorris/codecompanion.nvim/discussions)
- [x] My issue is not a minor or cosmetic quirk (e.g., formatting, spacing, or other non-functional details)

### Neovim version (nvim -v)

NVIM v0.11.4

### Operating system/version

Debian GNU/Linux 12 (bookworm)

### Adapter and model

Gemini 3 Pro Preview

### Describe the bug

When the agent makes a tool call with invalid JSON encoding it leads to a decoding failure and that breaks the chat buffer.

Further usage of the chat buffer leads to:
```
[ERROR] 2026-01-07 16:46:21
Error: [{
  "error": {
    "code": 400,
    "message": "Request contains an invalid argument.",
    "status": "INVALID_ARGUMENT"
  }
}
]
```

### Steps to reproduce

When in a folder
```
test
├── 1
├── 2
├── 3
├── 4
└── 5

1 directory, 5 files
```

And I instruct `:CodeCompanionChat read all files in here at once @{files}`, I get the following error:

```
[ERROR] 2026-01-07 16:42:09
Couldn't decode the tool arguments: {"start_line_number_base_zero":0,"filepath":"1","end_line_number_base_zero":-1}{"end_line_number_base_zero":-1,"filepath":"2","start_line_number_base_zero":0}{"filepath":"3","end_line_number_base_zero":-
1,"start_line_number_base_zero":0}{"start_line_number_base_zero":0,"end_line_number_base_zero":-1,"filepath":"4"}{"end_line_number_base_zero":-1,"start_line_number_base_zero":0,"filepath":"5"}
```

### Expected behavior

The tool call should error gracefully and report back to the agent that it has used the wrong format and should try again.

### minimal.lua file

```lua
vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

-- Your CodeCompanion setup
local plugins = {
	{
		"olimorris/codecompanion.nvim",
		dependencies = {
[...]

PR #2629, opened 2026-01-08
fix(tools): gemini concatenated tools

olimorris:
## Description

If an LLM function calls with incorrect JSON, CodeCompanion would handle this but fail to return the chat buffer to a usable state. An autocmd wasn't created early enough in the tool's execution.

Also, Gemini would commonly make multiple tool calls by concatenating the arguments instead of issuing individual function calling requests. This is now accounted for at an adapter level.

## Related Issue(s)

#2620 

## Checklist

### DIFF
--- a/lua/codecompanion/adapters/http/gemini.lua
+++ b/lua/codecompanion/adapters/http/gemini.lua
@@ -1,6 +1,24 @@
 local adapter_utils = require("codecompanion.utils.adapters")
 local openai = require("codecompanion.adapters.http.openai")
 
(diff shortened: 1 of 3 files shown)


bitromortac:
This seems to have introduced a regression in that code edits are cut off if "}{" appears in the edit contents here, https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/adapters/http/gemini.lua#L4-L20

Perhaps a solution would be to do a check if `args` is valid JSON first before fixing, at the cost of parsing.

PR #2632, opened 2026-01-08
initial take on anthropic docs api

georgeharker:
I have not yet added tests or README updates.

I had need of this myself.

Just curious on interest here - I had code companion do the bulk of the work with guidance.  It looks pretty much in line with how images and files work.

## Description

This adds a slash command (/document) to attach doc types like docx, pdf, rtf which allows anthropic users to pass files directly through the adapter.

## Related Issue(s)

https://github.com/olimorris/codecompanion.nvim/discussions/2216

## Checklist

### DIFF
--- a/lua/codecompanion/utils/attachments.lua
+++ b/lua/codecompanion/utils/attachments.lua
@@ -32,15 +32,19 @@ local function build_attachment_types()
 end
 
 ---Attachment-specific MIME types (fallback for types not in files_utils)
+---These are document and extended image types not in the minimal files.lua map
 ---@type table<string, string>
 local ATTACHMENT_MIME_TYPES = {
-  -- Images
+  -- Extended image types
   bmp = "image/bmp",
   tiff = "image/tiff",
   svg = "image/svg+xml",
-  -- Documents
+  -- Document types
+  rtf = "text/rtf",
   xlsx = "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
   pptx = "application/vnd.openxmlformats-officedocument.presentationml.presentation",
+  csv = "text/csv",
+  docx = "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
 }
 
 ---Get image file extensions from config
--- a/lua/codecompanion/utils/files.lua
+++ b/lua/codecompanion/utils/files.lua
@@ -221,10 +221,6 @@ function M.get_mimetype(path)
     png = "image/png",
     webp = "image/webp",
     pdf = "application/pdf",
-    rtf = "text/rtf",
-    xslx = "application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",
-    csv = "text/csv",
-    docx = "application/vnd.openxmlformats-officedocument.wordprocessingml.document",
   }
 
   local extension = vim.fn.fnamemodify(path, ":e")


olimorris:
Hey George. It's like you read my mind....I was giving this some serious thought earlier today.

Normally, I'd try and add a feature like this to all http adapters but given the enormity of that as a task, I'd rather we get something like this into CodeCompanion.

Let me know when you're happy with it and I can take a look later down the line. In the same way I restrict the image slash command by looking for `adapter.opts.vision`, via the config, I think we should add `file` (or whatever Anthropic call it) to the Anthropic adapter. Not sure if any of that makes sense....long day!

georgeharker:
great! I've updated to check for doc_upload capabilities rather than vision.

Added some docs.

I'm actually not sure what other adapters support pdf / docx and i'm sure there's possibly some differences in implementation.  I'm pretty sure that opencode and copilot don't properly proxy these requests to models even with an anthropic model requested, though they appear to read pdfs just fine using their inbuilt tools.

olimorris:
One reflection I had was "as a user, should I be expected to know the difference between `/file` and `/document`?" I'm cool with keeping document and file as separate classes, just think it makes the UX cleaner. If a user tries to send a PDF to an OpenAI adapter if we've set `opts.files = false` or it comes back as `nil`, we could just alert the user that this isn't supported yet.

georgeharker:
I was thinking about that too. 

I’ll try and tidy that up. 

It also looks like acp can ingest embedded files similar to how it does for images so working on that.

olimorris:
> It also looks like acp can ingest embedded files similar to how it does for images so working on that.

That's good to know. I have a feeling how I'm sharing files with ACP agents may not be correct. I can take a look at that too.

georgeharker:
> One reflection I had was "as a user, should I be expected to know the difference between `/file` and `/document`?" I'm cool with keeping document and file as separate classes, just think it makes the UX cleaner. If a user tries to send a PDF to an OpenAI adapter if we've set `opts.files = false` or it comes back as `nil`, we could just alert the user that this isn't supported yet.

I guess one issue is that /document allows attaching from remote URLs or using file_api if enabled.  I'll def tidy that portion up so you can look.  I do like the idea of a unified interface for sure, just not sure how to include URL and file api support to /file

olimorris:
We could always take the logic that I put in the `/image` slash command where the user is asked to select a URL or a file beforehand. That feels like a reasonable compromise to keep everything within one slash command.

georgeharker:
That’s what I did for document. I’m concerned that for folks used to /file just giving a picker it would now present and additional choice first. Maybe that’s not a big deal

georgeharker:
I'm happy to look at merging /file and /document.  Couple of options before I do that

1. I think merging implementations might be sensible given there's duplication between image and document at least
2. Should /document be kept separate but called /attachment - /file approximately means /textfile (not suggesting altering that) but maybe there are useful destinctions.  We could even have /attachment deal with text files - optionally fetched from a URL, but keep /file to be strictly local text file
3. alternatively should /document be rolled into /file - in which case it would present a url or file dialog first, which seems like it might be irritating for current use cases.

georgeharker:
I guess finally there's the option of /image and /document being merged (possibly into /attachment)

Lmk, and I can look at getting it done

georgeharker:
I've updated this PR with a /attachment approach.  

/files -> remains just straight to picker
/images -> remains as was

/attachment allows files | images | pdf | docx etc or URL based versions of these and supports Files API where enabled.

This presents a chooser for URL | FILE | Files API first then file chooser.

Attachments are unlike files which may get edited and may need tracking, and are intended just to be consumed as context.

The attachments for users used to files 1-interaction nature is not broken, though you can use /attachment to add images or files if you like

georgeharker:
Ps. If your prefer this be part of file I can do that - just wonder what image means in that context.

olimorris:
Thanks, George.

I'm going to mull this PR over. Having a separate slash command for PDFs, docx, etc via `/attachment` isn't particularly clean. If I'm a user, should I share my _hello_world.py_ file with an LLM via the `/file` slash command or via `/attachment`? I do technically want to attach it to my request. And then we can add images via `/attachment` too?

I think this PR is making the UX too complicated.

I also want to take the time to consider how I would add support for OpenAI's responses API, Copilot and Gemini. Historically, when I added function calling and image support, mapping out what I needed to do for all the other adapters gave me clarity on how the final API should look. I am going to opt to do the same here.

georgeharker:
Understood. There's definitely a tension between whats supported everywhere and adapter specific features.

georgeharker:
Just curious if there are adaptations you’d have liked to the PR, if you plan on doing something similar etc.

PR #2646, opened 2026-01-11
refactor(adapters): gemini parallel tool calls

olimorris:
## Description

Further refactor to ensure that the handling of Gemini's erroneous tool calls is more robust.

As a recap, this and the earlier PR (#2629) are to fix the way Gemini incorrectly parallel calls tools. In the case of trying to use the `@{read_file}` tool:

```json
{"filepath":"test.lua","start":0,"end":-1}{"filepath":"other.lua","start":0,"end":-1}
```

There are likely to be edge cases that continue to get the better of this refactor. With that in mind, we should be aiming to capture and handle these errors and notify Gemini of it's incorrect function calling rather than try and patch.

## Related Issue(s)

#2620 #2629

## Checklist


olimorris:
CC: @bitromortac

PR #2660, opened 2026-01-16
fix: cmp insert runs prematurely

pwang2:
## Description

When using SelectBehavior.Insert in nvim-cmp Tab mappings (a common pattern), cycling through #buffer or @tool completions stops at the 2nd entry because insertText is set on the items, causing immediate text insertion which re-triggers or closes the completion menu.

## AI Usage

Bug was explored and explained in opencode clearly

## Screenshots

image

## Checklist

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,7 +17,7 @@ jobs:
       matrix:
         os: [ubuntu-latest]
         nvim_tag: [nightly, v0.11.0]
-    name: ${{ matrix.os }}
+    name: ${{ matrix.os }} / ${{ matrix.nvim_tag }}
     runs-on: ${{ matrix.os }}
     env:
       NVIM: ${{ matrix.os == 'windows-latest' && 'nvim-win64\\bin\\nvim.exe' || 'nvim' }}
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,14 @@
 # Changelog
 
+## [18.4.1](https://github.com/olimorris/codecompanion.nvim/compare/v18.4.0...v18.4.1) (2026-01-16)
+
+
+### Bug Fixes
+
+* **acp:** fire event when calling /mode command ([#2656](https://github.com/olimorris/codecompanion.nvim/issues/2656)) ([8ebbcc2](https://github.com/olimorris/codecompanion.nvim/commit/8ebbcc27d20e9ac6e7491de6ef9aac1fbf6277a2))
+* **adapters:** copilot log error ([#2663](https://github.com/olimorris/codecompanion.nvim/issues/2663)) ([6fa8338](https://github.com/olimorris/codecompanion.nvim/commit/6fa8338b4332d57c153b40205613c3c56736a255))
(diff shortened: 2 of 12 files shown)


pwang2:
Sorry I can't make the `make test  pass here.. So Ieave it as draft here.

olimorris:
All looks good to me.

I'll give this a try later on this week and then we can merge if you're happy.

My question is, for users that don't have the same config and behaviour set up as you, how does this impact them if at all?

pwang2:
> All looks good to me.
> 
> I'll give this a try later on this week and then we can merge if you're happy.
> 
> My question is, for users that don't have the same config and behaviour set up as you, how does this impact them if at all?

Thanks, @olimorris.    you are right,  the syntax detail seems only for `nvim-cmp` here.  and the change are restricted to the cmp completion provider here.  so it should be good.

https://github.com/olimorris/codecompanion.nvim/tree/main/lua/codecompanion/providers/completion/cmp

olimorris:
Sorry I wasn't clear. I meant that you have `cmp.select_next_item({ behavior = cmp.SelectBehavior.Insert })` set in your nvim-cmp config. For other cmp users that don't have that, how does this PR affect them?

I don't want to fix this for you and break it for others.

pwang2:
https://github.com/hrsh7th/nvim-cmp/blob/85bbfad83f804f11688d1ab9486b459e699292d6/doc/cmp.txt#L192

the possible behavior here are either insert or select. 

- when user uses `Select`,  with the PR change, it will still select the completion item and await another key stoke to insert. so it will be okay. 

- when user uses `Insert`(as I do),  the PR also works as  it only prepares for the completion items and leave the handling to nvim-cmp. 

I just tested with my branch with both cases, both work. 

BTW, the test failure seems to be irrelevant to my code change here but my local setup issue as I still see the error on main branch.  The checks are passed in the actions.

olimorris:
Testing this out now. This PR overwrites preceding text in the chat buffer:

https://github.com/user-attachments/assets/4d048cc6-c338-4fc5-8cae-a3af2fc5d04f

Before:

https://github.com/user-attachments/assets/23cdfb13-1392-4eee-8a0d-b3acd57bc088

I was testing this with the `minimal.lua` which perhaps isn't configured to reflect how cmp users have their config.

pwang2:
Sorry , @olimorris, you are right.  My test is too naive without leading text..  I will check again.  Thank you for your time. 

ps: You might also notice that  / command has the same behavior on main branch.    So we need a consistent handling in general. 

<video src='https://github.com/user-attachments/assets/43324e7b-4dd6-477b-80d3-9372fbe255d7'/>

alvaromrveiga:
I am a little late to this but in case someone finds it: 

I fixed it by changing the behavior of my selection keymaps. Since `nvim-cmp` uses `cmp.SelectBehavior.Insert` by default, switching to `cmp.SelectBehavior.Select` does not modify the buffer so `TextChangedI` is not triggered during selection. 

Example config snippet:

```lua
local cmp = require("cmp")
-- ...
["<C-j>"] = cmp.mapping(function(fallback)
  if cmp.visible() then
    cmp.select_next_item({ behavior = cmp.SelectBehavior.Select })
  else
    fallback()
  end
end, { "i", "s" }),
["<C-k>"] = cmp.mapping(function(fallback)
  if cmp.visible() then
    cmp.select_prev_item({ behavior = cmp.SelectBehavior.Select })
  else
    fallback()
  end
end, { "i", "s" }),
```

PR #2680, opened 2026-01-21
fix: Remove tool related globals to support concurrent tool usage.

slg95:
## Description

The globals codecompanion_current_tool and codecompanion_cancel_tool are not needed (_G.codecompanion_cancel_tool is never used) and turn concurrent tool calls into a race condition. Instead, workflows can use the chat parameter they are already provided in the condition() function callback to check the current running tool.

I plan on making a follow up PR to fix the fact that tools are not cancellable right now, but one step at a time.

## AI Usage

I used Claude Sonnet initially, but ended up making the edits myself after it tried to overcomplicate things

## Checklist

- [X] _(optional)_ I've updated the README and/or relevant docs pages

### DIFF
--- a/doc/codecompanion.txt
+++ b/doc/codecompanion.txt
@@ -5331,8 +5331,8 @@ required.
                 role = "user",
                 opts = { auto_submit = true },
                 -- Scope this prompt to the cmd_runner tool
-                condition = function()
-                  return _G.codecompanion_current_tool == "cmd_runner"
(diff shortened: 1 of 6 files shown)


olimorris:
Thanks for this. A PR to make tools cancellable would be welcome.

> I used Claude Sonnet initially, but ended up making the edits myself after it tried to overcomplicate things

Can relate!

PR #2683, opened 2026-01-23
fix(acp): make ACP connection async to prevent UI hang on chat open

aweis89:
## Description

Fix UI hang/freeze when opening a chat with ACP adapters (like Claude Code, OpenCode).

When opening a chat with an ACP adapter, Neovim would freeze because `connect_and_initialize()` used blocking `vim.wait()` calls for initialize, authenticate, and session creation requests. Even though the connection was triggered via `vim.schedule()`, the blocking calls would freeze the UI for several seconds.

This PR introduces async RPC request handling:
- `send_rpc_request_async(method, params, callback)` - Non-blocking RPC requests with callback dispatch
- `connect_and_initialize_async(callback)` - Chains init steps using async callbacks instead of blocking
- `helpers.create_acp_connection_async(chat, callback)` - Creates connection without blocking UI
- Updated `ACPHandler:ensure_connection()` to handle partially-connected states

The UI now remains responsive while ACP connections establish in the background. If the user sends a message before async connection completes, `ensure_connection()` completes it synchronously (fast since the process is already started).

## AI Usage

AI (Claude Opus 4.5 via Cursor and OpenCode) wrote the initial implementation before refactoring, as well as the tests.

## Checklist

### DIFF
--- a/lua/codecompanion/acp/init.lua
+++ b/lua/codecompanion/acp/init.lua
@@ -539,8 +539,6 @@ end
 ---@param id number
 ---@return table|nil
 function Connection:wait_for_rpc_response(id)
-  -- Timeout is already in milliseconds (adapter.defaults.timeout = 20000)
-  -- vim.wait() expects milliseconds, so no conversion needed
   local timeout = self.adapter_modified.defaults.timeout or TIMEOUTS.DEFAULT
 
   local ok = vim.wait(timeout, function()


olimorris:
Firstly, appreciate the PR and the clear problem statement. This is not something I've actually experienced. Can you be clear about your setup and how you have OpenCode configured?

Some initial thoughts:
- At a summary level this is net +300 LOCs and the sync version does this in probably about 70. I would have expected an async PR to reuse a lot of that logic
- There looks to be duplication across `connect_and_initialize()` and `connect_and_initialize_async`
- The new methods introduced have zero test coverage

Finally, can you ensure the PR conforms to the default template - it may seem onerous but it makes it makes much easier for me to review all PRs if they look the template, including the AI usage declaration.

aweis89:
> This is not something I've actually experienced. Can you be clear about your setup and how you have OpenCode configured?

This is my current OpenCode config. I apologize I didn't have a chance to narrow down the root cause yet, however I am using mcp + a cursor plugin. (I did of course confirm that with the changes in this PR, there is no discernible startup lag whatsoever, and I can immediately start inserting my prompt in the chat buffer):
```
{
  "$schema": "https://opencode.ai/config.json",
  "plugin": [
    "yet-another-opencode-cursor-auth"
  ],
  "mcp": {
    "sequential-thinking": {
      "type": "local",
      "command": [
        "npx",
        "-y",
        "@modelcontextprotocol/server-sequential-thinking"
      ],
      "enabled": true
    }
  },
  "theme": "gruvbox",
  "autoupdate": true,
  "share": "disabled",
  "model": "{env:OPENCODE_MODEL}",
  "permission": {
    "edit": "allow",
    "bash": "allow",
    "webfetch": "allow",
    "question": "deny"
  },
  "keybinds": {
    "messages_half_page_up": "ctrl+u",
    "messages_half_page_down": "ctrl+d",
    "session_child_cycle": "ctrl+l",
    "session_child_cycle_reverse": "ctrl+h"
  },
  "command": {
    "commit-staged": {
      "template": "Analyze staged changes and create a commit following conventional commit format (type(scope): description). Common types: feat, fix, refactor, docs, style, test, chore. Keep subject under 50 chars. Create the commit and output only the final commit message.",
      "agent": "build",
      "description": "Generate commit message for staged changes"
    }
  },
  "provider": {
    "cursor": {
      "name": "Cursor"
    },
  }
}
```

lua/codecompanion/acp/init.lua
```diff
@@ -386,23 +537,22 @@ end
 
 ---Wait for a specific response ID
 ---@param id number
----@return nil
+---@return table|nil
 function Connection:wait_for_rpc_response(id)
-  local start_time = uv.hrtime()
-  local timeout = (self.adapter_modified.defaults.timeout or TIMEOUTS.DEFAULT) * 1e6 -- Nanoseconds to milliseconds
```
olimorris, line 392: Sense check on this, we're losing the multiplier?

aweis89: The old code used manual timeout tracking using `uv.hrtime()` which is in nanoseconds so the conversation was needed.

Now that we're just using `vim.wait`, which handles timeout internally, and expects milliseconds, I don't believe the conversion is necessary.

```
vim.wait({time}, {callback}, {interval}, {fast_only})             *vim.wait()*
    Waits up to `time` milliseconds, until `callback` returns `true`
    (success). Executes `callback` immediately, then at intervals of
    approximately `interval` milliseconds (default 200). Returns all
    `callback` results on success.
```
```
uv.hrtime()                                                        *uv.hrtime()*

                Returns a current high-resolution time in nanoseconds as a
                number. This is relative to an arbitrary time in the past. It
                is not related to the time of day and therefore not subject to
                clock drift. The primary use is for measuring time between
                intervals.

                Returns: `integer`
```

lua/codecompanion/acp/init.lua
```diff
@@ -539,6 +539,8 @@ end
 ---@param id number
 ---@return table|nil
 function Connection:wait_for_rpc_response(id)
+  -- Timeout is already in milliseconds (adapter.defaults.timeout = 20000)
```
olimorris: We don't need to comment this in the code - it just noise. Just address via a comment on GitHub.

jugarpeupv:
ey @olimorris, is there something else needed for this to get merged?

i was using this fork until v19 was merged

thanks for maintaining this wonderful plugin btw ^^

olimorris:
All, closing this PR because this is not something that I can willingly accept without an issue demonstrating this behaviour.

There is inevitably going to be some lag with ACP adapters as you're communicating with a CLI tool that has some boot up. I'm fine with that as it's a one-off cost.

Gone are the days where I'm willing to accept a 900 LOC PR into the plugin without being able to understand why we need it.

jugarpeupv:
@olimorris from what i understood, the purpose of this PR was to make the acp connection async, so neovim does not freeze till the acp connection is established

maybe it is possible to discuss the implementation but in my opinion this is quite important, say you switch from one project to another in different neovim instances, everytime you open codecompanion you are frozen, you cant for example start pasting text into the buffer to mount the query for the llmm

jugarpeupv:
> All, closing this PR because this is not something that I can willingly accept without an issue demonstrating this behaviour.
> 
> There is inevitably going to be some lag with ACP adapters as you're communicating with a CLI tool that has some boot up. I'm fine with that as it's a one-off cost.
> 
> Gone are the days where I'm willing to accept a 900 LOC PR into the plugin without being able to understand why we need it.

@olimorris This is the demostration, it takes about 2 or 3 seconds for the codecompanion chat to be interactive

https://github.com/user-attachments/assets/7cd0b588-1fc4-4661-bebd-b72eae979088

olimorris:
All - #2978 now implements this. There was already an async library in CodeCompanion that I was able to leverage which reduced the LOC significantly.

Thank you for the initial work on this.

PR #2699, opened 2026-01-26
fix(log): Prevent double reporting of errors.

slg95:
## Description

When an error was logged, it triggered the log handlers for both ERROR and WARN. So anytime log:error() was called, the message was printing from both the echo and notify handlers.

The issue is that when we setup the handlers:

```lua
  -- Set the log root
  log.set_root(log.new({
    handlers = {
      {
        type = "echo",
        level = vim.log.levels.ERROR,
      },
      {
        type = "notify",
        level = vim.log.levels.WARN,
      },
      {
        type = "file",
        filename = "codecompanion.log",
        level = vim.log.levels[config.opts.log_level],
      },
    },
  }))
```

`vim.log.levels.WARN` < `vim.log.levels.ERROR`. So anytime the echo handler will run, the notify must also run. So the errors will always be reported twice since the log handler:

```lua
function LogHandler:log(level, msg, ...)
  if self.level <= level then
    local text = self.formatter(level, msg, ...)
    self.handle(level, text)
  end
end
```
always logs if the level is <= the configured handler level.

Another potential issue here is that the echo and notify handlers do not take into account the configured log level. So warning messages will still appear even if someone configures log level as `ERROR`. Not sure if that is intentional or not, since the configuration value on affects the "file" handler right now.

## AI Usage
After a few rounds of prompting, Claude Sonnet 4.5 recommended the change (he got the cause of the double error printing wrong twice, but the third time he got it)

## Checklist

### DIFF
--- a/lua/codecompanion/init.lua
+++ b/lua/codecompanion/init.lua
@@ -385,10 +385,6 @@ CodeCompanion.setup = function(opts)
   log.set_root(log.new({
     handlers = {
       {
-        type = "echo",
-        level = vim.log.levels.ERROR,
-      },
-      {
         type = "notify",
         level = vim.log.levels.WARN,
       },


olimorris:
This isn't something I've ever been aware of. Can you share a minimal.lua file that encapsulates the error so I can test this?

slg95:
> This isn't something I've ever been aware of. Can you share a minimal.lua file that encapsulates the error so I can test this?

I had Claude make this, read through it and it seems reasonable:

```lua
-- test_double_error.lua - Test for double error printing fix
--
-- Run from the codecompanion.nvim directory with:
--   nvim --headless -u test_double_error.lua
--
-- Or interactively:
--   nvim -u test_double_error.lua

-- Minimal setup - just add runtime paths
vim.opt.rtp:prepend(vim.fn.getcwd())
vim.opt.rtp:prepend(vim.fn.expand("~/.local/share/nvim/lazy/plenary.nvim"))
vim.opt.rtp:prepend(vim.fn.expand("~/.local/share/nvim/lazy/nvim-treesitter"))

-- Load treesitter (needed for codecompanion)
vim.cmd([[runtime plugin/nvim-treesitter.lua]])

-- Setup codecompanion
  opts = {
    log_level = "ERROR",
  },
})

-- Run the test
local log = require("codecompanion.utils.log")

print("\n")
print("=====================================================")
print("Testing log:error() - should print ONCE, not twice")
print("=====================================================")
print("")

log:error("TEST ERROR: If you see this twice, bug is NOT fixed")

vim.schedule(function ()
  print("")
  print("=====================================================")
  print("Test complete.")
  print("=====================================================")
  print("")
end)

-- Exit if headless
if #vim.api.nvim_list_uis() == 0 then
  vim.cmd("qa!")
end

```

olimorris:
Thanks for that. Can confirm this is in fact the case

PR #2705, opened 2026-01-29
chore(copilot): use GH_HOST for Copilot endpoints and token lookup

AmadeusK525:
This PR is a continuation of #2090, but with a complete "fix" for the GH_HOST issue. There are several places where the variable needs to be read-referenced, and [as I said there](https://github.com/olimorris/codecompanion.nvim/pull/2090#issuecomment-3596631490) I don't think this fix is actually good, but it does work (I've been using my fork for a couple of months).

### DIFF
--- a/lua/codecompanion/adapters/http/copilot/init.lua
+++ b/lua/codecompanion/adapters/http/copilot/init.lua
@@ -34,8 +34,11 @@ end
 ---@return table
 local function handlers(adapter)
   local model_opts = resolve_model_opts(adapter)
+  local current_url = adapter.url or "https://api.githubcopilot.com/chat/completions"
+  local base_url = current_url:gsub("/chat/completions$", ""):gsub("/responses$", "")
+
   if model_opts.endpoint == "responses" then
-    adapter.url = "https://api.githubcopilot.com/responses"
+    adapter.url = base_url .. "/responses"
 
     local responses = require("codecompanion.adapters.http.openai_responses")
 
@@ -74,7 +77,7 @@ local function handlers(adapter)
     return responses.handlers
   end
 
-  adapter.url = "https://api.githubcopilot.com/chat/completions"
+  adapter.url = base_url .. "/chat/completions"
   return require("codecompanion.adapters.http.openai").handlers
 end
 
--- a/lua/codecompanion/adapters/http/copilot/stats.lua
+++ b/lua/codecompanion/adapters/http/copilot/stats.lua
@@ -27,8 +27,17 @@ local function get_statistics()
 
   local oauth_token = token.fetch({ force = true }).oauth_token
 
+  local host = vim.env.GH_HOST or "github.com"
+  local endpoint
+  if host == "github.com" then
+    endpoint = "https://api.github.com/copilot_internal/v2/token"
+  else
+    -- GitHub Enterprise usually puts the API under /api/v3
+    endpoint = string.format("https://%s/api/v3/copilot_internal/v2/token", host)
+  end
+
   local ok, response = pcall(function()
-    return Curl.get("https://api.github.com/copilot_internal/user", {
+    return Curl.get(endpoint, {
       sync = true,
       headers = {
         Authorization = "Bearer " .. oauth_token,
--- a/lua/codecompanion/adapters/http/copilot/token.lua
+++ b/lua/codecompanion/adapters/http/copilot/token.lua
@@ -104,7 +104,7 @@ local function get_oauth_token()
 
       userdata = vim.json.decode(userdata)
       for key, value in pairs(userdata) do
-        if string.find(key, "github.com") then
+        if string.find(key, vim.env.GH_HOST or "github.com") then
           return value.oauth_token
         end
       end
@@ -136,8 +136,17 @@ local function get_copilot_token()
   _token_fetch_in_progress = true
   log:trace("Authorizing GitHub Copilot token")
 
+  local host = vim.env.GH_HOST or "github.com"
+  local endpoint
+  if host == "github.com" then
+    endpoint = "https://api.github.com/copilot_internal/v2/token"
+  else
+    -- GitHub Enterprise usually puts the API under /api/v3
+    endpoint = string.format("https://%s/api/v3/copilot_internal/v2/token", host)
+  end
+
   local ok, request = pcall(function()
-    return Curl.get("https://api.github.com/copilot_internal/v2/token", {
+    return Curl.get(endpoint, {
       headers = {
         Authorization = "Bearer " .. (M._oauth_token or ""),
         Accept = "application/json",


olimorris:
I think it would be cleaner and better in the long run to have these URLs stored in the `env` table. These are technically not public APIs so subject to change at any point

AmadeusK525:
@olimorris I agree, although I really don't have the time to tackle this task. But I mean, this at least *works* for now, without this workaround it doesn't (from my tests), but I guess It's okay if you don't want to merge this, it doesn't seem like this feature has too much demand :)

olimorris:
@AmadeusK525 this is still a valid PR imo and something the community would benefit from.

Appreciate you're time constrained however this could be given to an LLM to finish off. My only ask was that users have a way of overwriting the base and token URLs in their config.

npertschy:
Huge thanks to @AmadeusK525 
I ran into the same situation, where my company decided to switch to GHE. So I came across this solution. At least for my setup, there is still a minor issue left. When parsing the json to extract the token, the string.find fails for the GHE-style json, because the key to look for is not only the host but also has an ID-string appended to it. 
Extending the string.find(key, pattern, 1, true) like this fixes the issue for me

npertschy:
Today I took some time with Copilot and worked out a solution for this issues. I introduced a github_enterprise_url to copilot.env. I would like to open a PR for this. How can I do this?

lua/codecompanion/adapters/http/copilot/stats.lua
```diff
@@ -27,8 +27,17 @@ local function get_statistics()
 
   local oauth_token = token.fetch({ force = true }).oauth_token
 
+  local host = vim.env.GH_HOST or "github.com"
+  local endpoint
+  if host == "github.com" then
+    endpoint = "https://api.github.com/copilot_internal/v2/token"
+  else
+    -- GitHub Enterprise usually puts the API under /api/v3
+    endpoint = string.format("https://%s/api/v3/copilot_internal/v2/token", host)
```
lh-ak, line 36:
```suggestion
    endpoint = string.format("https://%s/api/v3/copilot_internal/user", host)
```

lua/codecompanion/adapters/http/copilot/stats.lua
```diff
@@ -27,8 +27,17 @@ local function get_statistics()
 
   local oauth_token = token.fetch({ force = true }).oauth_token
 
+  local host = vim.env.GH_HOST or "github.com"
+  local endpoint
+  if host == "github.com" then
+    endpoint = "https://api.github.com/copilot_internal/v2/token"
```
lh-ak, line 33:
```suggestion
    endpoint = "https://api.github.com/copilot_internal/user"
```

this seems to be a copy paste error, the path was "/copilot_internal/user" not "*/token" it's for stats not to request a token.

lua/codecompanion/adapters/http/copilot/init.lua
```diff
@@ -34,8 +34,11 @@ end
 ---@return table
 local function handlers(adapter)
   local model_opts = resolve_model_opts(adapter)
+  local current_url = adapter.url or "https://api.githubcopilot.com/chat/completions"
+  local base_url = current_url:gsub("/chat/completions$", ""):gsub("/responses$", "")
```
lh-ak, line 38: do I miss something here or is this basically enough if we expect base url in adapter.url

```suggestion
  local base_url = adapter.url or "https://api.githubcopilot.com"
```

lua/codecompanion/adapters/http/copilot/init.lua
lh-ak, line 102:
```suggestion
  url = "https://api.githubcopilot.com",
```

lh-ak:
Was so free to add some suggestions, feel free to ignore them.

This is something I am looking forward to be merged, thank you for the work :)

olimorris:
You all are far more knowledgable when it comes to Copilot and how it authorizes users. I am still very keen to merge this based on your feedback.

All I ask is that we give consideration to how users can override variables. I'm conscious some people use Copilot via an enterprise setup and through VPNs.

AmadeusK525:
Sorry guys, I switched jobs and in my new job Copilot is not provided via Enterprise, so I don't have an easy way to test this anymore (and it's not something I currently need). @lh-ak feel free to take over the PR (or open a new one that superseeds this one)

lh-ak:
I looked into it and after `copilot_acp` I personally don't see any value in adding that feature, is there anyting that http version can do that acp version can't?

npertschy:
I prepared a viable solution to this problem, but I am not able to push or contribute yet. I would love to.

Here is the summary of the changes I propose

Allow the Copilot adapter to authenticate against a GitHub Enterprise
instance by setting `github_enterprise_url` in the adapter's `env`
table. Falls back to the `GH_HOST` environment variable, then
`github.com`.

- Add `resolve_endpoints()` to token module, returning the correct
  token, auth hostname, and stats endpoints for both public GitHub
  and GHE instances
- Pass the adapter through all `token.fetch()` / `token.init()` call
  sites so endpoint resolution has access to the configured host
- Fix `handlers()` in the adapter to derive the base URL dynamically
  rather than hardcoding `api.githubcopilot.com`, so GHE-issued
  endpoint URLs are preserved across chat/completions and responses
  endpoint switches
- Fix `stats.lua` to use the stats endpoint (3rd return value of
  `resolve_endpoints`) instead of the token endpoint
- Invalidate the module-level OAuth/Copilot token cache when the
  target host changes, preventing a cached public-GitHub token from
  being reused for GHE requests and vice versa
- Add six tests covering `resolve_endpoints` URL construction,
  protocol stripping, `GH_HOST` fallback, and cache invalidation
- Document the `github_enterprise_url` option in adapters-http.md

olimorris:
At this point, I'm thinking it's easier if someone just puts this in a gist or a separate repo so others can use it.

npertschy:
I‘m a little bit disappointed. This feature seems to be really no big deal, neither usability- nor featurewise. 
Currently, I am living with a patched version of the of codecompanion, but I need to undo and reapply the changes for every regular update. Kind of inconvenient. 
It‘s not the issue, that I insist on contributing myself, but I really like codecompanion and would like to have a convenient setup for my working environment

616b2f:
> I‘m a little bit disappointed. This feature seems to be really no big deal, neither usability- nor featurewise. 
> Currently, I am living with a patched version of the of codecompanion, but I need to undo and reapply the changes for every regular update. Kind of inconvenient. 
> It‘s not the issue, that I insist on contributing myself, but I really like codecompanion and would like to have a convenient setup for my working environment 

You can still use copilot, just use 'copilot_acp' adapter, it works fine with GHE. So there is a more robust solution already available.

npertschy:
Thanks for the suggestion, but that would involve another tool and I currently don't like that approach. I decided to setup a adapter plugin myself. Have a look if you like
https://github.com/npertschy/copilot-ghe.nvim

PR #2718, opened 2026-02-03
Prevent bad tool calling messages from locking the chat session

slg95:
## Description

If you attempt to send a message in-between a tool call message from the LLM and the result of that tool call, it creates an invalid message history and the LLM may refuse to send a reply. I know this is a problem with Claude, and all the other major LLMs also claim this is not supported (haven't verified with any other provider). This stops you from adding subsequent messages in the chat session since the issue is that the messages need to be reordered. This can be fixed through the debug window, but its a pain to try to do so.

## AI Usage

Claude Opus helped with research, and Claude Sonnet did the implementation. Sonnet did all of the coding work, but I iterated over the result a few times and reviewed the final output.

## Checklist

### DIFF
--- a/tests/interactions/chat/tools/builtin/test_tool_output.lua
+++ b/tests/interactions/chat/tools/builtin/test_tool_output.lua
@@ -43,6 +43,12 @@ function tool_call(c)
         type = "function",
       },
     }
+
+    -- Add the tool call message to chat.messages (as done() would do)
+    _G.chat:add_message({
+      role = "assistant",
+      tool_calls = { _G.tool.function_call },
+    }, { visible = false })
   ]])
 end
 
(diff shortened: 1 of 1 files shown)


olimorris:
Can you share some detail on where and how this is happening? The buffer is locked so prevents any input from being sent.

I'd rather get to the root cause than patch with code.

slg95:
> Can you share some detail on where and how this is happening? The buffer is locked so prevents any input from being sent.
> 
> I'd rather get to the root cause than patch with code.

Of course. I have two videos, before and after the patch. (sorry they are so short, I figure you can pause as needed. GItHub has a max of 10MB for file uploads)

Before:

https://github.com/user-attachments/assets/0a96a07c-8ab2-43b3-8355-bdbe58b4a9e8

After:

https://github.com/user-attachments/assets/ad56f975-bed8-4aa8-890e-8dd172a62b95

Since you can send messages while the buffer is locked, sending an empty message causes the chat window to be unable to fix itself. So, another solution could be to prevent empty messages or prevent sending messages until the tool call result message is added to the message table.

PR #2733, opened 2026-02-06
fix(adapters): prevent openai image messages from being merged as text

petobens:
## Description

When adding multiple images via `/image`, each image is stored as its own hidden user message with `_meta.tag = "image"` and `content = <base64>`. The `build_messages()` code tries to “combine image + next user text”, but it does not check whether the “next user message” is actually another image. So with 3 images in a row, it does this:

- image1 gets combined with “text” = base64(image2), image2 is skipped
- image3 becomes the only remaining real image message
- result, the model effectively receives 1 or 2 images, not all

This PR implements a minimal fix: only combine with the next message if the next message is not an image-tagged message.

## AI Usage

gpt-5.2 mostly to trace the bug around the code.

## Related Issue(s)

#2732 

## Checklist

### DIFF
--- a/tests/adapters/http/test_openai_responses.lua
+++ b/tests/adapters/http/test_openai_responses.lua
@@ -153,6 +153,74 @@ T["Responses"]["build_messages"]["images"] = function()
   h.eq(expected, adapter.handlers.request.build_messages(adapter, messages))
 end
 
+T["Responses"]["build_messages"]["multiple consecutive images are not merged as text"] = function()
+  local messages = {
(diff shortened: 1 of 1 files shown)


olimorris:
Fantastic catch, thank you. Could we add a simple test case for this? Doesn't have to be actual base64 encoding we test but keen to make sure future adapter edits don't undo this.

petobens:
Added a test. Dunno if it was what you had in mind but happy to modify it further if needed.

Issue #2789, opened 2026-02-21
[Bug]: OpenAI strict mode rejects MCP tools with nested object schemas — missing additionalProperties: false


bahaaza:
## Pre-submission checklist

- [x] I have read the [documentation](https://codecompanion.olimorris.dev)
- [x] I have updated the plugin and all dependencies to the latest versions
- [x] I have searched for existing [issues](https://github.com/olimorris/codecompanion.nvim/issues) and [discussions](https://github.com/olimorris/codecompanion.nvim/discussions)
- [x] My issue is not a minor or cosmetic quirk (e.g., formatting, spacing, or other non-functional details)

## Neovim version (nvim -v)

v0.12.0

## Operating system/version

macOS

## Adapter and model

Copilot (OpenAI models via GitHub Models)

## Describe the bug

When using OpenAI-compatible models (Copilot, GitHub Models) with MCP servers that have tools containing **nested object schemas** (e.g., [`@modelcontextprotocol/server-memory`](https://github.com/modelcontextprotocol/servers)), the API returns a **400 error** because `additionalProperties: false` is not set on nested object definitions.

OpenAI requires `additionalProperties` to be supplied and set to `false` on **every** object schema in a function/tool definition — not just the top-level `parameters` object.

This is a continuation of the same class of issue fixed in [PR #2334](https://github.com/olimorris/codecompanion.nvim/pull/2334) (related to [#2314](https://github.com/olimorris/codecompanion.nvim/issues/2314)), which added recursive handling for `required` arrays and `null` types but did **not** add `additionalProperties: false` to nested objects.

| OpenAI strict mode requirement | Handled by PR #2334 | Status |
|---|---|---|
| `required` must include every key in `properties` | `make_all_properties_required` | ✅ Fixed |
| All property types must include `null` | `make_all_properties_required` | ✅ Fixed |
| `additionalProperties: false` on all objects | Not handled | ❌ Missing |

### Error

```
[ERROR] 2026-02-22 01:24:59
Error: {"error":{"message":"Invalid schema for function 'knowledge_graph_memory__create_entities':
In context=('properties', 'entities', 'type', '0', 'items'),
'additionalProperties' is required to be supplied and to be false.",
"code":"invalid_request_body"}}

[ERROR] 2026-02-22 01:24:59
[chat::_submit_http] Error: {
[...]

Issue #2314, opened 2025-10-28
[Bug]: Copilot w/ GPT-5 series models fails with "Invalid schema" for several tools


cairijun:
### Pre-submission checklist

- [x] I have read the [documentation](https://codecompanion.olimorris.dev)
- [x] I have updated the plugin and all dependencies to the latest versions
- [x] I have searched for existing [issues](https://github.com/olimorris/codecompanion.nvim/issues) and [discussions](https://github.com/olimorris/codecompanion.nvim/discussions)
- [x] My issue is not a minor or cosmetic quirk (e.g., formatting, spacing, or other non-functional details)

### Neovim version (nvim -v)

v0.12.0-dev-1454+gcdc3702f8d

### Operating system/version

Linux

### Adapter and model

Copilot GPT-5 / GPT-5-mini / GPT-5-Codex

### Describe the bug

When using Copilot GPT-5 series models, some tools will fail with errors like `Invalid schema for function 'file_search': In context=(), 'required' is required to be supplied and to be an array including every key in properties.`. At lease these tools are affected: `file_search`, `get_changed_files`, `grep_search`.

Other models are not affected, including GPT-4.1, Claude-*, Gemini. I do not have an openai account so I didn't test the openai adapter.

The OpenAI docs says ["To use Structured Outputs, all fields or function parameters must be specified as required"](https://platform.openai.com/docs/guides/structured-outputs/supported-schemas#all-fields-must-be-required). Maybe this is revelant.

### Steps to reproduce

1. `nvim --clean -u minimal.lua`
2. `:CodeCompanionChat copilot`
3. `gd`, change `model` to `gpt-5`/`gpt-5-mini`/`gpt-5-codex`, then Ctrl-S
4. Submit the followings:
    ```text
    @{files}
    Who are you?
    ```

### Expected behavior

All tools should work normally.

### Screenshots or recordings (optional)

<img width="1308" height="866" alt="Image" src="https://github.com/user-attachments/assets/fa3637d7-3868-4c5f-bbd4-13aa8632110e" />

### minimal.lua file

```lua
https://github.com/olimorris/codecompanion.nvim/blob/6fe6b45b2f3b2e856801cba29d24ecc1bf2f8941/minimal.lua
```

### Log output (optional)

```log
[INFO] 2025-10-28 17:19:52
Chat request started
[INFO] 2025-10-28 17:19:52
Request body file: /data/tmp/nvim.8wZwft/3.json
[...]

PR #2790, opened 2026-02-22
fix(tools): add additionalProperties: false recursively in enforce_strictness

bahaaza:
## Description

Extend `make_all_properties_required` to set `additionalProperties = false` on all nested object schemas when enforcing OpenAI strict mode.

OpenAI requires `additionalProperties: false` on **every** object schema in function/tool definitions, not just the top-level `parameters` object. The current implementation (added in PR #2334) correctly recurses into nested objects for `required` arrays and `null` types, but does not set `additionalProperties = false` on them.

This causes 400 errors when MCP servers (e.g. `@modelcontextprotocol/server-memory`) provide tools with nested object schemas like `create_entities`, whose `entities` array contains item objects with `properties`.

The fix adds `additionalProperties = false` at three points in `make_all_properties_required`:
- On array items with object type
- On nested object properties
- On the current object itself

This only affects adapters that opt into `strict_mode = true` (Copilot, OpenAI Responses).

## AI Usage

AI was used to help trace the schema flow through the codebase, identify the root cause, and draft the issue write-up. The fix itself (3 lines).

## Related Issue(s)

- #2789
- #2314 (original strict mode issue)
- PR #2334 (added `make_all_properties_required` but missed `additionalProperties`)

## Screenshots

N/A

## Checklist

- [x] I've run `make all` to ensure docs are generated, tests pass and StyLua has formatted the code
- [x] I've added test coverage for this fix/feature


olimorris:
Thanks for this. I've changed the base to `develop` as I will be merging this in the coming days and will test this out later today

PR #2799, opened 2026-02-24
fix(tools): respect function.strict = false in to_new_openai

bahaaza:
## Description

When a tool schema sets `strict = false` on the `["function"]` key (e.g. mcphub's `use_mcp_tool`), `to_new_openai` ignored it because it only checked `parameters.strict`. This caused `enforce_strictness` to be applied even when the tool explicitly opted out.

For tools with dynamic schemas like `tool_input` (an object with no predefined `properties` but `additionalProperties = false`), forcing strict mode made the schema reject all properties — the LLM would send empty `tool_input: {}` or the API would reject the schema entirely, resulting in **silent tool failure**.

### Root Cause

```lua
-- Before: only checked parameters.strict (nil for mcphub tools)
if opts.strict_mode and not schema["function"].parameters.strict then
    schema = M.enforce_strictness(schema)
end
```

mcphub sets `strict = false` on `schema["function"].strict`, not on `schema["function"].parameters.strict`, so the opt-out was never seen.

### Fix

Resolve the tool's strict preference by checking `function.strict` first, falling back to `parameters.strict`. Enforcement is skipped only when the resolved value is explicitly `false`:

```lua
local tool_strict = schema["function"].strict
if tool_strict == nil then
    tool_strict = schema["function"].parameters.strict
end
if opts.strict_mode and tool_strict ~= false then
    schema = M.enforce_strictness(schema)
end
```

| `function.strict` | `parameters.strict` | `strict_mode` | Enforce? |
|---|---|---|---|
| `nil` | `nil` | `true` | ✅ Yes |
| `true` | any | `true` | ✅ Yes |
| `false` | any | `true` | ❌ No |
| any | any | `false` | ❌ No |

## AI Usage

CodeCompanion (Claude) was used to help trace the execution path and draft the fix. The core analysis, root cause identification, and fix logic were directed by me.

## Related Issue(s)

Fixes silent tool failure when using mcphub.nvim with the OpenAI Responses adapter (`strict_mode = true`).

## Screenshots

N/A — logic-only change, no UI impact.

## Checklist


olimorris:
Closing as this doesn't adhere to the PR template and doesn't declare the AI usage

PR #2808, opened 2026-02-26
feat(adapters): add formatting acp tool calls

cenk1cenk2:
## Description

- This adds the tool output mechanism to `acp` adapters just like the `http` adapters.
- Since we have no real control over tool calls here and `acp` adapters can call multiple tools at the same time and also have updates, it does not work exactly the same way and instead we keep a map of called tools and as they finalize relog the tool call.

**There is a tiny problem with this fold mechanism here, even though the folds are created perfectly sometimes the endline hangs after the fact and I can not for the life of me figure out why. Just have my suspicions on extmarks since it was never reliable with `acp` adapters.** I can fix this issue with some guidance, if you find this PR worthy of something that can be merged. I think this happens because of the manual folding and having no line at the end, when the new line is appended the fold point shifts somehow, so currently it just adds an empty line in between.

## AI Usage

- Claude Code - Opus
- Codex - GPT 5.3 / xhigh

To figure out the formatting chunk complying with the `acp` specification.

## Screenshots

image
image

## Checklist


tests/screenshots/tests-interactions-chat-tools-builtin-test_tool_output.lua---Tool-output---Folds---can-be-folded
```diff
@@ -10,8 +10,8 @@
 09|                                        │~                                      
 10|!! **Weather Tool**: Ran successfully:··│~                                      
 11|                                        │~                                      
-12|Let me know if you need anything else!  │~                                      
-13|~                                       │~                                      
+12|                                        │~
```
olimorris, line 13: This is less optimal than what we have now, surely?

olimorris:
Hi @cenk1cenk2, I'm unclear what this PR contributes and what the problem statement is. In a worse case scenario, we're adding a new handler to every single ACP adapter and presumably asking users who have their own custom ACP adapters to do the same?

cenk1cenk2:
Hey @olimorris,

Thank you for your time and willingness to understand.

This PR just adds the ability to show tool outputs from the ACP adapter in a fold, just like the HTTP adapters.

So currently it just shows call result like below.

```
Tool: mcp_create_issue
```

Now it behaves more like below.

```
Tool: mcp_create_issue -> actual call
Tool: mcp_create_issue -> when finished in fold
issue created
url: https://...
```

Most of the agents assume that we see the output of the tool calls, so it usually needs to be prompted like "Can you give me the url of the created issue` and such. This way we can just fetch it from the tool result.

tests/screenshots/tests-interactions-chat-tools-builtin-test_tool_output.lua---Tool-output---Folds---can-be-folded
cenk1cenk2: Yeah it is much less than optimal, but somehow when a new line is added, the fold gets broken. I did not also understand it quite good because the folding mechanism works amazingly well.

olimorris:
One of my concerns here is that the output from an agent can be huge. So we'd need to think how that is handled.

I'll think about this as a PR for inclusion. I defo think there's work to be done to make tool calls from ACP agents a little nicer.

cenk1cenk2:
Great news, thanks.

You are definetely right because with longer conversations the buffer will be full of these stuff which will make things much slower to work with.

Whenever you have any feedback on how this can be improved in your eyes, I would be happy to revise as you wish.

PR #2849, opened 2026-03-04
feat(ui): revise confirmation dialog to be non-blocking

cenk1cenk2:
## Description

- Changes the confirmation dialog to be async with a floating window instead of the confirm UI, so whenever tool requests come in through the ACP, it would not block the UI. Currently, whenever there are multiple tool calls spawned, and there is no orchestration of tasks with ACP adapters, the UI hangs, so you do not actually know what you are agreeing to.
- The spawned confirmation dialog will have more information about the arguments that the tool is getting called. Sometimes, utilizing a tool from the MCP adapter attached, you might want to investigate the call in detail just to make sure everything is as expected.

## AI Usage

- Claude Opus 4.6

It is used to design the floating window interface. It is also used to ensure that there are no holes in the async mechanism, where blocking might be needed and testing.

## Screenshots

image

## Checklist


olimorris:
Hey @cenk1cenk2 

Appreciate you taking the time to put together a PR. I recognise that `vim.fn.confirm` isn't ideal for a lot of users but this represents a huge change that I'll have to think about. Right now, I broadly agree with this but will give it a couple of weeks to think it through whilst I work on the roadmap.

cenk1cenk2:
Thank you for your time yet again to consider it.

olimorris:
FYI, #2917 will address this.

PR #2885, opened 2026-03-12
fix(adapters): handle non-stream content with reasoning for responses

xhebox:
format

## Description

Output text is not necessarily the first block. Deepseek-reasoner met this.

## Checklist

### DIFF
--- a/tests/adapters/http/test_openai_responses.lua
+++ b/tests/adapters/http/test_openai_responses.lua
@@ -536,16 +536,12 @@ T["Responses"]["No Streaming"]["can process reasoning output"] = function()
   -- Match the format of the actual request
   local json = { body = data }
 
-  h.expect_contains(
-    "**Choosing descriptive terms**",
-    adapter.handlers.response.parse_chat(adapter, json).output.reasoning.content
-  )
+  local output = adapter.handlers.response.parse_chat(adapter, json).output
 
-  h.eq(
-    "rs_0a10a8c968d594670168e91d0204ac8195b26b3e4de997f65c",
-    adapter.handlers.response.parse_chat(adapter, json).output.reasoning.id
-  )
-  h.eq("gAAAAABo6", adapter.handlers.response.parse_chat(adapter, json).output.reasoning.encrypted_content)
+  h.expect_contains("**Choosing descriptive terms**", output.reasoning.content)
+  h.eq("rs_0a10a8c968d594670168e91d0204ac8195b26b3e4de997f65c", output.reasoning.id)
+  h.eq("gAAAAABo6", output.reasoning.encrypted_content)
+  h.eq("Dynamic expressive", output.content)
 end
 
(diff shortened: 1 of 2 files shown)


olimorris:
Can you share what model you've encountered this with? The PR description is lacking detail.

xhebox:
> Can you share what model you've encountered this with? The PR description is lacking detail.

Deepseek behind an LLM gateway called axonhub. But I think it is general, the fake test case is exactly same as the deepseek api. However it did not test the content at all.

EDIT: openai  has the similar example too, https://developers.openai.com/api/docs/guides/migrate-to-responses. I met the problem when I use `/compact`, which is non-streaming.

tests/adapters/http/test_openai_responses.lua
```diff
@@ -536,16 +536,12 @@ T["Responses"]["No Streaming"]["can process reasoning output"] = function()
...
 
-  h.eq(
-    "rs_0a10a8c968d594670168e91d0204ac8195b26b3e4de997f65c",
-    adapter.handlers.response.parse_chat(adapter, json).output.reasoning.id
-  )
-  h.eq("gAAAAABo6", adapter.handlers.response.parse_chat(adapter, json).output.reasoning.encrypted_content)
+  h.expect_contains("**Choosing descriptive terms**", output.reasoning.content)
+  h.eq("rs_0a10a8c968d594670168e91d0204ac8195b26b3e4de997f65c", output.reasoning.id)
+  h.eq("gAAAAABo6", output.reasoning.encrypted_content)
+  h.eq("Dynamic expressive", output.content)
```
xhebox, line 544: FYI, sample data is not fully tested. Before the change, `output.content` is nil, while it is "Dynamic expressive" actually. The first block can be `item.type == reasoning`.

@olimorris

olimorris: I have no idea what you mean by "not fully tested".

I haven't tried this PR out yet but will run it through OpenAI's models at some point this weekend.

xhebox: > I have no idea what you mean by "not fully tested".
> 

mock data `tests/adapters/http/stubs/openai_responses_inline.txt")` has two items, `type = reasoning` for the first, `type = message` for the second.
We tested the reasoning content but not the message content in this test. It would not pass if we do check the message content.

olimorris:
Sorry for not coming back to this sooner. Can you give me a set of steps to test this?

xhebox:
> Sorry for not coming back to this sooner. Can you give me a set of steps to test this?

open a chat buffer with reasoning-enabled model(like deepseek-reasoner), make some random chats, then `/compact`.

Bug will be trigerred if the response has both reasoning and normal text content.

xhebox:
Any news? I've tested it on a lot of other APIs like nvidia or GLM.

PR #2889, opened 2026-03-16
Fix acp and slash commands match patterns

alexghergh:
Fixes an issue where acp commands (i.e. \ commands) would error and not be correctly shown in chat.

Also fixes cmp completion for \ commands, and allows '-' as a match character for both acp and slash commands, which allows commands like \review-branch or /document-code to correctly show in the cmp completion window.

## Description

Previously, typing \ in the chat window would correctly open the cmp completion window, but because the regex was broken, it would match a literal `\w` instead of matching a regex `word` pattern.

Also adds a match for patterns such as \review-branch or /init-readme or other acp / slash commands with hyphens. From my experience, other characters (such as `_`) don't appear in e.g. codex / claude code anyway, so those are not matched. This fixes an issue with nvim-cmp that would not match those patterns correctly before, and therefore `source:execute()` would not be triggered.

## AI Usage

AI authored the PR. I manually checked the implementation.

## Checklist

### DIFF
--- a/lua/codecompanion/providers/completion/cmp/acp_commands.lua
+++ b/lua/codecompanion/providers/completion/cmp/acp_commands.lua
@@ -18,8 +18,8 @@ function source:get_trigger_characters()
 end
 
 function source:get_keyword_pattern()
-  local escaped = vim.pesc(trigger)
-  return escaped .. [[\w\+]]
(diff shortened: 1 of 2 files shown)


alexghergh:
Not sure about tests for this new functionality. Would these changes warrant new tests?

PR #2898, opened 2026-03-18
feat(tools): memory can have whitelisted paths

olimorris:
## Description

The memory tool enables users to save their conversation history to disk so it can be resumed at a later date.

The tool follows Antropic's [specification](https://platform.claude.com/docs/en/agents-and-tools/tool-use/memory-tool) to the letter, even restricting the use of directories outside of `/memories`.

This PR relaxes that restriction, enabling users to white list paths and files that memory files can be saved in.

## AI Usage

Opus 4.6

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -48,6 +48,7 @@ doc/.vitepress/cache
 doc/tags*
 doc/node_modules
 
+.codecompanion.lua
 .direnv/
 .testenv/
 .repro/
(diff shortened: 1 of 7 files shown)


powerman:
It looks like Claude Code itself uses `~/.claude/projects/-home-USER-PATH-TO-PROJECT/memory/` as target directory for `/memories`. Is it possible to configure CodeCompanion in a same way, so both tools will share same memories for same project?

olimorris:
Yep. You'd whitelist that directory.

I recall those memory files have random names as opposed to anything sensible so there's a lot of context that would be required to be processed before it finds the correct memory file.

For Claude, I believe it manages this by using sub-agents. Something that CodeCompanion is yet to implement but is definitely on my roadmap.

PR #2910, opened 2026-03-20
feat(acp): add support for `session/list` and `session/load`

olimorris:
## Description

Now that `session/list` is officially part of the ACP specification, we can support in CodeCompanion.

If you're in a chat buffer with an ACP adapter, the `/resume` slash command will be available to you and you can select a session to load.

Does not appear that there is a way to rename session from Claude Code.

## AI Usage

Opus 4.6 for general reviewing of the spec and how it could be implemented in the current ACP class.

## Related Issue(s)

#2313

## Checklist


mrjones2014:
Does this mean the `codecompanion-history.nvim` plugin essentially isn't needed if you only use ACP adapters now?

olimorris:
@mrjones2014 I don't know how that plugin works, but it looks like it doesn't support the protocol at all. So I'd be surprised if it was effectively listing and loading ACP sessions.

Whether this works and is available to you will depend on whether your agent CLI supports it. This was only added to the spec a couple of weeks ago.

alexghergh:
@mrjones2014 Correct, you only need that plugin for HTTP / API calls, *not* Claude Code or Codex (or other ACP agents). The agents hold their own history on disk (see `~/.codex/sessions` for codex or `~/.claude/projects` for claude).

In practice, you can keep it if you occasionally want to do HTTP calls, and save those sessions manually (from the plugin's README.md: `sc - Save current chat manually (customizable via opts.save_chat_keymap)`), but you do want it disabled for ACP calls.

olimorris:
@iamthebot - Slash commands are working for me on restore:

2026-03-25 07_19_16 - Ghostty@2x

 and in a brand new, clean chat buffer:

2026-03-25 07_21_02 - Ghostty@2x

I hear you regarding the diff. I forgot to add in a check to see if you're active in the chat buffer and then display the diff. I will also inline the diff into the chat buffer at some point this week.

alexghergh:
@olimorris One thing this PR probably breaks is token usage for restored messages. Since the [session usage update](https://agentclientprotocol.com/rfds/session-usage) ACP call isn't yet standardized, and CodeCompanion doesn't yet implement that parsing update (even though most ACP clients including Codex, Claude Code and Gemini include it, as off-spec for now), there is no way to count tokens. You *could* do that through e.g. `tokens.calculate()`, but that only works if you have `chat.add_message()` called on the restored session messages, or if you manually parse the `chat.bufnr` variable (the visible transcript).

What I did before this PR is implement my own `session/load` wrapper, which does exactly that. Of course that doesn't quite work **anyway** because it's not a true estimate (doesn't count the prompt text, other stuff like AGENTS.md etc.).

Another workaround is: until the ACP usage update call gets standardized users can manually parse those updates and display them somehow in the chat, but otherwise there's no way to count tokens for the resumed session internally.

I do understand if you expect the token usage update to get standardized, and then later implement that usage parsing logic anyway, though. Just pointing this out.

alexghergh:
Also, what do you think about also firing a `CodeCompanionChatDone` event at the end of replay? That way, users can hook into it normally for parsing / displaying autocommands that should automatically run on LLM response.

olimorris:
> @olimorris One thing this PR probably breaks is token usage for restored messages.

I'm not sure what you mean. CodeCompanion never sends the historic conversation back to the agent. The agent holds the state so only new messages are persisted. So there's no issue of additional token usage. Or, are you referring to the displaying of tokens?! Because that was never part of the ACP spec and so I haven't implemented it.

olimorris:
> Also, what do you think about also firing a CodeCompanionChatDone event at the end of replay?

I don't know what you mean by "replay". At the end of `session/load`? There's a case for a new `CodeCompanionChatResumed` event, for sure.

alexghergh:
Yes, I realized I wasn't very clear indeed. I did mean token display. Right now it stays 0 for all ACP connections, because indeed it isn't standardized. My workaround is using `tokens.calculate` to estimate it internally for now.

Yes, by "replay" I mean the conversation history you receive back from the ACP connection. What I meant is indeed at  the end of `session/load`, it would be nice if there was a `CodeCompanionChatDone` event. A `CodeCompanionChatResumed` event is even better.

mrjones2014:
This is very epic, thank you for constantly making huge improvements!

olimorris:
@alexghergh shall we pick this up in a PR or another discussion? I'm absolutely keen to see if this can be supported

alexghergh:
@binronn It seems opencode supports `session/load`, so that's not a problem. Likely related to PR #2945, that fixes issues with session creation.

PR #2945, opened 2026-03-27
ACP: Separate connection and session creation logic

alexghergh:
This PR fixes some issues introduced from the PR that added `session/list` and `session/load`.

## Description

Implement proper separation between ACP connections and sessions for a number of code paths. Code paths that do not need a session (i.e. `session/list`; change adapter only, without changing model) do not create ACP sessions, while code that do require a session (`session/load`, `session/new`, `/mode`, change model) create a session as well.

Tighten ACP session state handling by clearing cached modes/models when `session/load` or `session/new` omit them, so stale session metadata is not reused across fallback paths (the ACP spec says that upon `session/new` and `session/load`, RPC calls MAY return model / models).

Properly expose slash commands on ACP connections, without requiring a session to be established (a session IS established on message submit or other paths that do need it, see above).

*NOTE*: Right now, ACP commands are broken on fresh chats, because there's no ACP session established (only an ACP connection). The spec mandates that ACP commands are session-bound. For now, this is still broken in this PR, but a subsequent PR will fix that too (still working on the implementation). Any command or interaction that establishes a session (i.e. submitting a user message, calling model selection picker, calling /mode etc.) WILL correctly populate the ACP commands as well, as a workaround for now.

## AI Usage

Codex usage + manual testing / refactoring.

## Related Issue(s)

#2948

## Checklist

### DIFF
--- a/lua/codecompanion/interactions/chat/acp/handler.lua
+++ b/lua/codecompanion/interactions/chat/acp/handler.lua
@@ -90,8 +90,7 @@ function ACPHandler:ensure_session()
   end
 
   -- Map bufnr -> session_id so completion providers can look up ACP commands for this buffer
-  local acp_commands = require("codecompanion.interactions.chat.acp.commands")
-  acp_commands.link_buffer_to_session(self.chat.bufnr, conn.session_id)
+  require("codecompanion.interactions.chat.helpers").link_buffer_to_acp_session(self.chat)
 
   return true
 end
--- a/lua/codecompanion/interactions/chat/helpers/init.lua
+++ b/lua/codecompanion/interactions/chat/helpers/init.lua
@@ -9,30 +9,39 @@ local M = {}
 local api = vim.api
 local fmt = string.format
 
----@param chat CodeCompanion.Chat
----@return CodeCompanion.Chat.ACPHandler
-local function acp_handler(chat)
-  return require("codecompanion.interactions.chat.acp.handler").new(chat)
-end
-
 ---Create a new ACP connection for the given chat
 ---@param chat CodeCompanion.Chat The chat instance
 ---@return boolean
 function M.create_acp_connection(chat)
-  return acp_handler(chat):ensure_connection()
+  return require("codecompanion.interactions.chat.acp.handler").new(chat):ensure_connection()
 end
 
 ---Ensure the chat has an ACP connection and session
 ---@param chat CodeCompanion.Chat The chat instance
 ---@return boolean
 function M.ensure_acp_session(chat)
-  local handler = acp_handler(chat)
+  local handler = require("codecompanion.interactions.chat.acp.handler").new(chat)
   if not handler:ensure_connection() then
     return false
   end
   if not handler:ensure_session() then
     return false
   end
+  return true
+end
+
+---Link the chat buffer to the active ACP session
+---@param chat CodeCompanion.Chat The chat instance
+---@return boolean
+function M.link_buffer_to_acp_session(chat)
+  if not chat.acp_connection or not chat.acp_connection.session_id then
+    return false
+  end
+
+  require("codecompanion.interactions.chat.acp.commands").link_buffer_to_session(
+    chat.bufnr,
+    chat.acp_connection.session_id
+  )
   return true
 end
 
--- a/lua/codecompanion/interactions/chat/init.lua
+++ b/lua/codecompanion/interactions/chat/init.lua
@@ -757,6 +757,9 @@ function Chat:change_model(args)
     self:set_system_prompt()
     self:apply_settings()
   elseif self.adapter.type == "acp" then
+    if not helpers.ensure_acp_session(self) then
+      return self
+    end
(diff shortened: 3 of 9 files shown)


olimorris:
Thanks for this. Super well thought out PR. I'll prioritize this and test it out later today.

alexghergh:
Hmm, there are still some rough edges on the PR. Mainly, it looks like ownership is quite spread across multiple locations, so updates are not easy. Especially chat metadata updates + buffer -> session linking would benefit most from a top-down ownership update I think, so fewer places are actually responsible for these changes. Right now, paths like /mode, /resume etc. update metadata by themselves on new session / mode / model updates.

For example, inside `ensure_acp_session` would be a good place to update chat metadata, since that initial session call (whether it's a `session/new` or `session/load`) will populate the mode and model. However, I wanted it to be a clearer split of responsibility. But it might be cleaner overall to actually include the metadata update there.

olimorris:
> Right now, ACP commands are broken on fresh chats

Do you mean ACP slash commands? Can you let me know how to recreate the issue you're facing?

alexghergh:
Yes, ACP slash commands. Just open a new chat, and try to type `\`. The completion will be empty. For example, claude exposes `\init`, `\compact` etc. These do not appear on a fresh chat. However, changing the mode, changing the model etc. will create an ACP session and correctly populate these ACP slash commands too.

alexghergh:
Also even if there's no completion (I'm using `nvim-cmp` by the way), even if you submit the command, it won't trigger the auto-conversion from `\init` to `/init`.

olimorris:
This is an ACP chat which has, say, Claude Code, as a default adapter?

alexghergh:
Correct, it needs to be the default adapter when you open a fresh chat, or otherwise you just need to obtain a chat _without_ a session, only an ACP connection.

If you cannot reproduce this, I can also upload a screencast.

olimorris:
I've opened an issue and a minimal.lua for this as it is indeed a bug.

I think this PR is doing too much. I expect the actual fix to be much leaner. I propose we focus on fixing this and then any sensible refactors I'd be happy to merge later on.

alexghergh:
You mean fix the ACP slash commands issue?

The problem is that it might be tied together with this PR, in the sense that it's only *correctly* solved once there's a clear separation between ACP connection creation and ACP session creation logic, since ACP slash commands are inherently bound to an ACP session. I'm not sure fixing the ACP slash commands in a separate PR would last long, if changes in this PR also get merged, since that code gets modified anyway.

Also, aside from the actual ACP connection / session logic split, this PR fixes the issue with `/mode` and `select_model()` code paths, which inherently require an ACP session. However, since a fresh chat doesn't create a session, **those code paths** need to create the session. So those should also be probably added as "bugs", since neither work correctly now in fresh chats (again, with e.g. a default adapter on a fresh chat, you don't get completion for `/mode`, and also trying to change a model through `select_model()` doesn't do anything, since that method itself fast exits because there's no session and it won't create one either).

There's also the problem of stale metadata that doesn't get updated on session creation, which also requires clearer boundaries between connection / session split, I think.

I do think this PR does at least 2 things, so indeed it may be too big: fix the issues with `/mode`, `select_model()` and ACP slash commands, **and** refactor into a better separation between connection and session logic. Would that be a better PR split? Or is there something else you're looking at for actual fixes?

alexghergh:
I've reflected a bit about the split into multiple PRs and here's the proposed path forward:
1. ACP session bootstrap primitives (helpers for `ensure_acp_session()`, `link_buffer_to_acp_session()`). Only refactoring, no bug fixes.
2. `/mode` for fresh ACP chats creates an ACP session. Fixes `/mode` unavailable in fresh chats.
3. `select_model()` / `change_model()` for fresh ACP chats create an ACP session. Fixes that code path being unavailable in fresh chats.
4. connection vs session separation (`change_adapter()` connection-only, `/resume` buffer relinking, `session/load` vs `session/new` scoping). Clearer separation of concerns. Refactoring, no bug fixes.
5. ACP slash commands bug fix.
6. metadata cache update / ownership cleanup, refactoring + stale metadata bug fix.

Some of the later parts depend on earlier changes, so take this as a dependency graph if you will. If 2 + 3 feel too small individually, I can combine them into one single PR.

Let me know what you envisioned would be the ideal fixes / split, since you mentioned you think the fixes might actually be leaner.

olimorris:
I've reviewed this PR and I still think it's doing too much. As in, I don't think it needs this level of refactor to fix the underlying issue. I've tagged you in a separate PR that proposes a leaner way of addressing this.

alexghergh:
Yeah, that's much cleaner indeed. The only obvious issue is still the separation between connection and session (which your PR doesn't have), but I'm not sure however how that affects users in practice. I assumed that establishing sessions like your PR does would pollute the resume list picker with empty sessions, but doesn't seem to be the case somehow?

I'll close this one out. Everything works now since the bugs were tied to session creation.

PR #2955, opened 2026-03-29
fix(background): pass full response to parse handler, not just body

ImmanuelHaffner:
## Description

Both ask_sync and ask_async were passing `response.body` (a raw JSON string) to the adapter's parse handler (e.g. `chat_output`).  In non-streaming mode the handler expects the full response table `{body, status, headers, …}` and internally does `data = data.body`. Passing the pre-extracted string caused `data.body` to resolve to nil, silently breaking all background interactions (e.g. auto-title generation never produced a title).

Pass the full `response` table so the handler can extract `.body` itself, consistent with how the main chat HTTP path works.

## AI Usage

Yes, Claude Opus 4.6 through CC ;)

## Screenshots

N/A

## Checklist

### DIFF
--- a/lua/codecompanion/interactions/background/init.lua
+++ b/lua/codecompanion/interactions/background/init.lua
@@ -96,7 +96,7 @@ local function ask_sync(background, messages, opts)
   end
 
   local parse_handler = opts.parse_handler or "parse_chat"
-  local result = adapters.call_handler(background.adapter, parse_handler, response.body)
+  local result = adapters.call_handler(background.adapter, parse_handler, response)
   return result, nil
 end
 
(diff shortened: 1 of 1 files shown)


olimorris:
This would potentially break user's configs as they're now expected to handle the body themselves. Given that would trigger a new major version of CodeCompanion, I'll hold off on this. Appreciate I may have done something dumb so I won't rule out changing this but given the limited feedback from people, I suspect this is okay for now. THank you for the PR though.

ImmanuelHaffner:
Sorry for necro-bumping this but i'd like to confirm. This was eventually implemented like i asked, right?

https://github.com/olimorris/codecompanion.nvim/blob/b0680d12ae1a7fc1b20ab8c78b8c44b34e72e245/lua/codecompanion/interactions/background/init.lua#L108-L116

---

https://github.com/olimorris/codecompanion.nvim/blob/b0680d12ae1a7fc1b20ab8c78b8c44b34e72e245/lua/codecompanion/interactions/background/init.lua#L139

PR #2964, opened 2026-04-01
fix: Update chat metadata after ACP session setup

alexghergh:
## Description

Fix metadata staleness after an ACP session is created. In practice, after an ACP session is created, model and mode remain stale because `ensure_acp_session` does not update those.

A chat metadata update _is_ called earlier after an ACP connection is established, however that is too early to update mode / model information as well.

## AI Usage

Codex

## Checklist


alexghergh:
Also, related to this PR, since the ACP session creation path is async (because it is called through a `vim.schedule()` call), it would be useful to have an event fired, such as `ACPSessionEstablished` or `ACPSessionCreated`, so that users can hook into it and read mode / model information after an ACP session is established. Even an event such as `ChatCreated` is too early, since it is fired on the sync path, likely before the session is created.

olimorris:
Thanks for this. #2978 included this as part of some broader ACP changes.

Issue #2969, opened 2026-04-02
[Bug]: Default model is not selected when using ACP


mintelm:
### Pre-submission checklist

- [x] I have read the [documentation](https://codecompanion.olimorris.dev)
- [x] I have updated the plugin and all dependencies to the latest versions
- [x] I have searched for existing [issues](https://github.com/olimorris/codecompanion.nvim/issues) and [discussions](https://github.com/olimorris/codecompanion.nvim/discussions)
- [x] My issue is not a minor or cosmetic quirk (e.g., formatting, spacing, or other non-functional details)

### Neovim version (nvim -v)

NVIM v0.11.6

### Operating system/version

Arch 6.6.114.1-microsoft-standard-WSL2

### Adapter and model

copilot_acp - any model

### Describe the bug

When I add a default model in the configuration,
I observe that it is not selected after startup, instead another one is selected.
I expect that the default model i chose in the configuration is selected.

### Steps to reproduce

1. open neovim with minimal config `nvim -u minimal.lua`
2. open codecompanion (`<leader>cc`)
3. press `gd` to open debug info
5. notice that another (not configured) model is used

### Expected behavior

I expected that the configured model is used and not "a random" one.

### Screenshots or recordings (optional)

Screenshot shows my debug info after starting. Notice that the adapter is working, and there's also a model present called `claude-opus-4.6`, which I configured to be used by default, but it is not, instead `claude-sonnet-4.6` is used.  

<img width="2456" height="305" alt="Image" src="https://github.com/user-attachments/assets/e4aac1c3-6187-4c27-a078-bf9005fe3fc1" />

### minimal.lua file

```lua
---@diagnostic disable: missing-fields

vim.env["CODECOMPANION_TOKEN_PATH"] = vim.fn.expand("~/.config")

vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

-- Your CodeCompanion setup
local plugins = {
    {
        "olimorris/codecompanion.nvim",
        dependencies = {
            { "nvim-lua/plenary.nvim" },
            {
                "nvim-treesitter/nvim-treesitter",
                lazy = false,
                build = ":TSUpdate",
            },
        },
        opts = {
            interactions = {
                chat = {
                    adapter = {
                        name = 'copilot_acp',
                        model = 'claude-opus-4.6',
                    },
                },
            },
        },
[...]

PR #2977, opened 2026-04-03
fix(acp): support session/set_config_option for models

olimorris:
## Description

Previously `session/set_model` was used to change the model on an adapter. However, this was always liable to change as it was in beta...and change it did. To `session/set_config_option`.

Source: https://agentclientprotocol.com/protocol/schema#sessionconfigoptioncategory

## AI Usage

Opus 4.6

## Related Issue(s)

#2969

## Checklist

### DIFF
--- a/lua/codecompanion/adapters/acp/init.lua
+++ b/lua/codecompanion/adapters/acp/init.lua
@@ -109,14 +109,12 @@ function Adapter.resolve(adapter, opts)
     adapter = Adapter.extend(config.adapters.acp[adapter] or adapter)
 
     if opts.model then
-      adapter = vim.tbl_deep_extend("force", adapter, {
-        defaults = { model = opts.model },
-      })
+      adapter.defaults = adapter.defaults or {}
+      adapter.defaults.model = opts.model
     end
     if opts.mode then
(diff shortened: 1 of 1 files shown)


alexghergh:
This seems to break setting reasoning. Previously I had something like this:

reason_old

But now I only have:

reason_new

It seems that ACP does support setting thinking level too (as per your link, `thought_level`).

This probably warrants a new method, `select_reasoning()` or something, and then chain it into the adapter workflow? (i.e. change adapter will do adapter + model + reasoning selection).

I do think that Claude Code doesn't have reasoning levels per se, so maybe that's why this wasn't considered.

olimorris:
Alas, `session/set_model` was never part of the ACP spec and any references to it have now been removed from their site.

As noted in their [announcement](https://agentclientprotocol.com/announcements/session-config-options-stabilized) in Feb, models, modes, reasoning all goes through Session Config Options now.

I'll add the ability to dynamically modify these in a future PR.

alexghergh:
Yeap, I totally get that, I'm definitely not saying "you should keep compatibility" or something. I figured that Claude Code doesn't have an actual temperature knob, so I surfaced this issue here since you might not have been aware about it.

PR #2997, opened 2026-04-07
fix(chat): update formatting for permission requests

cenk1cenk2:
## Description

I feel like I am starting to be annoying to you, so please tell me to stop if that is the case.

- Add a way to see more details on the incoming permission request where it is formatted as a document depending on the kind of the call.

## AI Usage

- Claude Opus for extracting this out of the other draft PR I try to do before.

## Screenshots

swappy-20260407_210350
swappy-20260407_210238

## Checklist

### DIFF
--- a/lua/codecompanion/interactions/chat/acp/request_permission.lua
+++ b/lua/codecompanion/interactions/chat/acp/request_permission.lua
@@ -248,6 +248,64 @@ local function approve_in_chat(permission, choices, prompt_opts)
   })
 end
 
+---Format the title and prompt body for a non-diff tool call approval
+---@param tool_call table|nil
+---@return string title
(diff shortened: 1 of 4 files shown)


olimorris:
I think this is a useful addition to the plugin. However, I'd prefer the formatting to sit within the tools themselves. This puts the tool owner in control of how it looks in the chat buffer - something that CodeCompanion should not be responsible for. Afterall, CodeCompanion just executes tools it doesn't know/care whether the content is a URL or a command.

We have a `prompt` output function in the tools themselves, can't we make better use of that?

Also, having the class that orchestrates the tools in the queue (`orchestrator.lua`) also handle formatting doesn't feel correct - these should be separated.

PR #3005, opened 2026-04-09
fix(gemini): map thinkingLevel to thinkingBudget for Gemini 2.5 models

maxronner:
## Description

This fixes Gemini reasoning config serialization for `gemini-2.5-*` models.

Previously, the adapter mapped `thinkingLevel` directly into `generationConfig.thinkingConfig`, which caused Gemini 2.5 requests to fail because those models expect `thinkingConfig` to contain a numeric `thinkingBudget`. This change moves `thinkingLevel` handling into `form_parameters`, where the adapter now builds the correct `thinkingConfig` shape for Gemini 2.5 models while preserving the existing `{ thinkingLevel = ... }` shape for other reasoning models, including Gemini 3.x.

A model-specific budget mapping is added for Gemini 2.5 Pro, Flash, and Flash-Lite. Serialization is also guarded so `thinkingConfig` is only added for reasoning models and only when `thinkingLevel` is present.

Test coverage was added for the Gemini 2.5 budget mappings, the existing Gemini 3.x `thinkingLevel` shape, non-reasoning models, preservation of existing `generationConfig` keys, and the absence of `thinkingConfig` when `thinkingLevel` is unset.

*Note: for `gemini-2.5-pro`, **`none`** maps to **`128`**, not `0`, because the API rejects `thinkingBudget = 0`.*

Docs were regenerated via `make all` per the project guidelines, otherwise untouched.

### References

[https://ai.google.dev/gemini-api/docs/thinking](https://ai.google.dev/gemini-api/docs/thinking)

## AI Usage

I used `claude-opus-4-6` to cross-check this change against the request-shaping patterns used in `lua/codecompanion/adapters/http/anthropic.lua` and to draft the test cases. I reviewed and revised those tests before including them. I also used `gpt-5-4` for wording and review feedback, but the implementation and final changes are mine.

## Related Issue(s)

[https://github.com/olimorris/codecompanion.nvim/discussions/2994](https://github.com/olimorris/codecompanion.nvim/discussions/2994)

## Screenshots

N/A

## Checklist

* [x] I've read the [contributing](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md) guidelines and have adhered to them in this PR
* [x] I confirm that this PR has been majority created by me, and not AI (unless stated in the "AI Usage" section above)
* [x] I've run `make all` to ensure docs are generated, tests pass and [StyLua](https://github.com/JohnnyMorganz/StyLua) has formatted the code
* [x] *(optional)* I've added [test](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md#testing) coverage for this fix/feature
* [ ] *(optional)* I've updated the README and/or relevant docs pages

### DIFF
--- a/lua/codecompanion/adapters/http/gemini.lua
+++ b/lua/codecompanion/adapters/http/gemini.lua
@@ -62,6 +62,34 @@ local function decode_args(args)
   return {}
 end
 
+---Gemini 2.5 models use `generationConfig.thinkingConfig = { thinkingBudget = N }`
+---rather than a raw thinking level. Models opt into this budget-based shape via
+---`opts.thinking_budget`, which selects a row from `THINKING_LEVEL_BUDGETS`.
+---Other reasoning models continue to use the `{ thinkingLevel = "<level>" }` shape.
+---
+---Gemini 2.5 Pro cannot fully disable thinking: the API rejects thinkingBudget = 0,
+---so `none` maps to 128 — the minimum valid budget — rather than a true off state.
+local THINKING_LEVEL_BUDGETS = {
+  pro = { none = 128, low = 1024, medium = 8192, high = 24576 },
+  flash = { none = 0, low = 1024, medium = 8192, high = 24576 },
+  flash_lite = { none = 0, low = 512, medium = 8192, high = 24576 },
+}
+
+local function resolve_thinking_config(opts, level)
+  local tier = type(opts.thinking_budget) == "string" and opts.thinking_budget
+  if not tier then
+    return { thinkingLevel = level }
+  end
+
+  local budgets = THINKING_LEVEL_BUDGETS[tier]
+  local budget = budgets and budgets[level]
+  if budget == nil then
+    return nil
+  end
+
+  return { thinkingBudget = budget }
+end
+
 ---@class CodeCompanion.HTTPAdapter.Gemini: CodeCompanion.HTTPAdapter
 return {
   name = "gemini",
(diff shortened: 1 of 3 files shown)


lua/codecompanion/adapters/http/gemini.lua
```diff
@@ -62,6 +62,34 @@ local function decode_args(args)
...
+---
+---Gemini 2.5 Pro cannot fully disable thinking: the API rejects thinkingBudget = 0,
+---so `none` maps to 128 — the minimum valid budget — rather than a true off state.
+local THINKING_LEVEL_BUDGETS = {
+  pro = { none = 128, low = 1024, medium = 8192, high = 24576 },
+  flash = { none = 0, low = 1024, medium = 8192, high = 24576 },
+  flash_lite = { none = 0, low = 512, medium = 8192, high = 24576 },
+}
+
+local function resolve_thinking_config(opts, level)
```
olimorris: Can we add some type annotations here? And it's common practice in the plugin to have `level, opts` as the parameter order

lua/codecompanion/adapters/http/gemini.lua
```diff
@@ -62,6 +62,34 @@ local function decode_args(args)
   return {}
 end
 
+---Gemini 2.5 models use `generationConfig.thinkingConfig = { thinkingBudget = N }`
+---rather than a raw thinking level. Models opt into this budget-based shape via
+---`opts.thinking_budget`, which selects a row from `THINKING_LEVEL_BUDGETS`.
+---Other reasoning models continue to use the `{ thinkingLevel = "<level>" }` shape.
+---
+---Gemini 2.5 Pro cannot fully disable thinking: the API rejects thinkingBudget = 0,
```
olimorris: This is excessive commenting. A single line explaining what they do and a link to Google's own docs will suffice here

olimorris:
2026-04-09 08_00_07 - Google Chrome

I've just seen that these models are being shutdown in 2-3 months time

maxronner:
Happy to fix the remaining blockers if you still want to take this for the 2.5 models.

I can see both sides, though. With Gemini 2.5 this close to EOL, adding model-specific handling here probably only makes sense if we’re okay with a bit of temporary special-casing. If that’s the preference, I think it’d be reasonable to land this with a follow-up note to remove the internal mapping once 2.5 support is gone.

olimorris:
I'll give this a test later on tonight and merge. I think there's still value in having this in the adapter.

PR #3008, opened 2026-04-09
feat(chat): add plan fold for outputting the current state of plans

cenk1cenk2:
## Description

At somepoint this was a feature request where, we would have plan updates and thanks to the new structure I think it works perfectly just as a `thinking` block.

### Notes

- I have messed up with the icon mechanism a bit to be able to reuse it, it does not get everything fully from the config still reuses stuff but leaves room for extension.
- This also caused ambiguity where it now renames the existing `icon_info` to make it specific to `tool_icon` where we now can have additional icons.
- I was a bit indifferent of reusing the `reasoning` treesitter parser, just to comply just created somewhat of a copy with ai.

## AI Usage

- Claude Opus 4.6 - I was going to do the boilerplate but it had so good of an example with the thinking block, it just copied the whole thing which looks quite good tbh, then manually moved things around to match what I would like to do API wise and reusing the functionality.

## Screenshots

image
image

## Checklist


olimorris:
Thanks for this PR, I don't have any plans to incorporate this into `main` right now.

PR #3074, opened 2026-04-28
fix(chat): hide tool context for ACP chats

meijieru:
## Summary
- hide tool and tool-group context entries by default when the chat adapter is ACP
- keep HTTP chat tool context behavior unchanged
- add regression coverage for ACP single tools and tool groups

## Test
- make test_file FILE=tests/interactions/chat/tools/test_tool_registry.lua

### DIFF
--- a/lua/codecompanion/interactions/chat/init.lua
+++ b/lua/codecompanion/interactions/chat/init.lua
@@ -719,6 +719,7 @@ function Chat:change_adapter(adapter, cb)
   end
 
   self:set_system_prompt()
+  self.tool_registry:update_context_visibility()
   self:update_metadata()
   self:apply_settings()
   fire()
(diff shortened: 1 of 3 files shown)


olimorris:
FYI ACP adapters can't use tools so this should not be an issue.

meijieru:
That makes sense. However, I currently have default_tools = { "agent" } configured primarily for the HTTP adapter. To address this, we could either implement this display filter for tool context or scope the default_tools configuration specifically to the HTTP adapter.

olimorris:
This PR is too light on details. You've ignored the default PR template, you've not shared any config which allows me to see what the issue your trying to fix is. And broadly speaking, 131 lines of new code with 3 lines of commentary in a PR is not going to work for me.

PR #3097, opened 2026-05-05
feat(chat): per-tab chats

georgeharker:
## Description

For a long time i've wanted to be able to open a new chat window in a different tab without closing the first.  I often work in multiple tabs at once and whilst you can cycle chats in the chat window it hasn't been possible to have more than one open at a time.

This PR adds display.chat.window.pertab for tab-local chats

When enabled, chat buffers visible in another tab are not hidden when opening or cycling chats in the current tab. Cycling via { / } is scoped to chats that are either visible in the current tab or not currently visible in any tab, so chats opened in other tabs are never stolen.  

The Toggle command jumps to the existing tab when the chat is open elsewhere (with a notify) instead of moving the chat. The close keymap prefers a chat that's already in or available to the current tab when auto-opening a sibling chat after close. Chats survive tab closure and fall back into the hidden pool, where they remain cycle-eligible from any tab.  Ie. if a tab is closed, the same behavior as currently without per-tab chats happens, and the cycling in a separate window will cycle through the now closed chat window's tabs.

pertab is mutually exclusive with sticky; if both are enabled, the sticky autocmd's callback no-ops (a warning is logged at setup) so the chat does not follow tab switches under pertab semantics.

## AI Usage

Drafted in part with claude to figure out the right interception points, design mine,

## Related Issue(s)

N/a

## Screenshots

N/a

## Checklist

### DIFF
--- a/lua/codecompanion/config.lua
+++ b/lua/codecompanion/config.lua
@@ -1095,7 +1095,11 @@ The user is working on a %s machine. Please respond with system specific command
       -- Window options for the chat buffer
       window = {
         buflisted = false, -- List the chat buffer in the buffer list?
-        sticky = false, -- Chat window follows when switching tabs
+        sticky = false, -- Chat window follows when switching tabs (ignored when `pertab` is true)
+        pertab = false, -- Treat each tab as having its own chat window? Chats opened in
+        -- another tab won't be hidden when opening/cycling chats in the current tab,
+        -- and `{`/`}` cycling is scoped to chats that are either visible in the current
+        -- tab or not currently visible in any tab. Mutually exclusive with `sticky`.
 
         layout = "vertical", -- float|vertical|horizontal|tab|buffer
         full_height = true, -- for vertical layout
--- a/lua/codecompanion/config.lua
+++ b/lua/codecompanion/config.lua
@@ -1096,10 +1096,7 @@ The user is working on a %s machine. Please respond with system specific command
       window = {
         buflisted = false, -- List the chat buffer in the buffer list?
         sticky = false, -- Chat window follows when switching tabs (ignored when `pertab` is true)
-        pertab = false, -- Treat each tab as having its own chat window? Chats opened in
-        -- another tab won't be hidden when opening/cycling chats in the current tab,
-        -- and `{`/`}` cycling is scoped to chats that are either visible in the current
-        -- tab or not currently visible in any tab. Mutually exclusive with `sticky`.
+        pertab = false, -- Treat each tab as having its own chat window?
 
         layout = "vertical", -- float|vertical|horizontal|tab|buffer
         full_height = true, -- for vertical layout
--- a/lua/codecompanion/init.lua
+++ b/lua/codecompanion/init.lua
@@ -524,10 +524,20 @@ CodeCompanion.setup = function(opts)
   end
 
   local window_config = config.display.chat.window
-  if window_config.sticky and window_config.layout ~= "buffer" and window_config.layout ~= "tab" then
+  if window_config.sticky and window_config.pertab then
+    log:warn(
+      "[CodeCompanion] `display.chat.window.sticky` and `display.chat.window.pertab` are mutually exclusive. "
+        .. "Disabling `sticky` because `pertab` makes chats tab-local by design."
+    )
+  elseif window_config.sticky and window_config.layout ~= "buffer" and window_config.layout ~= "tab" then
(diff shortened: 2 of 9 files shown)


lua/codecompanion/interactions/chat/keymaps/init.lua
```diff
@@ -299,6 +299,48 @@ M.regenerate = {
   end,
 }
 
+---Whether the given buffer is "available" to the current tab \u2014 i.e. either
```
olimorris: Can we revise these comment blocks down to a single line like the rest of CodeCompanion? They're very typical of Claude Code's verbose writing style in recent months

lua/codecompanion/interactions/chat/keymaps/init.lua
```diff
@@ -299,6 +299,48 @@ M.regenerate = {
...
+---it is not currently displayed in any window, or it is displayed in a window
+---in the current tab. A buffer that is only visible in another tab is NOT
+---available to the current tab. Used in pertab mode to avoid stealing chats
+---from other tabs.
+---@param bufnr number
+---@return boolean
+local function bufnr_available_to_current_tab(bufnr)
+  local wins = vim.fn.win_findbuf(bufnr)
```
olimorris: Can we use `nvim_list_wins()` here? It's going to be so much faster than the vimscript version

lua/codecompanion/interactions/chat/keymaps/init.lua
```diff
@@ -299,6 +299,48 @@ M.regenerate = {
...
+  for _, w in ipairs(wins) do
+    if api.nvim_win_get_tabpage(w) == current_tab then
+      return true
+    end
+  end
+
+  return false
+end
+
+---Build a registry filter for chat cycling.
```
olimorris: Same comment about this large comment block. I think the code is simple enough to know what this function is doing given its name

lua/codecompanion/init.lua
```diff
@@ -524,10 +524,20 @@ CodeCompanion.setup = function(opts)
...
+  if window_config.sticky and window_config.pertab then
+    log:warn(
+      "[CodeCompanion] `display.chat.window.sticky` and `display.chat.window.pertab` are mutually exclusive. "
+        .. "Disabling `sticky` because `pertab` makes chats tab-local by design."
+    )
+  elseif window_config.sticky and window_config.layout ~= "buffer" and window_config.layout ~= "tab" then
     api.nvim_create_autocmd("TabEnter", {
       group = api.nvim_create_augroup("codecompanion.sticky_buffer", { clear = true }),
       callback = function(args)
+        -- No-op if pertab has been enabled by a subsequent setup() call.
```
olimorris: Same comments here. "No-op" is a very Claude Code thing to say right now when most of CodeCompanion is plain english.

lua/codecompanion/init.lua
```diff
@@ -524,10 +524,20 @@ CodeCompanion.setup = function(opts)
...
+        -- pertab and sticky are mutually exclusive; pertab wins.
```
olimorris: Also, we're duplicating the "mutually excusive" comment as we state it in `log:warn`

lua/codecompanion/config.lua
```diff
@@ -1095,7 +1095,11 @@ The user is working on a %s machine. Please respond with system specific command
       -- Window options for the chat buffer
       window = {
         buflisted = false, -- List the chat buffer in the buffer list?
-        sticky = false, -- Chat window follows when switching tabs
+        sticky = false, -- Chat window follows when switching tabs (ignored when `pertab` is true)
+        pertab = false, -- Treat each tab as having its own chat window? Chats opened in
```
olimorris: Let's keep this to just `Treat each tab as having its own chat window?`. The docs provide the additional detail should the user need it

olimorris (commented):
Totally up for this PR and a great suggestion. Some minor nitpicks around how Claude has inserted verbose comments. I do wonder if we can use a faster Lua function in place of vimscript though.

olimorris:
Note to self: Still need to try this out

georgeharker:
Addressed comments

georgeharker:
Let me know if you give it a try and find anything. I've been running with this for a few days and it seems to behave well - allows me to have multiple chats open at once as I multiplex different aspects of work.

olimorris:
Great PR, thank you.

PR #3116, opened 2026-05-16
fix(rules): Resolve relative @ includes against source file directory

toniher:
## Description

Add logic to both claude and cli parsers to resolve relative @ include paths against the directory of the source file rather than the current working directory. This allows includes to work correctly when the source file is located outside the `cwd`. Includes absolute paths (starting with / or ~) and cases where source_dir cannot be determined are left unchanged. Add corresponding test cases for both parsers.

Coming from discussion here: https://github.com/olimorris/codecompanion.nvim/discussions/2688#discussioncomment-16753440

## AI Usage

Claude Code with Sonnet 4.7

Prompt:

````
In the parser of the file `/users/bi/thermoso/remote-work/others/codecompanion.nvim.claude-include/lua/codecompanion/interactions/shared/rules/parsers/claude.lua` (with buffer number: 11), if the file is a user global one the file paths after '@' are not found if they are relative. That is: ~/.claude/CLAUDE.md can have the following content:

@RTK.md


but I want it to get the content of RTK.md that is relative to that file ~/.claude/CLAUDE.md , not from the working directory. How can it be handled in the code?
````

...

```
Can you provide a test for this new change considering recommendations from CONTRIBUTING based on Mini.test
```

## Checklist

- [X] _(optional)_ I've added [test](https://github.com/olimorris/codecompanion.nvim/blob/main/CONTRIBUTING.md#testing) coverage for this fix/feature

### CODE REVIEW

File: lua/codecompanion/interactions/shared/rules/parsers/cli.lua
Comment (olimorris) [inline lua/codecompanion/interactions/shared/rules/parsers/cli.lua:1]: I think the CLI parser should be excluded from this. This is for CLI interactions which should not be applied to rules. Only for CLI interactions

File: tests/interactions/shared/rules/parsers/test_claude_parser.lua [hunk 1]
```diff
@@ -56,4 +56,46 @@ T["Claude parser"] = function()
...
+    }
+  ]])
+
+  -- Create a temp directory to simulate e.g. ~/.claude/
+  local dir = child.lua("return vim.fn.tempname()")
+  child.fn.mkdir(dir, "p")
+
+  -- Create an included file inside that directory
+  local included_name = "RTK.md"
+  local included_path = dir .. "/" .. included_name
```
olimorris: This should be a joinpath otherwise it will break on Windows.

lua/codecompanion/interactions/shared/rules/parsers/claude.lua
```diff
@@ -43,6 +46,13 @@ return function(file)
         local path = line:match("^%s*@(%S+)")
         if path and not seen[path] then
           seen[path] = true
+          -- If path is relative and source file is outside cwd, resolve against source dir
+          if source_dir and not path:match("^[/~]") then
+            local resolved = vim.fs.normalize(vim.fs.joinpath(source_dir, path))
+            if vim.fn.filereadable(resolved) == 1 then
```
olimorris: I have a module _utils.files.exists_ which we should use in place of any vim functions. It's native and wayyyyy faster.

olimorris (changes requested):
> against the directory of the source file rather than the current working directory

Looking at the code, is this actually the case? It reads like it tries the source file dir and then tries the cwd? Otherwise, this is potentially a breaking change for users.

olimorris:
But absolutely solid addition to the plugin and one that I should have in from the start. We should also update the configuration page for rules to clarify how the file resolution works.

lua/codecompanion/interactions/shared/rules/parsers/claude.lua
toniher: Applied!

toniher:
This should be addressed! More clarity and a bit of documentation.

olimorris:
Super thank you. And apologies for being slow, I've been travelling so only got chance to look at this now.

Regarding my post above:

> against the directory of the source file rather than the current working directory

Is this actually the case? It reads like it tries the source file dir and then tries the cwd? Otherwise, this is potentially a breaking change for users.

toniher:
> Super thank you. And apologies for being slow, I've been travelling so only got chance to look at this now.
> 
> Regarding my post above:
> 
> > against the directory of the source file rather than the current working directory
> 
> Is this actually the case? It reads like it tries the source file dir and then tries the cwd? Otherwise, this is potentially a breaking change for users.

Yeah, it simply checks if there is '/' or '~' prepended (absolute path). If a file is in `~/.claude/CLAUDE.md` and `CLAUDE.md` has ``@RTK.md`` as content, it should import `~/.claude/RTK.md`. Users can keep  ``~/.claude/RTK.md`` as content and everything should work the same. That is how users must be keeping their files for global usage right now. It looks to me that thanks to this change, relative paths can be used. For project usage, nothing changes… If you see any case that might require an extra test, it can be added…

toniher:
A potential breaking change could be that a global CLAUDE.md (`~/.claude/CLAUDE.md`) has, let's say:

```
@INSTRUCTIONS.md
```
Should it load from the current directory or find it a `~./claude` ?
If INSTRUCTIONS.md exists at `~/.claude`, that file is loaded; otherwise it will look for it in the current project directory and will load it if it exists. If no file exists in any of the two paths, there will be a warning prompt. 

The outcome t is explicit in the context load, but I can write it down in the documentation.

olimorris:
That's entirely logical and what I think user's would expect from rules files. Defo worth documenting this

toniher:
> That's entirely logical and what I think user's would expect from rules files. Defo worth documenting this

What would you add more apart from this: https://github.com/olimorris/codecompanion.nvim/pull/3116/commits/828adc49b1be5df36a954709f86827b8112a3aa4 ?

olimorris:
Just added a paragraph to the docs.

Thanks so much for this PR. A huge QOL improvement.

PR #3147, opened 2026-06-09
feat(mcp): support lazy mcp server config evaluation

Ajnasz:
## Description

This PR adds support for lazy (deferred) evaluation of MCP server configurations. Instead of providing a static config table, users can now supply a function that returns the config. The function is called once, the first time the server is needed, and the result is memoized back into the config.

This is useful when the config depends on values that should be resolved at runtime (e.g., reading environment variables like API keys) rather than at plugin setup time.

A new `get_server_config()` helper handles the resolution and memoization. All internal call sites (`start_servers`, `enable_server`, `disable_server`, `toggle_server`, `transform_to_acp`) now go through this helper.

The `servers` type annotation has been updated to reflect the new accepted type:
```lua
table<string, CodeCompanion.MCP.ServerConfig | fun(): CodeCompanion.MCP.ServerConfig>
```

Documentation has been updated with a "Lazy / Deferred Config" example.

## AI Usage

This PR was authored with assistance from GitHub Copilot CLI.

## Related Issue(s)

Referring to discussion https://github.com/olimorris/codecompanion.nvim/discussions/3144

## Screenshots

N/A

## Checklist

### DIFF
--- a/doc/configuration/mcp.md
+++ b/doc/configuration/mcp.md
@@ -43,6 +43,24 @@ require("codecompanion").setup({
 })
 ```
 
+```lua [Lazy / Deferred Config]
+require("codecompanion").setup({
+  mcp = {
+    servers = {
+      -- The function is called once, only when the server is first needed.
+      ["tavily-mcp"] = function()
+        return {
+          cmd = { "npx", "-y", "tavily-mcp@latest" },
+          env = {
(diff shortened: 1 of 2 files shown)


olimorris:
Thanks for the PR.

One edit I made was to ensure that we're not overwriting the user's config with the expanded function. We should memoize the resolved config but store it at a module level so that if `M.refresh` is called, we re-evaluate the server config function.

PR #3157, opened 2026-06-13
feat(http): write headers to disk

olimorris:
## Description

Something that has been on my todo list for far too long...prevent users from accidentally copying API keys when they share their logs.

This PR writes headers to a temp file and then shares the path to the file with `curl`.

## AI Usage

Sonnet 4.6

## Checklist

### DIFF
--- a/lua/codecompanion/http.lua
+++ b/lua/codecompanion/http.lua
@@ -9,9 +9,9 @@ local utils = require("codecompanion.utils")
 
 ---@class CodeCompanion.HTTPClient
 ---@field adapter CodeCompanion.HTTPAdapter
----@field static table
----@field opts nil|table
 ---@field methods table
+---@field opts nil|table
+---@field static table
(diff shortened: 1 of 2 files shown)


euri10:
> Something that has been on my todo list for far too long...prevent users from accidentally copying API keys when they share their logs.

the drawback being that now we get sensitive data on file...

2 things I noticed while reading this quickly:
1. the file is removed only in case log level is info or error, why not the rest ?
2. plenary will write that file using it's default 0666, so it opens the file as user/group/other readable/writable before umask is applied, maybe forcing it to 600 would be better this said it alleviates readability issues but not leftovers files

olimorris:
> the drawback being that now we get sensitive data on file...

It was always on file anyway, in the log. The difference is now it's in a temporary location so will be cleaned out by nvim on exit. Or, via the cleanup function at the end of the request

PR #3175, opened 2026-06-20
feat: add Requesty adapter

Thibaultjaigu:
## Summary

Requesty is an OpenAI-compatible model gateway using the same `provider/model` format as OpenRouter; this mirrors the existing OpenRouter adapter.

The adapter fetches the model list from `https://router.requesty.ai/v1/models`, builds the schema, and delegates request handling to the OpenAI adapter. Requesty reports model capabilities as booleans (`supports_tool_calling`, `supports_reasoning`, `supports_vision`) and exposes the context window as `context_window`, so the model parsing maps those onto the same structure the schema fields expect; everything else mirrors the OpenRouter adapter.

## Files changed

- `lua/codecompanion/adapters/http/requesty.lua` — new self-contained adapter
- `lua/codecompanion/config.lua` — register `requesty = "requesty"` in `adapters.http`
- `doc/index.md` — add Requesty to the adapter list and provider line
- `README.md` — add Requesty to the supported provider line

## Verification

- `luac -p` passes on `requesty.lua` and `config.lua`
- `stylua --check` clean on both Lua files
- Live-tested against the Requesty API with `openai/gpt-4o-mini`:
  - `GET /v1/models` → HTTP 200 (554 models parsed; e.g. `openai/gpt-4o-mini` resolves to context window 128000, tools = true, vision = true)
  - `POST /v1/chat/completions` → HTTP 200 with a valid completion
- The repo test suite (`make test`) requires Neovim 0.11.0+; the local environment has 0.10.3, where the suite fails to load on an unrelated `vim.validate` signature change in `utils/log.lua`. The adapter table and its model-parsing were instead verified directly by loading the module against the real `/v1/models` payload.

---

I work at Requesty. This mirrors the existing OpenRouter adapter as closely as possible. Happy to adjust naming/placement or close it if it is not a fit.


olimorris:
Hi @Thibaultjaigu and thanks for the PR.

Unfortunately this isn't something I want to support in the core of CodeCompanion right now. I outlined in #800 my thoughts on supporting adapters in CodeCompanion.

Supporting adapters reduces the time I get to spend on new features. OpenRouter is an exception as it's something that is heavily used by the community and now, by myself.

As far as this PR, I'd recommend sharing the adapter in the discussions as a "show and tell" so others can utilise it. CodeCompanion has been completely designed so I am not a bottleneck or the person holding back new adapters from being used by the community.

Finally, and a nit pick, this PR has not adhered to the PR template or the COMMUNITY.md file.

PR #3179, opened 2026-06-20
feat(adapters): add additional `OpenRouter` features

olimorris:
## Description

This PR builds on the OpenRouter adapter which was added in #3168, adding some additional features:

- Server tools
- Preservation of reasoning content
- Increasing the reasoning effort choices
- Attributing CodeCompanion to requests for better usage tracking
- Support for presets and provide routing (via the schema)
- Support for sticky sessions

The documentation has been updated accordingly.

Shout out to @stasjok in #1013 for their ideas.

## Supporting Documentation

- https://openrouter.ai/docs/app-attribution#attribution-headers
- https://openrouter.ai/docs/guides/features/server-tools/overview
- https://openrouter.ai/docs/guides/best-practices/reasoning-tokens
- https://openrouter.ai/docs/guides/best-practices/reasoning-tokens#reasoning-effort-level
- https://openrouter.ai/docs/guides/best-practices/prompt-caching#using-session_id-for-sticky-sessions

## AI Usage

Claude Code + Opus to interpret the OpenRouter docs and implement my TDD

## Related Issue(s)

[#1013](https://github.com/olimorris/codecompanion.nvim/discussions/1013#discussioncomment-17365008)

## Checklist


stasjok:
When trying to tackle a reasoning preserving, I stumbled on the the strange thing: the index in all reasoning deltas was 0, even for the reasoning blocks of different types. I don't know why this is, maybe they increment deltas only in one data line, but for this reason I I couldn't merge deltas by index. In this PR in `openrouter_reasoning_streaming.txt` file I can see this too: `"index":0` for both `reasoning.summary` and `reasoning.encrypted`. To be honest I don't how this is supposed to be reconstructed correctly. Different providers return different types: OpenAI has summary, and then finally one encrypted block (maybe it's even enough to send only encrypted back). In `T["OpenRouter adapter"]["forms reasoning output"]` test here, encrypted block doesn't have the index. Is this test has correct input data? Also I think it's best to add reasoning output from OpenAI/Anthropic/Google/xAI models, because they have a different combination of reasoning blocks.

stasjok:
Regarding presets. Since presets can also be used in place of the model (with `@preset/NAME`), I think it would be useful to queue presets (https://openrouter.ai/docs/api/api-reference/presets/list-presets) and add them to the model selection. Maybe make it configurable whether the provider returns models and/or presets. I can see the case, that I can configure e.g. a server-side presets `chatgpt-smart` and `chatgpt-fast`, and I can change from the OpenRouter interface what model is the default for these presets without configuring Nvim.

Otherwise, thank you very much! I think I can scrap my custom provider now. It was unfinished, but I needed too much customization/overrides compared to `openai_compatible`. Now most of the staff should work.

stasjok:
> When trying to tackle a reasoning preserving, I stumbled on the the strange thing: the index in all reasoning deltas was 0, even for the reasoning blocks of different types. I don't know why this is, maybe they increment deltas only in one data line, but for this reason I I couldn't merge deltas by index. In this PR in `openrouter_reasoning_streaming.txt` file I can see this too: `"index":0` for both `reasoning.summary` and `reasoning.encrypted`. To be honest I don't how this is supposed to be reconstructed correctly. Different providers return different types: OpenAI has summary, and then finally one encrypted block (maybe it's even enough to send only encrypted back). In `T["OpenRouter adapter"]["forms reasoning output"]` test here, encrypted block doesn't have the index. Is this test has correct input data? Also I think it's best to add reasoning output from OpenAI/Anthropic/Google/xAI models, because they have a different combination of reasoning blocks.

Here is my unfinished provider: https://github.com/stasjok/dotfiles/blob/082c045060748e136021b68316180d66de057932/nvim/ai/codecompanion/openrouter.lua. Maybe something would be useful, e.g. `build_reasoning`. I remember that index was misleading, so I started to increment it myself. Maybe you'll find a better or more correct way, because I wasn't sure what I supposed to get in the end. Maybe we don't need to store/send back index at all. I definitely suggest to record chunks for OpenAI/Anthropic/Google/xAI and build from there.

stasjok:
I think tools should use `openrouter:` prefix, like `{ "type": "openrouter:web_search" }`. Also is there a way to set parameters for server tools from user configuration? Only by overriding `available_tools`'s callback?

olimorris:
I want users to be able to seamlessly switch between adapters and therefore be able to use the same tools. Having to remember that OpenRouter tools are named something different isn't ideal.

And yes, you'd need to overwrite it.

stasjok:
> Having to remember that OpenRouter tools are named something different isn't ideal.

No, user-facing name is the same. I mean that it should be:

```lua
available_tools = {
    ["web_fetch"] = {
      description = "Gives any model the ability to fetch content from a specific URL",
      ---@param self CodeCompanion.HTTPAdapter.OpenRouter
      ---@param meta { tools: table }
      callback = function(self, meta)
        table.insert(meta.tools, {
          type = "openrouter:web_search",
        })
      end,
    },
```

as per documentation: https://openrouter.ai/docs/guides/features/server-tools/web-search

olimorris:
From my testing it doesn't make a difference

PR #3192, opened 2026-06-28
fix(chat): handle model override for ACP adapters

impankratov:
## Description

When trying to start a chat with ACP adapter and specify the model:
```
:CodeCompanionChat adapter=opencode model=opencode/big-pickle
```

codecompanion crashes with error:

```lua
  Lua :command callback: ...k/core/opt/codecompanion.nvim/lua/codecompanion/init.lua:133: attempt to index field 'schema' (a nil value)
  stack traceback:
    ...k/core/opt/codecompanion.nvim/lua/codecompanion/init.lua:133: in function 'chat'
    ...t/codecompanion.nvim/lua/codecompanion/commands/init.lua:164: in function <...t/codecompanion.nvim/lua/codecompanion/commands/init.lua:139>
```

ACP adapters (opencode for example) lack a schema property, so adapter.schema.model.default assignment crashes with attempt to index field schema (a nil value).

Fix: check for schema presence first, fall back to adapter.defaults.model for ACP adapters.

## AI Usage

opencode's big pickle

## Checklist

### DIFF
--- a/doc/codecompanion.txt
+++ b/doc/codecompanion.txt
@@ -1,4 +1,5 @@
-*codecompanion.txt*          For NVIM v0.11          Last change: 2026 June 23
+*codecompanion.txt*
+                                   For NVIM v0.11    Last change: 2026 June 28
 
 ==============================================================================
 Table of Contents                            *codecompanion-table-of-contents*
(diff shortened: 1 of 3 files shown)


olimorris:
The command is incorrect. You need to specify `adapter=`. This is covered in the docs.

Also, ACP adapters don't use a schema as hyperparameters are defined by the agent itself. But you can set model options.

impankratov:
Sorry, got the wrong command in the example. Updated it. However error is real and the fix allows to to set the desired model.

>Also, ACP adapters don't use a schema as hyperparameters are defined by the agent itself. But you can set model options.

Sorry, not following here. So I shouldn't use the adapter + model params in case of ACP adapter? Like use only `CodeCompanionChat adapter=opencode` and choose the model afterwards via `/acp_session_options`?

olimorris:
Yep exactly that. The syntax is for http adapters only.

impankratov:
Oh, I see, thank you for clarification.
Any thoughts on how this should be implemented for ACP? Like a completely separate command maybe?
My use case is to have the ability to launch new chat with specific adapter/model combo

olimorris:
> Any thoughts on how this should be implemented for ACP?

You can technically accomplish this now...I think. You can specify an adapter with specific session config options

impankratov:
thanks, will try that)
closing the pr

PR #3204, opened 2026-07-06
fix(codex): filter ACP auth env by auth method

ShIRannx:
## Description

This PR updates the Codex ACP adapter so `codex-acp` only receives API-key environment variables when the selected `auth_method` actually requires them.

In `chatgpt` mode, the adapter now launches without `OPENAI_API_KEY` or `CODEX_API_KEY`, which avoids upstream `codex-acp` API-key detection overriding ChatGPT auth.

This behavior change is specifically needed for the newer `agentclientprotocol/codex-acp` project:
https://github.com/agentclientprotocol/codex-acp

The older `zed-industries/codex-acp` project does not exhibit this problem and behaves correctly already:
https://github.com/zed-industries/codex-acp

This PR also preserves key-based behavior by passing only the relevant key for `openai-api-key` and `codex-api-key`, and adds regression coverage for adapter setup plus ACP process launch.

## AI Usage

Codex (GPT-5)

## Screenshots

N/A

## Checklist

## Tests

- `make test_file FILE=tests/adapters/acp/test_acp.lua`
- `make test_file FILE=tests/acp/test_acp.lua`
- Full suite note: `make test` was run in this environment, but it reports existing screenshot/UI failures unrelated to this change and missing `tree-sitter` CLI parser compilation.
- Formatting note: `make format` could not be completed in this environment because `stylua` is not installed.

### DIFF
--- a/lua/codecompanion/adapters/acp/codex.lua
+++ b/lua/codecompanion/adapters/acp/codex.lua
@@ -1,5 +1,14 @@
 local helpers = require("codecompanion.adapters.acp.helpers")
 
+local AUTH_ENV_BY_METHOD = {
+  ["openai-api-key"] = {
+    OPENAI_API_KEY = true,
+  },
+  ["codex-api-key"] = {
+    CODEX_API_KEY = true,
+  },
+}
+
 ---@class CodeCompanion.ACPAdapter.Codex: CodeCompanion.ACPAdapter
 return {
   name = "codex",
@@ -24,6 +33,7 @@ return {
   },
   env = {
     OPENAI_API_KEY = "OPENAI_API_KEY",
+    CODEX_API_KEY = "CODEX_API_KEY",
   },
   parameters = {
     protocolVersion = 1,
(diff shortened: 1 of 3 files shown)


olimorris:
Thanks for the PR.

I'm not understanding the problem statement, unfortunately. I'm aware of the new repo change for the Codex ACP project but not understanding what that means in terms of changes for CodeCompanion.

A quick look at the code and you're proposing that CodeCompanion's defaults change to allow fallback to different API keys? And Codex ACP doesn't do that itself?

ShIRannx:
Thanks, that is a fair point. After looking at it again, I think this is probably something that should be addressed in codex-acp rather than in CodeCompanion.

What this PR is working around is the newer agentclientprotocol/codex-acp process inferring API-key auth from inherited env vars even when chatgpt auth is selected. My initial thought was to prevent CodeCompanion from passing those vars in that mode, but I agree that this pushes server-specific auth behavior into the client.

So I think the cleaner fix is for ACP itself to respect the explicitly selected auth method and ignore unrelated API-key env when chatgpt is chosen. Happy to close this and move the fix upstream instead.

PR #3219, opened 2026-07-12
feat(rules): opt-in autoload of default rules for prompt-library prompts

ImmanuelHaffner:
## Description

Prompt-library prompts that name no `rules` currently bypass rule loading entirely: `get_callbacks` only invokes `add_callbacks` when the prompt declares a `rules` field, so the global `rules.opts.chat.autoload` default (e.g.
`"default"`) never applies to them. A plain chat, by contrast, autoloads the default rules.

This PR makes that autoload behaviour available to prompt-library prompts, but **as an opt-in** so it doesn't change existing behaviour on upgrade:

- Adds a new `rules.opts.chat.autoload_prompt_library` option, defaulting to `false`.
- When enabled, `get_callbacks` falls back to the global `autoload` default for a prompt that names no rules of its own — matching a plain chat.
- When disabled (the default), a prompt without a `rules` field loads no rules, preserving the historic behaviour.
- A prompt can still opt out explicitly with `rules = "none"`, regardless of the setting.

Why opt-in: silently loading default rules into every library prompt is a behavioural change that shouldn't go live with a release without the user asking for it. Gating it behind a config flag lets users adopt it deliberately.

Also updates the docs: the option is documented on the Rules and Prompt Library configuration pages, the vimdoc is regenerated, and two pre-existing unclosed `::: tip` divs in `prompt-library.md` are fixed (they were causing `panvimdoc` to emit an "unclosed Div" warning and swallow the new callout out of the
generated vimdoc).

## Supporting Documentation

N/A — this is an internal configuration/behaviour change with no external protocol or LLM documentation to reference.

## AI Usage

The implementation, docs, and this PR description were drafted with AI assistance (CodeCompanion) and reviewed/edited by me. The AI helped locate the relevant `get_callbacks` logic, add the config flag, write the doc prose, regenerate the vimdoc, and diagnose/fix the unclosed-div issue.

## Screenshots

N/A — no UI changes.

## Checklist

### DIFF
--- a/lua/codecompanion/interactions/init.lua
+++ b/lua/codecompanion/interactions/init.lua
@@ -39,15 +39,21 @@ end
 local function get_callbacks(selected)
   local opts = selected.opts or {}
   local callbacks = opts.callbacks or {}
-
   --TODO: Remove opts.rules fallback in v20.0.0
   local rules = selected.rules or opts.rules
-  if rules and rules ~= "none" then
-    local rules_cb = rules_helpers.add_callbacks(callbacks, rules)
-    if rules_cb then
-      callbacks = rules_cb
-    end
+  if rules == "none" then
+    -- Explicit opt-out: this prompt loads no rules at all.
+    return callbacks
   end
+  -- When the prompt names rules, load those. When it names none (`rules == nil`),
+  -- fall back to the global `rules.opts.chat.autoload` default by passing no name
+  -- to `add_callbacks`, so prompt-library prompts get the default rules just like a
+  -- plain chat does.
+  local rules_cb = rules_helpers.add_callbacks(callbacks, rules)
+  if rules_cb then
+    callbacks = rules_cb
+  end
+
   return callbacks
 end
 


olimorris:
Can you share some example prompt library prompts along with this PR so I can verify?

ImmanuelHaffner:
Sure. Here are two examples.

## Default Opt-in

```mdx
---
name: "Rules: load default"
description: "Start a chat that loads the default rules group"
interaction: chat
opts:
  is_slash_cmd: true
  alias: "rules-default"
  auto_submit: false
---

## system

This chat explicitly loads the `default` rules group. The rules files
configured under that group (for example `AGENTS.md` and `CLAUDE.md`) are
attached to the chat buffer so you carry the project's persistent instructions
and context.

## user

I've loaded the default rules for this project. Confirm which rules files you
can see, then wait for my next instruction.
```

## Opt-out

```mdx
---
name: "Rules: opt out"
description: "Start a chat that loads no rules, overriding the autoload default"
interaction: chat
rules: none
opts:
  is_slash_cmd: true
  alias: "rules-none"
  auto_submit: false
---

## system

This chat explicitly opts out of rules with `rules: none`. No rules files are
attached, including any group set via the global `rules.opts.chat.autoload`
default. Use this when you want a clean chat without the project's persistent
instructions.

## user

I've started this chat with rules disabled. Confirm that no rules files are
attached, then wait for my next instruction.
```

ImmanuelHaffner:
The whole point for me is that i want the CC rules system to pick up on project-local and user-wide `AGENT.md` (or `CLAUDE.md`), also when running prompts from the prompt lib. Rarely would a prompt be better off *not* having this additional context.

ImmanuelHaffner:
Oh, i opned the PR from the wrong branch. Let me fix that

PR #3222, opened 2026-07-13
feat(rules): opt-in autoload of default rules for prompt-library prompts

ImmanuelHaffner:
## Description

Prompt-library prompts that name no `rules` currently bypass rule loading entirely: `get_callbacks` only invokes `add_callbacks` when the prompt declares a `rules` field, so the global `rules.opts.chat.autoload` default (e.g.
`"default"`) never applies to them. A plain chat, by contrast, autoloads the default rules.

This PR makes that autoload behaviour available to prompt-library prompts, but **as an opt-in** so it doesn't change existing behaviour on upgrade:

- Adds a new `rules.opts.chat.autoload_prompt_library` option, defaulting to `false`.
- When enabled, `get_callbacks` falls back to the global `autoload` default for a prompt that names no rules of its own — matching a plain chat.
- When disabled (the default), a prompt without a `rules` field loads no rules, preserving the historic behaviour.
- A prompt can still opt out explicitly with `rules = "none"`, regardless of the setting.

Why opt-in: silently loading default rules into every library prompt is a behavioural change that shouldn't go live with a release without the user asking for it. Gating it behind a config flag lets users adopt it deliberately.

Also updates the docs: the option is documented on the Rules and Prompt Library configuration pages, the vimdoc is regenerated, and two pre-existing unclosed `::: tip` divs in `prompt-library.md` are fixed (they were causing `panvimdoc` to emit an "unclosed Div" warning and swallow the new callout out of the
generated vimdoc).

#### Default Opt-in

When `rules.opts.chat.autoload_prompt_library = true`:

```mdx
---
name: "Rules: load default"
description: "Start a chat that loads the default rules group"
interaction: chat
opts:
  is_slash_cmd: true
  alias: "rules-default"
  auto_submit: false
---

## system

This chat explicitly loads the `default` rules group. The rules files
configured under that group (for example `AGENTS.md` and `CLAUDE.md`) are
attached to the chat buffer so you carry the project's persistent instructions
and context.

## user

I've loaded the default rules for this project. Confirm which rules files you
can see, then wait for my next instruction.
```

#### Opt-out

```mdx
---
name: "Rules: opt out"
description: "Start a chat that loads no rules, overriding the autoload default"
interaction: chat
rules: none
opts:
  is_slash_cmd: true
  alias: "rules-none"
  auto_submit: false
---

## system

This chat explicitly opts out of rules with `rules: none`. No rules files are
attached, including any group set via the global `rules.opts.chat.autoload`
default. Use this when you want a clean chat without the project's persistent
instructions.

## user

I've started this chat with rules disabled. Confirm that no rules files are
attached, then wait for my next instruction.
```

## Supporting Documentation

N/A — this is an internal configuration/behaviour change with no external protocol or LLM documentation to reference.

## AI Usage

The implementation, docs, and this PR description were drafted with AI assistance (CodeCompanion) and reviewed/edited by me. The AI helped locate the relevant `get_callbacks` logic, add the config flag, write the doc prose, regenerate the vimdoc, and diagnose/fix the unclosed-div issue.

## Screenshots

N/A — no UI changes.

## Checklist

### DIFF
--- a/doc/configuration/rules.md
+++ b/doc/configuration/rules.md
@@ -250,13 +250,13 @@ require("codecompanion").setup({
 
 :::
 
-#### Autoloading Rules for Prompt Library Prompts
+#### Rules in Prompt Library Prompts
 
-By default, [prompt library](/configuration/prompt-library) prompts that don't name any `rules` of their own load **no** rules. This differs from a plain chat buffer, which autoloads the `autoload` groups.
+By default, prompt library prompts will never autoload rule groups. A prompt only gets rules if it names them itself, via its own rules field.
 
-To make prompt-library prompts also autoload the `autoload` groups, opt in with `autoload_prompt_library`:
+To have prompts fall back to the autoload groups when they don't name their own rules, enable autoload_for_prompt_library:
 
-```lua{6} [Autoload for prompts]
+```lua{6} [Autoload for prompt library prompts]
 require("codecompanion").setup({
   rules = {
     opts = {
@@ -269,7 +269,7 @@ require("codecompanion").setup({
 })
 ```
 
-With this enabled, a prompt that names no rules behaves like a plain chat and loads the `autoload` groups. A prompt can still opt out explicitly with `rules = "none"` (see [Rules](/configuration/prompt-library#rules) in the prompt library docs), and a prompt that names its own rules always loads those instead.
+With this enabled, a prompt that explictly names no rules, will have the autoload groups loaded in the chat buffer. However, a prompt that names its own rules will use those instead.
 
 ## Parsers
 
--- a/doc/configuration/rules.md
+++ b/doc/configuration/rules.md
@@ -250,6 +250,27 @@ require("codecompanion").setup({
 
 :::
 
+#### Autoloading Rules for Prompt Library Prompts
+
+By default, [prompt library](/configuration/prompt-library) prompts that don't name any `rules` of their own load **no** rules. This differs from a plain chat buffer, which autoloads the `autoload` groups.
+
+To make prompt-library prompts also autoload the `autoload` groups, opt in with `autoload_prompt_library`:
+
+```lua{6} [Autoload for prompts]
+require("codecompanion").setup({
+  rules = {
+    opts = {
+      chat = {
+        autoload = "default",
+        autoload_prompt_library = true,
+      },
+    },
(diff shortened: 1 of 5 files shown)


lua/codecompanion/config.lua
```diff
@@ -1081,6 +1081,11 @@ The user is working on a %s machine. Please respond with system specific command
         ---@type string|fun(): string
         autoload = "default",
 
+        ---Also autoload the `autoload` rule groups for prompt-library prompts
+        ---that name no rules of their own? Prompts can still opt out with `rules = "none"`.
+        ---@type boolean
+        autoload_prompt_library = false,
```
olimorris: This implies that rules autoload the prompt library. `rules.opts.chat.autoload_prompt_library`.

Also consider this sentence:

"Also autoload the `autoload` rule groups for prompt-library prompts that name no rules of their own? Prompts can still opt out with `rules = "none"`."

Most of config.lua uses single sentences to describe what the config option is.

olimorris:
Great PR, thanks. I've trimmed the docs - they were still confusing to me when I read them.

My take is that I think we should turn this on by default. It's not a breaking change and I think it's actually expected behaviour for a lot of users. And, it's a simple change in their config or their prompts should they wish to turn it off.

I've left a comment above regarding what the config option is called.

ImmanuelHaffner:
Awesome, thanks for taking it over the finish line.

doc/configuration/rules.md
```diff
@@ -269,7 +269,7 @@ require("codecompanion").setup({
 })
 ```
 
-With this enabled, a prompt that names no rules behaves like a plain chat and loads the `autoload` groups. A prompt can still opt out explicitly with `rules = "none"` (see [Rules](/configuration/prompt-library#rules) in the prompt library docs), and a prompt that names its own rules always loads those instead.
+With this enabled, a prompt that explicitly names no rules, will have the autoload groups loaded in the chat buffer. However, a prompt that names its own rules will use those instead.
```
ImmanuelHaffner: The word "explicit" is a bit weird, because not naming rules is not really "explicit"

olimorris:
Final change before I merged was renaming the config variable to `rules.opts.chat.autoload_groups_in_prompt_library`. Read clearer to me this morning

PR #3225, opened 2026-07-15
feat: configurable reasoning effort for Anthropic + a chat keymap to change it

ImmanuelHaffner:
## Description

This PR adds two things:

1. **Configurable reasoning effort for the Anthropic adapter.** Anthropic's `/v1/messages` API accepts an `output_config.effort` level that controls how many tokens Claude spends across thinking, text, and tool calls. The adapter now exposes an `effort` schema option, gated per-model on the effort levels the model advertises (parsed from `capabilities.effort` in the models list), and sets `output_config.effort` in the request body only when the chosen level is supported by the active model.

2. **A generic `ge` chat keymap to change the reasoning effort.** It mirrors the existing `ga` adapter/model picker: it opens a `vim.ui.select` of the current model's supported effort levels (current one marked), and writes the choice into `chat.settings` so it flows through the normal request path with no adapter-specific glue. The picker is deliberately adapter-agnostic — it discovers the effort setting by probing known schema keys (`effort`, `reasoning.effort`, `reasoning_effort`) and reads the valid levels and default straight from that schema entry.

### Adapter support

Because the picker is generic, any HTTP adapter that already exposes an effort-style schema entry under one of the probed keys works with `ge` for free.

**Confirmed working**

- **Anthropic** (`effort`) — verified against the live API (field placement and valid levels `low`/`medium`/`high`/`xhigh`/`max`, per-model gating), plus unit tests for both the adapter wiring and the picker.

**Should work, not manually tested** (already expose a matching effort schema key, so the generic picker should drive them unchanged):

- **OpenAI** (`reasoning_effort`)
- **OpenAI Responses / Codex** (`reasoning.effort`)
- **OpenRouter** (`reasoning.effort`)
- **GitHub Models** (`reasoning_effort`)
- **DeepSeek** (`reasoning_effort`)

**Needs extra effort**

- **Gemini** — has an effort-style control, but under the key `thinkingLevel`, which the picker doesn't currently probe. Adding it is a one-line change to the picker's key list (not included here).
- **All other adapters** (xAI/Grok, Azure OpenAI, OpenAI-compatible, Novita, Mistral, Copilot, Ollama, Hugging Face) expose no effort schema entry. Adding support to any of these is a per-adapter task: parse the provider's effort capability into `opts.reasoning.supported`, add a gated schema entry, and wire the value into the request body — the same pattern this PR applies to Anthropic.

## Supporting Documentation

- Anthropic effort: https://platform.claude.com/docs/en/build-with-claude/effort
- Anthropic models list (effort capability): https://platform.claude.com/docs/en/api/models-list

## AI Usage

This PR was written with AI assistance (CodeCompanion, using Anthropic Claude). The AI was used to investigate the Anthropic API (confirming the exact request field and valid values against the live endpoint), to draft the adapter wiring and the generic `change_effort` keymap, and to write the accompanying tests. All changes were reviewed and directed by me.

#### Reasoning Effort Picker (after pressing `ge`)

image

#### Effort selection confirmation notification

image

#### Chat debug window shows selected effort

image

#### `which-key` preview after pressing `g`

image

## Checklist

### DIFF
--- a/doc/codecompanion.txt
+++ b/doc/codecompanion.txt
@@ -4205,6 +4205,7 @@ available to the user in normal mode are:
 - `close`: `<C-c>` to close the chat buffer
 - `stop`: `q` to stop the current request
 - `change_adapter`: `ga` to change the adapter for the current chat
+- `change_effort`: `ge` to change the reasoning effort for the current model
 - `clear`: `gx` to clear the chat buffer's contents
 - `copilot_stats`: `gS` to show copilot usage stats
 - `btw`: `gm` type a message to the LLM whilst it's streaming
--- a/doc/usage/chat-buffer/index.md
+++ b/doc/usage/chat-buffer/index.md
@@ -138,6 +138,7 @@ The plugin has a host of keymaps available in the chat buffer. The keymaps avail
 - `stop`: `q` to stop the current request
 
 - `change_adapter`: `ga` to change the adapter for the current chat
+- `change_effort`: `ge` to change the reasoning effort for the current model
 - `clear`: `gx` to clear the chat buffer’s contents
 - `copilot_stats`: `gS` to show copilot usage stats
 - `btw`: `gm` type a message to the LLM whilst it's streaming
--- a/lua/codecompanion/adapters/http/anthropic.lua
+++ b/lua/codecompanion/adapters/http/anthropic.lua
@@ -168,6 +168,14 @@ return {
         end
       end
 
+      -- Effort controls how many tokens Claude spends across thinking, text, and tool
+      -- calls. It lives at output_config.effort in the request body, not on `thinking`.
+      -- Ref: https://platform.claude.com/docs/en/build-with-claude/effort
+      local supported_efforts = models and models.opts and models.opts.reasoning and models.opts.reasoning.supported
+      if self.temp.effort and supported_efforts and vim.tbl_contains(supported_efforts, self.temp.effort) then
+        params.output_config = vim.tbl_deep_extend("force", params.output_config or {}, { effort = self.temp.effort })
+      end
+
       return params
     end,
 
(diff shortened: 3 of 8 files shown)


olimorris:
I'll think about this. It feels pretty heavy to include in CodeCompanion for a single schema item.

An ideal solution would be to detect any schema item that is a table and allow the user to select those without having to open the debug window.

Also, what's your use case? Are you regularly changing efforts in your chats between responses? I question whether many other users are doing this so hence it this is really required.

ImmanuelHaffner:
> An ideal solution would be to detect any schema item that is a table and allow the user to select those without having to open the debug window.

Wait, this does *not* require you to open the chat debug window. That is only here to show the selected effort is applied. The picker `ge` works directly from the CC chat buffer.

> Also, what's your use case? Are you regularly changing efforts in your chats between responses? I question whether many other users are doing this so hence it this is really required.

Not within a session / chat. But i do want to select effort up front, based on the task ahead. I didn't find a convenient solution for this. Enabling `show_settings` and then editing this feels odd. It locks you into the adapter and model and i didn't get a completion support for writing my desired effort. This approach follows the adapter and model pickers directly. Same UX.

olimorris:
> Wait, this does not require you to open the chat debug window. That is only here to show the selected effort is applied. The picker ge works directly from the CC chat buffer.

Yep, I get that. My point is, we're singling out `effort` in this PR. For all other schema items, that happen to be a list, the user has to open the debug window. 

> But i do want to select effort up front, based on the task ahead

That's helpful context and a very good point. I agree that it would be nicer to allow a user to do that.

What about something like:

```
:CodeCompanionChat adapter=anthropic schema.effort=low
```

Which we could also allow in Lua via `require("codecompanion").chat(...)`?

For now, put the Anthropic adapter changes through in a separate PR as it's distinct.

ImmanuelHaffner:
> What about something like:
> 
> :CodeCompanionChat adapter=anthropic schema.effort=low
> Which we could also allow in Lua via require("codecompanion").chat(...)?

Actually i tried that, this was my first approach. But i don't get auto-complete inside the command for values after `=`, so the UX is quite bad. Is this an issue on my side? I would not expect users to know valid effort strings by heart 🙃

ImmanuelHaffner:
> For now, put the Anthropic adapter changes through in a separate PR as it's distinct.

IIUC you want me to split the PR?

ImmanuelHaffner:
Split done — the Anthropic adapter changes are now in #3318.

That's a fresh branch off current `main` carrying only the adapter commit: the `effort` schema entry plus the `output_config.effort` mapping in `form_parameters`, and its tests. 2 files, +81/−1, CI green.

One thing worth flagging from the rebase, since it accounts for the only pre-existing line the new PR touches: #3260 removed the `temperature` entry and left a gap at `order = 6`. So in #3318 `effort` takes `5` (next to `extended_thinking` / `thinking_budget`, where it belongs conceptually) and `max_tokens` moves `5` → `6`, which closes that gap. Every other entry keeps its current number.

I'll reduce this PR to just the keymap. See the next comment for where I'd like to take it.

ImmanuelHaffner:
> An ideal solution would be to detect any schema item that is a table and allow the user to select those without having to open the debug window.

Agreed — so rather than ship the effort-specific keymap, I'll implement the generalised picker.

It's worth more design discussion than a PR thread can hold comfortably, so I've opened #3319 with the details: an inventory of what's actually pickable across the adapters on `main`, how the discrete, boolean and numeric cases would each behave, and the bits of `schema.lua` and `debug.lua` it would reuse.

One correction to the premise that widened the scope for me, since it may change your read too: this keymap was never Anthropic-specific. It probes `effort`, `reasoning.effort` and `reasoning_effort`, and `main` already has choice-bearing reasoning settings in `openai`, `deepseek`, `kimi`, `githubmodels`, `openrouter`, `openai_responses`, `gemini` and `gemini_interactions` — roughly a dozen discrete settings across ten adapters, before counting `effort` from #3318.

The one question I'd like your call on before writing anything, because it shapes the whole design: **because `model` is itself a schema entry with `choices`, this picker subsumes the model half of `ga`. Would you be happy for it to take over `ga`** — one "change setting" picker listing the adapter, the model, and every other selectable schema item?

The adapter isn't a schema entry, so it would stay special-cased as a first-class item, and the ACP path in `list_acp_models` plus the `update_system_prompt` call after an adapter switch would carry over unchanged. Nothing `ga` does today would be lost, it would just sit one selection deeper. If you'd rather not disturb existing muscle memory, a separate keymap is just as easy — I have no strong preference.

Happy to keep the discussion in #3319 so this thread stays about the PR.

PR #3226, opened 2026-07-15
docs: OpenAI-compatible multi-model gateway example (DaoXE)

seven7763:
## Summary

Adds a short example under **Configuring HTTP Adapters** using the built-in `openai_compatible` adapter with a multi-model gateway.

Example: [DaoXE](https://daoxe.com) (`https://daoxe.com` + `/v1/chat/completions`).

- Docs only: `doc/configuration/adapters-http.md`
- Same pattern as the existing `llama.cpp` / remote Ollama examples

### Notes

- Account-scoped model IDs
- Chat Completions path only
- DaoXE not available in mainland China
- Contributor is affiliated with DaoXE

## Test plan

- [ ] Docs site / markdown renders
- [ ] Optional: configure adapter and chat with a real DaoXE key + model ID

### DIFF
--- a/doc/configuration/adapters-http.md
+++ b/doc/configuration/adapters-http.md
@@ -436,6 +436,50 @@ require("codecompanion").setup({
 })
 ```
 
+### OpenAI-compatible multi-model gateway (DaoXE)
+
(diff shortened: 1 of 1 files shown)


olimorris:
Thanks for this. I think this is too niche to go in the docs. Post it in the discussions under "Show and Tell" so users can find it.

PR #3238, opened 2026-07-19
fix(acp): update Codex authentication method IDs

alexghergh:
The new codex-acp [adapter](https://github.com/agentclientprotocol/codex-acp) (versioned 1.x) advertises `api-key` and `chat-gpt` as auth methods, while the built-in adapter still selects the method IDs from the retired Zed adapter. This fails with a misleading Responses API 401 in codecompanion.

Use the current method IDs and link the setup documentation to the new repository.

Also: Reject an explicitly configured method that the agent does not advertise and list the available IDs in the error; adapters without an explicit selection retain the first advertised method fallback (guards against silently picking the first advertised method).

## AI Usage

Codex 5.6 Sol.

## Checklist

### DIFF
--- a/lua/codecompanion/acp/init.lua
+++ b/lua/codecompanion/acp/init.lua
@@ -216,11 +216,22 @@ function Connection:_authenticate()
     if #auth_methods > 0 then
       local wanted = self.adapter_modified.defaults.auth_method
       local method_id
+      local available_method_ids = {}
       for _, m in ipairs(auth_methods) do
+        table.insert(available_method_ids, m.id)
         if m.id == wanted then
           method_id = m.id
           break
         end
+      end
+
+      if wanted and not method_id then
+        log:error(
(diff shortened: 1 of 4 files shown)


lua/codecompanion/acp/init.lua
```diff
@@ -216,12 +216,22 @@ function Connection:_authenticate()
     if #auth_methods > 0 then
       local wanted = self.adapter_modified.defaults.auth_method
       local method_id
+      local available_method_ids = {}
       for _, m in ipairs(auth_methods) do
+        table.insert(available_method_ids, m.id)
         if m.id == wanted then
           method_id = m.id
-          break
```
olimorris, line 222: Why remove `break` from here? In the wild, I know we're only working with a few authentication methods but I'd still prefer we return early if we can.

olimorris:
Super, thanks for the PR!

alexghergh:
kept the `break` in place.

PR #3240, opened 2026-07-19
fix(acp): normalize session/list responses

alexghergh:
## Description

On the [codex-acp](https://github.com/agentclientprotocol/codex-acp) adapter, running `session/list` does not correctly produce results inside of codecompanion, because `nextCursor`, a JSON null value, becomes a truthy vim.NIL value and starts session listing again, hitting the 500 sessions limit.

Also, collapse session titles to one line for the picker while preserving the original session metadata used to load the selection. Codex sends session "titles" as the full initial input (including system prompts, user prompt etc...), resulting in very very long multi-line session "titles". This preserves the same text while swapping e.g. newlines to one blank space, so essentially makes everything one line. Still not ideal, but a proper fix is codex-acp territory, not codecompanion.

## Supporting Documentation

- ACP schema - [nextCursor documentation](https://github.com/agentclientprotocol/agent-client-protocol/blob/b06dc4b4e1c2dd9ab388847111bfee0a016782bd/schema/v1/schema.json#L2375-L2378)

## AI Usage

Codex 5.6 Sol

## Checklist

### DIFF
--- a/lua/codecompanion/acp/init.lua
+++ b/lua/codecompanion/acp/init.lua
@@ -327,7 +327,7 @@ function Connection:session_list(opts)
       end
     end
 
-    cursor = result.nextCursor
+    cursor = type(result.nextCursor) == "string" and result.nextCursor or nil
   until not cursor or #all_sessions >= max_sessions
 
   return all_sessions


alexghergh:
Arguably we can be limiting the "title" string to e.g. 256 characters (claude code's ACP does that inside the ACP code), so session listing remains readable. Happy to change the PR to include that if you give a green light.

olimorris:
Should these not be applied upstream in the ACP adapter? Reading your description, this doesn't sound like a CodeCompanion issue to me. Also, there's no data or comparison between a Claude Code to understand the extend of this issue or why a PR is required. I'd be merging and taking you at your word.

alexghergh:
Fair questions, let me clarify:
- For pagination, the codex-acp is the only implementation I found that explicitly returns `nextCursor: null`. Claude returns every session in one response and omits `nextCursor`, and both Qwen and OpenCode paginate but omit it on the final page. However, the ACP [documentation](https://github.com/agentclientprotocol/agent-client-protocol/blob/main/schema/v1/schema.json#L2375-L2378) seems to describe an absent cursor as the end of pagination, but the official v1 schema explicitly permits `nextCursor` to be either a string or `null`. So this is in theory valid ACP, though I understand why you would not necessarily want to merge it. Neovim decodes `null` as `vim.NIL`, so CodeCompanion sends another `session/list` request with `cursor: null`, which restarts Codex pagination from page one, that's the actual issue. In my local test this produced 472 `session/list` requests and stopped at the configured 500 rows, despite there being only 24 unique session IDs.

- The title issue is indeed more of a Codex ACP issue, I'll admit that one. Claude indeed does sanitize titles and caps them at 256 characters on the ACP side, so arguably Codex could to.

olimorris:
Thanks for the explanation and happy to merge this.

In the future, do include references to the ACP schema / docs and any data payloads if possible. It makes it easier for me to cross-check back to the spec and in the future if ever I get an agent to review PRs, its additional context for them.

PR #3244, opened 2026-07-20
fix(anthropic): don't crash the chat on truncated tool_use arguments

ImmanuelHaffner:
## Description

Fixes a crash where a single truncated tool call permanently bricks a chat session.

When a `tool_use` block is truncated — e.g. the response hits `stop_reason = "max_tokens"` before the arguments JSON closes (easily triggered by a high reasoning `effort` with a small `max_tokens`, since `max_tokens` bounds thinking + text + tool JSON) — the partial JSON fragment is stored verbatim in the chat history. On the next submit, the Anthropic adapter's `build_messages` handler called `vim.json.decode` on that fragment with no guard, raising `E5108: Expected comma or object end but found T_END` and aborting the request. Because the poisoned message stays in history, **every** subsequent submit re-crashed and the session was effectively dead.

This PR guards the decode with `pcall`. A non-parseable `arguments` string now degrades to an empty-input `tool_use` block and logs an error, rather than raising. This keeps the `tool_use`/`tool_result` pairing balanced (so the API contract is preserved) and lets an already-poisoned session recover automatically on the next submit. Well-formed arguments are unaffected.

The change is Anthropic-specific because this is the only adapter that decodes tool arguments at build-messages time. `gemini.lua` already defends its tool argument decoding (`decode_args`, ref #2620); this brings the Anthropic adapter in line.

## Supporting Documentation

- Anthropic tool use: https://docs.anthropic.com/en/docs/build-with-claude/tool-use/overview
- Anthropic streaming / `stop_reason`: https://docs.anthropic.com/en/api/messages-streaming
- Prior art: existing `decode_args` defensive decoding in `gemini.lua` (#2620)

## AI Usage

I used CodeCompanion to help investigate the crash, write the root-cause analysis, and draft the fix, tests, and this description. The final code and commit were reviewed and edited by me.

## Screenshots

N/A — no user-visible UI change beyond an error being logged instead of the session crashing.

## Checklist


doc/codecompanion.txt
olimorris, line 1: Feel free to discard the edits in this file

lua/codecompanion/adapters/http/anthropic.lua
```diff
@@ -308,11 +308,27 @@ return {
           m.content = m.content or {}
           for _, call in ipairs(m.tools.calls) do
             local args = call["function"].arguments
+            ---@type table
```
olimorris, line 311: Will this leave the tool_call orphaned? In that, we'll have the tool_call in the message history but no corresponding result because we couldn't process it

ImmanuelHaffner: Good point, thanks. it does leave an orphaned tool call in the messages. i will investigate.

doc/codecompanion.txt
ImmanuelHaffner: Sure. On that end, this is from running StyLua as per your PR requirements.

ImmanuelHaffner:
Good question, and thanks for it — it was worth chasing. Short answer: this PR doesn't leave the truncated call orphaned, but you were right that something gets orphaned. It sits one layer down, and it's independent of this PR.

**Why the truncated call itself stays paired**

The guard doesn't drop the block — it degrades `input` to an empty dict and still emits the `tool_use`. Separately, at response time, the executor fails to decode those same arguments in `_resolve_and_prepare_tool` and calls `chat:add_tool_output(...)` with `You made an error in calling the <tool> tool`. `format_response` keys that output to `tool_call.id`, so the truncated `tool_use` does get its matching `tool_result` and the pairing holds.

**What actually gets orphaned**

In `Tools:execute`, the JSON-error branch does a bare `return`:

```lua
if not resolved_tool then
  if is_json_error then
    -- JSON error was already handled by _resolve_and_prepare_tool
    return
  end
```

That `return` fires *before* `orchestrator:setup_next_tool()`, so it abandons the orchestrator queue along with the rest of the loop. Any tool call that resolved earlier in the same turn is already sitting in that queue and never runs, so it never produces a result. And because a truncated `tool_use` is by construction the last block in the response, a turn carrying several tool calls loses the results of all the earlier, valid ones. Those are the orphans, and they're what makes the next request 400.

That's a pre-existing bug in the tool system rather than anything to do with the adapter, so I've put it in its own PR: #3317. It stops the loop bailing out, and adds a backstop in `Chat:submit` that completes any tool call still missing a result before the payload is built, reusing `_complete_orphaned_tool_calls` from #3073. The two PRs touch disjoint files and neither depends on the other, so they can merge in either order.

**On this PR**

Rebased onto current `main`, so the conflicts are gone. I also dropped the `doc/codecompanion.txt` changes you flagged — they only existed in a separate formatting commit, so skipping that commit removed them entirely. The diff is now just `anthropic.lua` and its tests.

Apologies for the long silence on this one.

PR #3249, opened 2026-07-20
fix(interactions): starting a chat from a focused chat window crashes with an empty message role

ImmanuelHaffner:
## Description

Starting a new chat from the action palette **while a chat window is focused** produces a broken chat: the request fails and the new chat is unusable. With the Anthropic adapter the error is:

```
Error 400: messages.0.role: Input should be 'user', 'assistant', 'system'
```

**Reproduction:**
1. Focus an existing chat window.
2. Press your action palette mapping (e.g. `<leader>a`) → this opens the *chat* command palette because the buffer is a `codecompanion` buffer.
3. Select **"CodeCompanion actions"**, then select **"Chat"**.

The resulting chat's message stack contains two messages with `role = ""`, which the API rejects.

**Root cause:**
Selecting "CodeCompanion actions" calls `require("codecompanion").actions()` from *inside* the picker's selection callback. With a picker whose prompt runs in insert mode (e.g. Telescope), the buffer context captured by `codecompanion.utils.context.get` records `mode = "i"`.

In `Interactions:chat()` the prompt key was derived from that raw mode:

```lua
local mode = self.buffer_context.mode:lower()  -- "i"
```

The built-in "Chat" prompt table is keyed only by `n` and `v`:

```lua
prompts = {
  n = function() return codecompanion.chat() end,
  v = { ... },
}
```

So `prompts["i"]` is `nil`, and the code falls through to `evaluate_prompts(prompts, ...)`, which iterates the whole `{ n = <fn>, v = <table> }` table and, for each entry lacking a `role`, emits `role = prompt.role or ""` — producing two empty-role messages.

**Fix:**
The prompt dispatch only distinguishes visual from non-visual, so derive the key from the authoritative `is_visual` flag rather than the raw captured mode. Any non-visual mode (including a leaked `"i"`) now collapses to normal and takes the correct branch:

```lua
local mode = self.buffer_context.is_visual and "v" or "n"
```

This prevents the whole class of "leaked mode" bugs regardless of which picker provider is used.

## Supporting Documentation

N/A — no external protocol/LLM documentation involved.

## AI Usage

I used CodeCompanion (Anthropic `claude-opus` via a Databricks adapter) to help diagnose the root cause and draft the fix, the regression test, and this description. I reviewed the analysis, confirmed the reproduction, and verified the fix and test myself.

## Screenshots

N/A

## Checklist

### DIFF
--- a/lua/codecompanion/interactions/init.lua
+++ b/lua/codecompanion/interactions/init.lua
@@ -133,7 +133,9 @@ function Interactions:chat()
     return self:workflow()
   end
 
-  local mode = self.buffer_context.mode:lower()
+  -- Prompts are only keyed by "n" and "v"; collapse any other captured mode
+  -- (e.g. picker insert mode leaking through) to normal so the correct branch runs
+  local mode = self.buffer_context.is_visual and "v" or "n"
   local prompts = self.selected.prompts
 
   if type(prompts[mode]) == "function" then
--- a/tests/prompt_library/test_prompt_library.lua
+++ b/tests/prompt_library/test_prompt_library.lua
@@ -270,4 +270,46 @@ T["Prompt Library"]["can ignore system prompt"] = function()
   h.eq(false, has_system_tag)
 end
 
+-- Regression: launching the action palette from a focused chat window captures
+-- the picker's insert mode. The "Chat" prompt table is keyed only by "n" and "v",
+-- so a leaked "i" mode must still resolve to the normal-mode branch rather than
+-- flattening the prompt table into messages with empty roles (which the API rejects).
+T["Prompt Library"]["collapses a leaked mode to normal for n/v keyed prompts"] = function()
+  local has_empty_role = child.lua([[
+    local Interactions = require("codecompanion.interactions")
+    local config = require("codecompanion.config")
+    local selected = {
+      name = "Chat",
+      interaction = "chat",
+      opts = { stop_context_insertion = true },
+      prompts = {
+        n = function()
+          return require("codecompanion").chat()
+        end,
+        v = {
+          {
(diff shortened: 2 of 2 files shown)


olimorris:
Thanks for this. Unfortunately, this should be raised as an issue so it has a complete `minimal.lua` file that I can test. Also, a PR should be used to discuss code and not the issue itself.

Unfortunately I can't recreate this.

https://github.com/user-attachments/assets/d74a7255-367a-445d-b403-edd8e596a107

ImmanuelHaffner:
Thanks for the review, and apologies for the slow follow-up. I've now raised this as an issue with a complete, tested `minimal.lua`: #3322

Two things in my original description here were wrong, and they're almost certainly why it wouldn't reproduce for you:

- **Telescope has nothing to do with it.** It reproduces with the default `vim.ui.select` provider. My commit message blamed a nested picker; that was a misdiagnosis.
- **The trigger is invoking the Action Palette while Neovim is in insert mode.** `<Cmd>` mappings deliberately preserve the current mode, so an insert-mode `<Cmd>CodeCompanionActions<CR>` mapping leaves `mode()` reporting `"i"` when the buffer context is captured — and that raw string is then used as the prompt lookup key. Typing `:CodeCompanionActions` at the command line leaves you in normal mode and everything behaves correctly, so if that's how you tested it, it would never have shown up.

The `minimal.lua` in the issue includes that insert-mode mapping as the trigger, plus a small `:ShowChatRoles` command so the empty-role messages are visible without sending a request at all.

I'm happy to reopen this PR, or open a fresh one against current `main`, whenever you've had a chance to look at the issue — whichever you prefer.

olimorris:
Fixed as part of #3332

PR #3295, opened 2026-08-03
feat(chat): files/buffers are parsed with context formatters

olimorris:
## Description

One of the things I've disliked with `CodeCompanion` is how from a user perspective the sharing of buffers and files feels completely separate. As a user, I don't care if I'm sharing a buffer or a file, I just want the LLM to see it and be updated automatically when they change.

Buffers have had the ability to share a `diff` with an LLM when they're edited for a while. The advantage of this has been that we're not resharing the whole buffer every time an edit's made so it's very token efficient. With files that was not possible and this is where the division between files and buffers was created in the plugin.

This PR closes that by watching files on the file system for any changes and sharing them as a `diff` with an LLM.

This PR also introduces context formatters. That is, a formatter for a specific file type. Taking the new built-in example of a Jupyter Notebook. Most users would not want to share images or dense HTML that can be output from a notebook. Instead, they probably want to share the value of the cells themselves and any text output. So it makes sense to have a context formatter that can pass the contents of a Jupyter Notebook before it ever reaches an LLM. Ensuring that any successive edits you make in that notebook would be shared as a `diff` rather than resharing the whole file.

## AI Usage

Opus

## Checklist


olimorris:
@greptileai

PR #3312, opened 2026-08-14
fix(utils): guard buffers.get_info against invalid buffer ids

aditya-gts:
## Description

Chat sessions cache the source buffer's number in `buffer_context.bufnr` when they're created. If that buffer is later closed/wiped, submitting a message that triggers any editor context expansion (`#buffer`, `#buffers`, etc.) crashes with `Invalid buffer id: N`.

`EditorContext.replace` in `interactions/shared/editor_context/buffer.lua` unconditionally calls `buf_utils.get_info(bufnr)` on the cached `bufnr` to build the fallback replacement text for the bare `#{buffer}` pattern, regardless of whether that pattern is present in the message or whether the buffer still exists. `get_info` called `nvim_buf_get_name` with no validity check, so an invalid handle throws instead of degrading gracefully.

`buffers.get_info()` now checks `nvim_buf_is_valid(bufnr)` up front and returns a `"[invalid buffer]"` placeholder table instead of erroring. This is the single call site all `get_info` callers go through, so it fixes the crash regardless of which editor-context tag triggered it.

## Supporting Documentation

N/A

## AI Usage

Used Claude Code to reproduce the crash from a real Neovim config, trace the stack to the root cause in `buffers.lua`, and write the one-line validity guard. Reviewed and understand the change; removed the accompanying test after maintainer feedback that it wasn't needed.

## Related Issue(s)

None filed - found and fixed directly from a user config crash (see reproduction screenshot in comments below).

## Screenshots

See reproduction screenshot in the comments below.

## Checklist

### DIFF
--- a/lua/codecompanion/utils/buffers.lua
+++ b/lua/codecompanion/utils/buffers.lua
@@ -59,6 +59,18 @@ end
 ---@param bufnr number
 ---@return table
 function M.get_info(bufnr)
+  if not api.nvim_buf_is_valid(bufnr) then
+    return {
+      bufnr = bufnr,
+      filetype = "",
+      name = "[invalid buffer]",
+      number = bufnr,
+      path = "[invalid buffer]",
+      relative_path = "[invalid buffer]",
+      short_path = "[invalid buffer]",
+    }
+  end
+
   local bufname = api.nvim_buf_get_name(bufnr)
 
   return {
--- a/tests/utils/test_buffers.lua
+++ b/tests/utils/test_buffers.lua
@@ -52,26 +52,4 @@ T["Utils->Buffers"]["add_line_numbers works"] = function()
   h.expect_match(result, "3 |test")
 end
 
-T["Utils->Buffers"]["get_info works for a valid buffer"] = function()
-  local info = child.lua([[
-    return _G.buf_utils.get_info(_G.test_buffer)
-  ]])
-
-  h.eq(info.bufnr, child.lua_get("_G.test_buffer"))
-  h.not_eq(info.path, "[invalid buffer]")
-end
-
-T["Utils->Buffers"]["get_info does not error for a deleted/invalid buffer"] = function()
-  local info = child.lua([[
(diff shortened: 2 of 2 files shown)


olimorris:
Thanks for this. Can you share how this arose in the first place? I've not come across this in the c. 2 years that I've had this utility function in CodeCompanion.

Firstly, please stick to the PR template. Secondly, we don't need a test for this.

aditya-gts:
I was trying to pass an open file as context into the CodeCompanion sidebar using #buffer shortcut. It would give me the error, I pasted above, and then I asked claude code to patch it so that I could use the plugin. Thanks for the quick response. I have attached a screenshot of exactly what I was doing while the error occured. 
Screenshot_14-Aug_18-39-28_7671

olimorris:
If you used `#{buffer:init.lua}`, then it shouldn't occur unless you've closed the buffer after the fact. But either way, I agree CodeCompanion should handle this better.

My preference is to return nil from the function when this happens and handle it across the plugin. I'll leave this PR open for a couple of weeks (have a lot of non-CodeCompanion stuff on) and work on this with you.

aditya-gts:
Thanks, will commit again with your preferred method till then. This time by learning how everything works instead of relying on claude.

lua/codecompanion/utils/buffers.lua
```diff
@@ -59,6 +59,18 @@ end
...
+    return {
+      bufnr = bufnr,
+      filetype = "",
+      name = "[invalid buffer]",
+      number = bufnr,
+      path = "[invalid buffer]",
+      relative_path = "[invalid buffer]",
+      short_path = "[invalid buffer]",
+    }
+  end
```
ImmanuelHaffner, line 72: How does a caller distinguish failed vs. successful call? I would have expected returning `nil` or an error and teaching callsites to handle it.

olimorris: It doesn't. Hence my comment above

PR #3314, opened 2026-08-15
feat(adapters): add TrustedRouter HTTP adapter

jperla:
Adds an HTTP adapter for TrustedRouter, an OpenAI-compatible router — one key reaches models from OpenAI, Anthropic, Google, DeepSeek and others, with per-request routing and failover.

Structured after `novita.lua` rather than `openrouter.lua`: TrustedRouter speaks plain Chat Completions, so the adapter delegates to the `openai` handlers and only supplies its own `get_models` against `https://api.trustedrouter.com/v1/models`.

I did not use the shared `adapters/utils/models/fetch` helper, because it requires a matching `transform.from_<name>` transformer and the OpenRouter one reads `supported_parameters`, which TrustedRouter's catalog does not return. The local fetch derives what the response actually carries — vision support from `architecture.input_modalities` — rather than gating on fields I could not verify.

Two behaviours documented in `adapters-http.md`, since neither is guessable:

- Model ids are namespaced. `anthropic/claude-opus-4-7` works; a bare `claude-opus-4-7` is rejected.
- Ids under `trustedrouter/` are routing policies, not single models — each picks an upstream per request and fails over. `zdr` and `e2e` constrain routing to no-retention and confidential-compute providers, which is the useful distinction when the chat buffer carries private source.

Verified both changed Lua files parse (`loadfile`). No adapter test included, matching `novita.lua`, which also ships without one — happy to add one if you'd prefer. `stylua` was not available in my environment, so the file follows `novita.lua`'s formatting by hand; shout if it needs a pass.

### DIFF
--- a/doc/configuration/adapters-http.md
+++ b/doc/configuration/adapters-http.md
@@ -626,6 +626,26 @@ require("codecompanion").setup({
 })
 ```
 
+### TrustedRouter
+
+Set `TRUSTEDROUTER_API_KEY` and select the adapter. Model ids are namespaced —
+`anthropic/claude-opus-4-7` works, a bare `claude-opus-4-7` is rejected.
+
+Ids under `trustedrouter/` are routing policies rather than single models: each picks
+an upstream per request and fails over if one is down.
+
+| Model | Picks for |
+| ----- | --------- |
+| `trustedrouter/auto` | Capability, with failover (default) |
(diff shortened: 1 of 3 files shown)


olimorris:
Thanks. As per #800, I limit the adapters in the project. Please share as a community adapter

PR #3317, opened 2026-08-23
fix(chat): don't lose tool results when one tool call is malformed

ImmanuelHaffner:
## Description

`Tools:execute` abandoned every queued tool call as soon as one call's arguments failed to parse.

When `_resolve_and_prepare_tool` hits unparseable arguments it reports the mistake to the LLM and returns a "handled" flag. The caller then did a bare `return` — and that `return` fires *before* `orchestrator:setup_next_tool()`, so it abandoned the orchestrator queue along with the rest of the loop. Any tool call that resolved earlier in the same turn was already sitting in that queue, and now never ran, so it never produced a result.

This matters most when a response is truncated mid-`tool_use` (for example `stop_reason=max_tokens`), because the malformed call is then by construction the *last* block in the response: a turn carrying several tool calls loses the results of all the earlier, valid ones. Anthropic rejects any request whose `tool_use` blocks have no matching `tool_result`, so the chat could no longer submit at all — the session was stuck.

Two changes:

- `Tools:execute` still reports the malformed call, but carries on through the loop so the remaining tools execute and produce their results. The premature `utils.fire("ToolsFinished")` moved out of the prepare step, which lets `ToolsFinished` fire from its normal places instead — the empty-queue branch, or the orchestrator once the queue drains.
- As a backstop, `Chat:submit` completes any tool call still missing a result before the payload is built. This also covers adapter errors and any other path that ends a turn with unmatched calls. It reuses `_complete_orphaned_tool_calls` from #3073, which now takes the stand-in text as an option so cancellation keeps its existing `Cancelled by user` wording.

On placement, in case it saves a review round: `Chat:done` is the wrong home because it runs *before* tools execute, so completing orphans there would cancel every call the LLM had just asked for. `Chat:tools_done` is too late, because the `ToolsFinished` autocmd can auto-submit through `Chat:submit` before `Tools:reset` ever reaches it. `Chat:submit` is the one point that is both early enough to matter and guaranteed to run after tools have finished, since it returns early on `current_request`. The call sits before the user message is appended, so the stand-in results stay adjacent to the `tool_use` they answer.

Testing notes: both new tests were confirmed failing before the fix and passing after — the first failed with `_G.weather_output` still unset, which is the abandoned queue showing itself directly, and the second failed with `has_orphaned_tool_calls()` returning true. The full suite locally is 1226 cases with 3 failures, which are the same 3 that `main` already fails on macOS (all in `tests/interactions/code_review/`, asserting on temp paths and branch-scoped storage). `stylua --check tests/ lua/` is clean.

## Supporting Documentation

Anthropic's [Handle tool calls](https://docs.anthropic.com/en/docs/agents-and-tools/tool-use/handle-tool-calls#handling-results-from-client-tools) page, which states that "Tool result blocks must immediately follow their corresponding tool use blocks in the message history" and names the resulting failure: "tool_use ids were found without tool_result blocks immediately after". The same section is why the stand-in results are inserted before the user message rather than after it — within the user turn, `tool_result` blocks must come first and any text must follow them.

## AI Usage

This PR was written with CodeCompanion, using Claude (Anthropic). The model traced the bug from the review question on #3244, proposed the two changes and argued the placement, and wrote both tests. I directed the investigation, chose the approach, and reviewed every diff before it landed.

## Related Issue(s)

Companion to #3244, which stops the Anthropic adapter crashing on the same truncated `tool_use`. The two are independent and touch disjoint files, so they can merge in either order.

This PR is the answer to the orphaned-`tool_call` question @olimorris raised in review on #3244. Short version: the adapter guard there does not orphan the truncated call — the executor already reports a result for it — but the question was well aimed, because the sibling calls in the same turn were being dropped, one layer down and independently of that PR.

## Screenshots

Not applicable.

## Checklist

### DIFF
--- a/lua/codecompanion/interactions/chat/init.lua
+++ b/lua/codecompanion/interactions/chat/init.lua
@@ -91,6 +91,7 @@ local CONSTANTS = {
   STATUS_SUCCESS = "success",
 
   BLANK_DESC = "[No messages]",
+  INCOMPLETE_TOOL_CALL = "This tool call did not complete and produced no result",
 
   SYSTEM_PROMPT = [[You are an AI programming assistant named "CodeCompanion", working within the Neovim text editor.
 
@@ -1311,10 +1312,6 @@ function Chat:submit(opts)
 
   opts = opts or {}
 
-  -- A tool call left without a result makes providers such as Anthropic reject the
-  -- entire request, so stand in a result before the payload is built
-  self:_complete_orphaned_tool_calls({ reason = "This tool call did not complete and produced no result" })
-
   if opts.callback then
     opts.callback()
   end
@@ -1326,6 +1323,7 @@ function Chat:submit(opts)
 
   -- Differentiate between the user submitting and CodeCompanion automatically doing it
   if opts.auto_submit then
+    self:_complete_orphaned_tool_calls({ reason = CONSTANTS.INCOMPLETE_TOOL_CALL })
     self:_inject_btw()
     self.buffer_diffs:check_for_changes(self)
   else
@@ -1340,6 +1338,8 @@ function Chat:submit(opts)
       log:info("Chat submission prevented by on_before_submit callback")
       return self:restore()
     end
+
+    self:_complete_orphaned_tool_calls({ reason = CONSTANTS.INCOMPLETE_TOOL_CALL })
 
     self.buffer_diffs:check_for_changes(self)
 
--- a/lua/codecompanion/interactions/chat/init.lua
+++ b/lua/codecompanion/interactions/chat/init.lua
@@ -1091,16 +1091,18 @@ function Chat:has_orphaned_tool_calls()
   return next(self:_orphaned_tool_calls()) ~= nil
 end
 
----Prevent any orphaned tool calls by "completing" them with a cancelled message
+---Prevent any orphaned tool calls by "completing" them with a stand-in result
+---@param opts? { reason?: string } The stand-in result sent to the LLM
 ---@return nil
-function Chat:_complete_orphaned_tool_calls()
+function Chat:_complete_orphaned_tool_calls(opts)
   local pending = self:_orphaned_tool_calls()
   if next(pending) == nil then
     return
   end
 
+  local reason = opts and opts.reason or "Cancelled by user"
   for id, call in pairs(pending) do
-    local output = adapters.call_handler(self.adapter, "format_response", call, "Cancelled by user")
+    local output = adapters.call_handler(self.adapter, "format_response", call, reason)
     if output then
       output.opts = vim.tbl_extend("force", output.opts or {}, { visible = false })
       output._meta = {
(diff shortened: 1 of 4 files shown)


olimorris:
Thanks for the above. I have a few questions about this, notably about the frequency with which this occurs. For reference, the agentic loop and tool execution logic has been in place for a long time and I've _never_ come across this across Anthropic, OpenAI, Kimi, OpenRouter adapters.

So I'm reluctant to merge a fix for something which I believe CodeCompanion is not responsible for. If an LLM is returning "unparseable arguments" then that really is the issue of the model. Now whilst it's not practical to assume a frontier model provider is going to "patch" that, the tools within CodeCompanion should handle this and return an error, therefore preserving the tool call->results chain. I don't think the fix should not be at as high a level as you're proposing in `chat/tools/init.lua`.

Ideally, a fix would be accompanied by an issue or at least the steps that enable me to recreate it.

ImmanuelHaffner:
Fair challenge. I've raised #3326 with a `minimal.lua` that reproduces it deterministically — no API key and no network needed, since the adapter there returns a canned response.

**On frequency.** It needs two things at once: a response with two or more tool calls, *and* one of them with unparseable arguments. That's why it stays invisible with frontier models. The realistic trigger is a response truncated at the output token limit part way through a tool call (the case in #3244); because the truncated block is by construction the last one, every valid call earlier in the same turn is lost.

**On responsibility.** Agreed that a model emitting unparseable arguments is the model's problem, and CodeCompanion already does the right thing there: it reports `You made an error in calling the X tool` and that error is paired with its own tool call. The bug isn't the malformed call — it's the calls next to it:

```lua
for _, tool in ipairs(tools) do
  local resolved_tool, error_msg, is_json_error = self:_resolve_and_prepare_tool(tool, id)
  if not resolved_tool then
    if is_json_error then
      -- JSON error was already handled by _resolve_and_prepare_tool
      return
    end
    ...
  else
    self.tool = resolved_tool
    orchestrator.queue:push(resolved_tool)
  end
end
...
orchestrator:setup_next_tool()
```

That `return` fires before `orchestrator:setup_next_tool()`, so tools already pushed onto the queue never execute, never produce output, and never reach the LLM. The log from the reproduction shows `[Tools] Completed` immediately after the decode error, with the valid `file_search` call untouched.

**On the level of the fix.** The arguments are decoded in `_resolve_and_prepare_tool`, in `chat/tools/init.lua` — the tool layer, which is where you're saying this belongs; there's no lower level where a call's arguments get parsed. The part that actually recovers the dropped siblings is deleting that one `return` (and not reporting the error twice).

The second half of the PR — completing any still-unanswered tool call in `Chat:submit` — is a separable backstop for turns that end with an unmatched call for other reasons (adapter errors, for instance). If you'd rather consider that on its own merits, say the word and I'll re-cut this PR down to the loop fix alone.

olimorris:
Reviewed this and looks good. I added a second `_complete_orphaned_tool_calls` call in chat/init.lua and altered the order that it's called. Notably, putting inside `auto_submit`.

ImmanuelHaffner:
Nice follow-up, handling `on_before_submit` here is a real improvement.
It lead me to ask https://github.com/olimorris/codecompanion.nvim/discussions/3330

PR #3354, opened 2026-09-02
fix: size code fences to their content instead of hard-coding four backticks

### LINKED ISSUE #3353: [Bug]: Tool output is wrapped in a fixed four-backtick fence, so payloads containing fences corrupt the chat buffer
### Pre-submission checklist

- [x] I have read the [documentation](https://codecompanion.olimorris.dev)
- [x] I have updated the plugin and all of its dependencies to the latest versions
- [x] I have searched for existing [issues](https://github.com/olimorris/codecompanion.nvim/issues) and [discussions](https://github.com/olimorris/codecompanion.nvim/discussions)
- [x] If [ACP](https://agentclientprotocol.com) related, I have confirmed that the issue is a CodeCompanion one and not an issue in the adapter itself
- [x] My issue is not a minor or cosmetic quirk (e.g. formatting, spacing, or other non-functional details)
- [x] I have reproduced this from a clean `minimal.lua` and included everything needed for someone else to hit the same bug

(And yes — I noticed the comment in `minimal.lua`. I did run the real file. 🙂)

### Environment

- Neovim: v0.12.4
- OS: macOS 15 (Apple Silicon)
- CodeCompanion: `main` at `6eec6b42` — the exact commit `lazy` fetched into `.repro` during the run below
- Adapter: `copilot`, i.e. the stock `minimal.lua` unedited
- No API call is needed to reproduce this

### What happened, and what did you expect?

Tool output is wrapped in a **hard-coded four-backtick fence**. When the payload itself contains a line whose first non-space characters are four or more backticks, that line closes the wrapper early, and the wrapper's own closing fence then *opens* a block — so fence parity is inverted for the rest of the buffer.

This is easy to hit in practice: CodeCompanion instructs the model to write four-backtick code blocks, so transcripts, notes, `AGENTS.md` files and READMEs are full of them. Any tool that returns the text of such a file trips it. In the reproduction below, a tool returns a README containing two ordinary four-backtick blocks, and the chat buffer ends up structured like this — 22 lines, and Tree-sitter sees **three** code blocks where there should be one:

| Block | Lines | What it actually is |
|-------|-------|---------------------|
| 1 | 6–13 | opens at the wrapper's fence, closed early by the README's *first* inner fence |
| — | 14–16 | the README's prose, now rendered as chat prose rather than tool output |
| 2 | 17–19 | the README's second block, now a top-level block of the chat |
| 3 | 22–22 | the wrapper's own closing fence, now an **unterminated** block that runs to the end of the buffer |

**Expected**: one code block spanning the whole tool result, with the payload preserved verbatim inside it.

Why this is functional rather than cosmetic:

- **It corrupts the structure the chat buffer's own Tree-sitter queries run against.** #913 shows where that leads: an unterminated fence swallows the following `## <user>` heading and the user's next prompt is silently dropped. That issue was closed as not planned on the grounds that four backticks would be enough — but a fixed count cannot be enough, because real files legitimately contain four-backtick fences. @TommyAmberson reports exactly that on #913 (a tool read a project README and the chat broke every time), and this is the same root cause seen from the producer side.
- **Some sites send the broken fence to the model, not just the buffer.** `read_file` and `cmd_tool` fence their `for_llm` payload, so when a file's own fences close the wrapper early the payload boundary the model receives is ambiguous.

Verified sites at `6eec6b42`:

| Site | Lines | Fence | Corrupts |
|------|-------|-------|----------|
| `lua/codecompanion/mcp/tool_bridge.lua` | 44–46, 61–66 | four | the chat buffer (`for_user`) |
| `lua/codecompanion/interactions/chat/tools/builtin/read_file.lua` | 109–111 | four | the payload sent to the model (`for_llm`) |
| `lua/codecompanion/interactions/chat/tools/builtin/cmd_tool.lua` | 100–102 | four | the payload sent to the model (`for_llm`) |
| `lua/codecompanion/interactions/chat/tools/builtin/cmd_tool.lua` | 61–63 | three | both |

The wider inventory is reproducible rather than guessed — `git grep -cF '````' -- lua/` reports 34 files. Many are prompt text documenting the convention to the model and should not change; the rest follow the same pattern across the editor-context providers, `/help`, `/symbols`, `code_review/store.lua`, `context_management/compaction.lua`, the Jupyter formatter and the newer tools. I have not triaged every one, so I have deliberately not listed them as confirmed.

A detail that widens the blast radius: the Tree-sitter Markdown parser is laxer than CommonMark about what closes a fence. Measured against `main`, a four-backtick payload line closes a four-backtick wrapper when indented by one, three **or four** spaces, and a five-backtick line closes it too. Three backticks are safe. So `git diff` *context* lines, which carry a leading space, are dangerous, while `+`/`-` lines are not.

**Suggested fix**: size the fence to the content, the way CommonMark and GitHub intend, never going below the site's current length so there is no cosmetic churn:

````lua
---A fence long enough that `content` cannot terminate it
---@param content string
---@param min? integer Minimum length; defaults to 4
---@return string
function M.fence(content, min)
  local longest = 0
  for run in tostring(content):gmatch("`+") do
    longest = math.max(longest, #run)
  end
  return ("`"):rep(math.max(min or 4, longest + 1))
end
````

Two notes if you want this as a PR:

- The payload must never be rewritten. There is no escape mechanism inside a fenced block, so "fixing up" the payload's own fences would mean altering bytes the user copies out, and it is undefined for an already-unbalanced payload (a truncated read, a `head -20`, a diff hunk starting mid-block). A longer outer fence is robust to all of those.
- One site round-trips its own fences: `interactions/code_review/store.lua` writes them and its reader matches `^````` literally, so it needs to capture the fence and compare lengths, while still parsing existing stored files.

I have this implemented behind a `codecompanion.utils.markdown` helper with a regression test (8 cases; 6 fail on `main`, 2 controls pass) and will open a PR if you are happy with the approach.

**A related question**: `lua/codecompanion/config.lua` tells the model that code blocks are always four backticks — which, as far as I can tell from #913, was adopted as a workaround for this very bug. Once fences are computed, would you accept relaxing that to *"open with four backticks, and use a longer fence than any run inside the code"*? I would send it as a separate docs/prompt PR rather than mixing it into the fix.

### minimal.lua

The stock file, unedited — this bug needs no special configuration.

````lua
---@diagnostic disable: missing-fields

--[[
NOTE: Set the config path to enable the copilot adapter to work.
It will search the following paths for a token:
  - "$CODECOMPANION_TOKEN_PATH/github-copilot/hosts.json"
  - "$CODECOMPANION_TOKEN_PATH/github-copilot/apps.json"
--]]
vim.env["CODECOMPANION_TOKEN_PATH"] = vim.fn.expand("~/.config")

vim.env.LAZY_STDPATH = ".repro"
load(vim.fn.system("curl -s https://raw.githubusercontent.com/folke/lazy.nvim/main/bootstrap.lua"))()

-- Your CodeCompanion setup
local plugins = {
  {
    -- Test with local version (delete if not required):
    --   dir = "~/Code/Neovim/codecompanion.nvim/fix-a",
    --   name = "codecompanion",
    dependencies = {
      { "nvim-lua/plenary.nvim" },
      {
        "nvim-treesitter/nvim-treesitter",
        lazy = false,
        build = ":TSUpdate",
      },

      -- Test with blink.cmp
      {
        "saghen/blink.cmp",
        version = "1.*",
        opts = {
          keymap = {
            preset = "enter",
            ["<S-Tab>"] = { "select_prev", "fallback" },
            ["<Tab>"] = { "select_next", "fallback" },
          },
          cmdline = { sources = { "cmdline" } },
          sources = {
            default = { "lsp", "path", "buffer", "codecompanion" },
          },
        },
      },

      -- Or, test with nvim-cmp
      -- { "hrsh7th/nvim-cmp" },
    },
    opts = {
      --Refer to: https://github.com/olimorris/codecompanion.nvim/blob/main/lua/codecompanion/config.lua
      interactions = {
        --NOTE: Change the adapter as required
        chat = { adapter = "copilot" },
        inline = { adapter = "copilot" },
      },
      opts = {
        log_level = "DEBUG",
      },
    },
  },
}

-- Leaving this comment in to see if the issue author notices ;-)
-- This is so I can tell if they've really tested with their own minimal.lua file
-- or if they're just copy-pasting from the docs and pasting it into the issue

require("lazy.minit").repro({ spec = plugins })

-- CONFIGURE PLUGINS HERE -----------------------------------------------------

-- Setup Tree-sitter
-- NOTE: Please restart Neovim to ensure parsers are loaded correctly
require("nvim-treesitter")
  .install({
    "lua",
    "markdown",
    "markdown_inline",
    "yaml",
  }, { summary = true, max_jobs = 10 })
  :wait(1800000)
````

### Steps to reproduce

`example.md` and `repro.lua` are both below, in the same directory as `minimal.lua`.

1. `nvim -u minimal.lua`
2. Restart once so the parsers load, as the file's own note says
3. `:CodeCompanionChat`
4. `:luafile repro.lua` — this runs the real `mcp/tool_bridge` success handler over the text of `example.md`, which is exactly what an MCP server returning a file's contents produces. No API key, no MCP server and no network round trip are involved
5. Look at the chat buffer: the tool result's code block ends at `example.md`'s *first* inner fence, the README's prose renders as chat prose, and the final fence opens a block that never closes

To see it as the renderer does rather than by eye:

````vim
:lua local q = vim.treesitter.query.parse('markdown', '(fenced_code_block) @b')
:lua local p = vim.treesitter.get_parser(0, 'markdown')
:lua for _, n in q:iter_captures(p:parse()[1]:root(), 0) do local s, _, e = n:range() print(s + 1, e) end
````

On `6eec6b42` that prints three blocks — `6 13`, `17 19`, `22 22` — where one is expected.

### Example files

`example.md`, the file the tool returns. Nothing unusual about it; it is an ordinary README that uses four-backtick fences, exactly as CodeCompanion's own system prompt asks the model to:

`````markdown
# Example project README

Install it:

````sh
brew install example
````

Then configure it:

````lua
require("example").setup({ enabled = true })
````

That's all.
`````

`repro.lua`, which stands in for the MCP server:

````lua
-- Simulates what an MCP server returns: the text of a file that happens to
-- contain four-backtick code fences.
--
-- Nothing here is specific to MCP. The same fixed-length wrapper is used by
-- several built-in tools; this one is just the easiest to drive without an
-- MCP server, an API key, or a network round trip.

local bridge = require("codecompanion.mcp.tool_bridge")

local _, tool_cfg = bridge.build(
  { name = "srv", cfg = {} },
  { name = "echo", description = "echoes", inputSchema = { type = "object" } }
)
local tool = tool_cfg.callback()

-- The orchestrator normally attaches this when the LLM calls the tool.
tool.function_call = {
  id = "call_1",
  type = "function",
  ["function"] = { name = "srv_echo", arguments = "{}" },
}

tool.output.success(tool, { table.concat(vim.fn.readfile("example.md"), "\n") }, {
  tools = { chat = require("codecompanion").last_chat() },
})
````

And the resulting chat buffer, verbatim, as written by the run above:

`````text
## Me

## CodeCompanion (Copilot)

MCP: srv_echo executed successfully:
````
# Example project README

Install it:

````sh
brew install example
````

Then configure it:

````lua
require("example").setup({ enabled = true })
````

That's all.
````
`````

### Log output (optional)

Nothing is logged, which is part of the problem — the corruption is silent. Happy to attach a trace-level log if it would help.

### Screenshots or recordings (optional)

N/A — the verbatim buffer contents above are more precise than a screenshot.

ImmanuelHaffner:
## Description

Every site that embeds arbitrary text in the chat buffer wrapped it in a **hard-coded** fence — four backticks almost everywhere, three in one place and five in another.
When the embedded text itself contains a line whose first non-space characters are four or more backticks, that line closes the wrapper early.
The payload is truncated as far as any Markdown parser is concerned, the wrapper's own closing fence becomes an *opening* fence, and fence parity is inverted for **the rest of the buffer**.

This is reachable from ordinary use: `@{read_file}` on a Markdown file that documents fenced code, `/help` on a doc page containing examples, a `@{cmd_runner}` result that echoes a fence, or any MCP tool returning Markdown.

````````text
BEFORE — reading a file that contains a four-backtick fence

````markdown        <- wrapper opens (hard-coded 4)
# Example
````                <- a line of the payload; closes the wrapper early
More content
````                <- the wrapper's closing fence now OPENS a block, and parity
                       is inverted for everything after it

AFTER — the wrapper is sized to what it wraps

`````markdown       <- 5, because the payload's longest run is 4
# Example
````                <- inert: too short to close a longer fence
More content
`````               <- closes correctly; parity preserved
````````

The two symptoms users report — "the code block is missing its terminating fence" and "two code blocks rendered interleaved" — are this one defect at different parities: with a single offending tool result the orphaned fence runs to the end of the buffer; with two results in a turn it pairs with the next result's opening fence.

**The fix.** A new `lua/codecompanion/utils/markdown.lua` with two functions:

- `fence(content, min)` returns a fence longer than the longest backtick run in `content`, never shorter than `min`.
- `code_block(content, { info = …, min = … })` wraps `content` in a block it cannot terminate, guaranteeing the closing fence starts its own line and using only the first line of `info` (a newline there would end the opening fence's line early).

This is the CommonMark and GitHub nesting rule, and the one Tree-sitter implements — which is what matters, because the chat buffer is parsed as Markdown.

**The payload is never rewritten.**
There is no escape mechanism inside a fenced code block, so "fixing" the payload's backticks would mean altering bytes the user reads and copies, and it is not even well defined for a payload whose own fences are already unbalanced (a truncated read, a diff hunk that begins inside a code block).
A longer outer fence is robust to all of those cases.

**Why 22 call sites rather than one chokepoint.**
Sizing has to happen where the fence is written.
`Chat:add_tool_output` receives an already-formatted string and writes it verbatim, and by then the correct length is unrecoverable — precisely because the pathological case is the one where payload lines are indistinguishable from wrapper lines.

**Each site keeps its existing fence length as a floor**, so output that does not collide is byte-identical to today: `cmd_tool`'s error path stays at three backticks, `approval_prompt` stays at five, everything else stays at four.
The diff is mechanical and the behaviour change is confined to payloads that were already corrupting the buffer.

**Deliberately not included.** `interactions/chat/tools/builtin/code_review/store.lua` writes a four-backtick block and later *reads its own output back* with `^````(%S*)%s*$`.
Making the writer dynamic needs a length-aware reader plus a legacy-read test, so it is a separate, self-contained PR rather than a rider on this one.

## Supporting Documentation

- CommonMark, [fenced code blocks](https://spec.commonmark.org/0.31.2/#fenced-code-blocks): "The content of the code block consists of all subsequent lines, until a closing code fence of the same type as the code block began with (backticks or tildes), and with **at least as many backticks or tildes** as the opening code fence."
- GitHub, [creating and highlighting code blocks](https://docs.github.com/en/get-started/writing-on-github/working-with-advanced-formatting/creating-and-highlighting-code-blocks#fenced-code-blocks): to include a fence inside a block, wrap it in a longer one.
- The chat buffer is parsed by [tree-sitter-markdown](https://github.com/tree-sitter-grammars/tree-sitter-markdown), which implements the same rule, so an early close is visible to the plugin's own queries and not merely a rendering artefact.

## AI Usage

This PR was written with CodeCompanion itself, using **Claude Opus 5 (xhigh)**.
I found and diagnosed the defect while debugging corrupted chat buffers in my own session, directed the investigation, chose the approach (a fresh outer fence rather than patching the payload, and per-site length floors to keep non-colliding output byte-identical), and reviewed every hunk.
The assistant produced most of the mechanical call-site conversions and drafted the two test files, which I then reviewed and adjusted.

## Related Issue(s)

- Fixes #3353

The consumer-side half of what I found is a **separate, pre-existing** bug and gets its own PR: an unbalanced fence above the last user header makes the Tree-sitter query miss that section, so `parser.messages()` returns `nil` and the user's typed prompt is silently dropped on submit.
That is #913, and a fix follows shortly.

Keeping them apart is deliberate: they sit at different layers (producing Markdown vs. consuming it), each is independently revertible, and this PR alone fixes the reported rendering symptom.
The parser fix additionally protects against causes that involve no tool at all — for instance a response truncated mid-code-block by `max_tokens`.

## Screenshots

N/A

## Checklist


olimorris:
Once again, thanks for all your hard work on this. This PR alongside #3355 add some much needed robustness to the chat buffer 👏🏼

PR #3355, opened 2026-09-02
fix(chat): recover a prompt hidden from Tree-sitter by an unbalanced fence

### LINKED ISSUE #913: [Bug]: Incorrect Markdown breaks the chat window

ImmanuelHaffner:
## Description

`queries/markdown/chat.scm` locates the user's message by walking Markdown sections.
An unbalanced code fence anywhere above the last `## <user>` header makes the Markdown parser swallow that header, so the query matches no user section and `parser.messages()` returns `nil`.
Because `helpers.has_user_messages()` is true mid-conversation, `Chat:submit()` does **not** warn "No messages to submit" — it sends the conversation *without the text the user just typed*, silently.

No tool is needed to reach this.
Any response truncated mid-code-block by `max_tokens` leaves an unbalanced fence, and from then on every prompt typed into that chat is discarded on submit:

``````text
## CodeCompanion

Here you go:

````lua
local x = 1          <- the response stopped here; the fence never closed

## Me                <- swallowed by the open block, so the query never sees it

please fix the bug   <- typed, then silently dropped on submit
``````

`parser.headers()` fails on the same buffer, and worse: it does not return `nil`, it returns the last header it can still *see*, which is an **earlier** one.
That value seeds `chat.header_line` for restored chats, so every later parse starts from the wrong row, `parser.messages()` finds a *previous* message and returns it, and the user's new prompt is replaced by an old one rather than merely dropped.
Measured on a buffer whose real last header is row 9: `parser.headers()` returns row 0.

One correction to the record, since it shaped how this was approached.
The ranged parse does not protect against this: `parser.messages` calls `parse({ header_line - 1, -1 })`, but the top-level Markdown `LanguageTree` still parses the whole buffer, so a fence broken *above* the range still swallows the header.
Measured both ways — `parse({ 0, -1 })` and `parse({ header_line - 1, -1 })` extract the user message as `""` on the same buffer.

**The fix is strictly additive. Tree-sitter remains the only primary path, and neither `M.messages` nor `M.headers` loses its existing query.**

- `M.messages` gains a fallback that runs *only* when the query yields nothing: it reads raw lines below `chat.header_line`, reuses `helpers.strip_context()`, and returns the trimmed join. The healthy path is reached by an earlier `return` and executes unchanged code, so it cannot be affected. A genuinely empty section still returns `nil`, which tool auto-submits depend on.
- `M.headers` gains a raw scan that considers candidates **only below** the row Tree-sitter reported — so a header the query did find is never traded for an earlier one — and accepts a candidate only when it is provably inside a `fenced_code_block` carrying fewer than two `fenced_code_block_delimiter` children, i.e. a fence that was never closed. A `## <user>` heading inside a *closed* block (someone pasting a transcript) therefore goes on being treated as code.

`chat.header_line` is what makes the fallback trustworthy rather than a guess: it comes from the builder's own bookkeeping, so the recovery never re-derives a position from a tree already known to be wrong, and it does not re-parse.

The recovery is reported through `log:warn` **once per buffer**, with the per-parse detail at `log:debug`.
Once per parse would be noise: that handler is registered at `vim.log.levels.WARN` and so reaches `vim.notify`, and since a malformed transcript is never rewritten, the fallback fires on every submit for the rest of the conversation.

Deliberately **not** included: any repair of the buffer. A corrupted transcript stays as it is — it is a record of what was actually said — and no closing delimiter is injected. This change only stops it costing the user their prompt.

Nine cases in `tests/interactions/chat/test_parser_resilience.lua`, of which four fail on current `main`. They cover the recovery, multi-line prompts, context stripping, both empty-section guards (balanced and broken), that a healthy buffer is byte-for-byte unaffected, and that the warning fires exactly once and not at all on a healthy buffer.

## Supporting Documentation

- [tree-sitter-markdown](https://github.com/tree-sitter-grammars/tree-sitter-markdown) is what the chat buffer is parsed with; an unterminated `fenced_code_block` extends to the end of the document and carries a single `fenced_code_block_delimiter` child, which is the signal used to tell "never closed" from "closed on purpose".
- CommonMark, [fenced code blocks](https://spec.commonmark.org/0.31.2/#fenced-code-blocks): a block runs "until a closing code fence … or the end of the containing block", which is why one missing fence changes the meaning of everything below it.

## AI Usage

This PR was written with CodeCompanion itself, using **Claude Opus 5 (xhigh)**.
I hit the bug in my own session, diagnosed it, and chose the approach — additive fallbacks only, Tree-sitter left as the primary path, and the "provably unclosed fence" guard in `M.headers` rather than a plain raw scan, which would have preferred a heading inside a closed code block over the real header.
The assistant produced the implementation and the tests from that, and I reviewed every hunk.

## Related Issue(s)

- Fixes #913

I realise #913 was closed as `NOT_PLANNED` and that #1385 was declined for replacing the Tree-sitter extraction, so this is deliberately a different shape: nothing is removed or replaced, the query stays the sole primary path in both functions, and the new code is unreachable unless the query has already come up empty (or, in `M.headers`, unless an unclosed fence is demonstrably hiding a later header). Nothing here constrains the ranged-parse work in #552 — the fallback consumes the tree that `M.headers` already parsed and adds no parse of its own. If you would still rather not carry it, I would be glad to know which part is the objection, because four people have now reported the symptom.

This is **independent of #3354** and the two can merge in either order — verified, not assumed: this branch is based on `main` (`6eec6b42`), does not contain #3354's commits, shares no files with it, and the two merge cleanly with each other. #3354 removes the most common *cause* of an unbalanced fence (fixed-length wrappers around tool output); this PR makes the *consequence* harmless whatever the cause, including the ones with no tool involved.

## Screenshots

N/A

## Checklist

### DIFF
--- a/lua/codecompanion/interactions/chat/parser.lua
+++ b/lua/codecompanion/interactions/chat/parser.lua
@@ -89,45 +89,25 @@ function M.get_settings_key(chat, opts)
   return key_name, node
 end
 
----Chat buffers already warned about, so a malformed transcript is reported once
----rather than on every submit. The offending fence stays broken for the rest of
----the conversation -- the transcript is a record and is never rewritten -- so the
----fallback below keeps firing for as long as the chat lives. Buffer numbers are
----not reused within a session, so this never suppresses a different buffer.
+---Chat buffers already warned about, so a recovery is reported once per buffer
 local warned_buffers = {}
 
----Recover the user's message by reading the buffer, when Tree-sitter cannot
----
----`queries/markdown/chat.scm` finds the message by walking Markdown sections, so
----an unbalanced code fence anywhere above the user header makes the parser
----swallow that header. The query then matches no user section and, because
----`helpers.has_user_messages()` is true mid-conversation, the text the user just
----typed is dropped silently on submit. Any producer of chat text can leave a
----fence open, including an LLM response truncated mid-block by `max_tokens`.
----
----This runs only when the query yielded nothing, and it does not re-parse.
----`start_range` is `chat.header_line`, the 1-based row of the user header, so the
----message is everything below that row: the position comes from the builder's own
----bookkeeping rather than from a tree already known to be wrong, which is what
----makes reading raw lines trustworthy here.
+---Recover the user's message from raw buffer lines when the query finds none
 ---@param chat CodeCompanion.Chat
----@param start_range number The 1-based row of the user header
+---@param start_range number
 ---@return { content: string }|nil
 local function recover_messages(chat, start_range)
-  -- `start_range - 1` is the header row 0-indexed, so `start_range` is the row below it
+  -- start_range is 1-based, so it already points at the row below the header
   local lines = vim.api.nvim_buf_get_lines(chat.bufnr, start_range, -1, false)
 
-  -- The row below a header is blank, and `strip_context` only strips *leading*
-  -- entries, so the context block has to be flushed to the front to be seen. The
-  -- Tree-sitter path never meets this: it is handed section text, not raw lines.
+  -- strip_context only strips leading entries, so the blank row under the header must go first
   while lines[1] and vim.trim(lines[1]) == "" do
     table.remove(lines, 1)
   end
   lines = helpers.strip_context(lines)
   local content = vim.trim(table.concat(lines, "\n"))
 
-  -- An empty section must stay empty: tools auto-submit with no user message, and
-  -- inventing one here would send a phantom prompt.
+  -- Tool auto-submits send no user message, so an empty section must stay empty
   if content == "" then
     return nil
   end
@@ -174,15 +154,8 @@ function M.messages(chat, start_range)
 end
 
 ---Is `row` inside a fenced code block that was never closed?
----
----A closed block has two `fenced_code_block_delimiter` children; an unterminated
----one has a single opening delimiter and runs to the end of the buffer. Only such
----a block can hide a header from `queries/markdown/chat.scm` unintentionally, so
----this is the one situation in which the scan below may overrule Tree-sitter: a
----heading inside a *closed* block is there because its author put it there, and
----must go on being treated as code.
----@param root TSNode The root of the tree `M.headers` has already parsed
----@param row number 0-indexed
+---@param root TSNode
+---@param row number
 ---@return boolean
 local function hidden_by_open_fence(root, row)
   local node = root:descendant_for_range(row, 0, row, 0)
@@ -194,7 +167,7 @@ local function hidden_by_open_fence(root, row)
           delimiters = delimiters + 1
         end
       end
-      return delimiters < 2
+      return delimiters < 2 -- a closed block has two delimiters; an unterminated one only an opener
     end
     node = node:parent()
   end
@@ -202,21 +175,10 @@ local function hidden_by_open_fence(root, row)
 end
 
 ---Find a user header that an unclosed code fence hid from Tree-sitter
----
----`M.headers` seeds `chat.header_line` for restored chats, and on a malformed
----buffer it does not fail cleanly: it returns the last header it can still *see*,
----which is an earlier one. Everything downstream then parses from that row, so
----`M.messages` finds a previous message and returns it -- worse than returning
----nothing, because the fallback above sees content and never fires, and the user's
----new prompt is replaced by an old one.
----
----Only rows below `after` are considered, so a header Tree-sitter did find is
----never traded for an earlier one, and every candidate must be provably hidden by
----an unclosed fence.
 ---@param chat CodeCompanion.Chat
----@param root TSNode The root of the tree `M.headers` has already parsed
----@param after number The 0-indexed row Tree-sitter reported, or -1 if it found none
----@return number|nil The 0-indexed row of the header, in the rows Tree-sitter reports
+---@param root TSNode
+---@param after number 0-indexed row, or -1 when Tree-sitter found none
+---@return number|nil
 local function recover_headers(chat, root, after)
   local lines = vim.api.nvim_buf_get_lines(chat.bufnr, after + 1, -1, false)
   for i = #lines, 1, -1 do
@@ -249,8 +211,6 @@ function M.headers(chat)
     end
   end
 
-  -- Tree-sitter stays authoritative unless an unclosed fence hid a later header
-  -- from it. When nothing is recovered its own result is returned untouched.
   local recovered = recover_headers(chat, root, last_match and last_match:range() or -1)
   if recovered then
     return recovered
--- a/lua/codecompanion/interactions/chat/parser.lua
+++ b/lua/codecompanion/interactions/chat/parser.lua
@@ -89,6 +89,61 @@ function M.get_settings_key(chat, opts)
   return key_name, node
 end
 
+---Chat buffers already warned about, so a malformed transcript is reported once
+---rather than on every submit. The offending fence stays broken for the rest of
+---the conversation -- the transcript is a record and is never rewritten -- so the
+---fallback below keeps firing for as long as the chat lives. Buffer numbers are
+---not reused within a session, so this never suppresses a different buffer.
+local warned_buffers = {}
+
+---Recover the user's message by reading the buffer, when Tree-sitter cannot
+---
+---`queries/markdown/chat.scm` finds the message by walking Markdown sections, so
+---an unbalanced code fence anywhere above the user header makes the parser
+---swallow that header. The query then matches no user section and, because
+---`helpers.has_user_messages()` is true mid-conversation, the text the user just
+---typed is dropped silently on submit. Any producer of chat text can leave a
+---fence open, including an LLM response truncated mid-block by `max_tokens`.
+---
+---This runs only when the query yielded nothing, and it does not re-parse.
+---`start_range` is `chat.header_line`, the 1-based row of the user header, so the
+---message is everything below that row: the position comes from the builder's own
+---bookkeeping rather than from a tree already known to be wrong, which is what
+---makes reading raw lines trustworthy here.
+---@param chat CodeCompanion.Chat
+---@param start_range number The 1-based row of the user header
+---@return { content: string }|nil
+local function recover_messages(chat, start_range)
+  -- `start_range - 1` is the header row 0-indexed, so `start_range` is the row below it
+  local lines = vim.api.nvim_buf_get_lines(chat.bufnr, start_range, -1, false)
+
+  -- The row below a header is blank, and `strip_context` only strips *leading*
+  -- entries, so the context block has to be flushed to the front to be seen. The
+  -- Tree-sitter path never meets this: it is handed section text, not raw lines.
+  while lines[1] and vim.trim(lines[1]) == "" do
+    table.remove(lines, 1)
+  end
+  lines = helpers.strip_context(lines)
+  local content = vim.trim(table.concat(lines, "\n"))
+
+  -- An empty section must stay empty: tools auto-submit with no user message, and
+  -- inventing one here would send a phantom prompt.
+  if content == "" then
+    return nil
+  end
+
+  if not warned_buffers[chat.bufnr] then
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

**Comment by olimorris** [comment]
Firstly, thank you very much for the PR.

Unfortunately, it's just too comment-heavy. I'd ask you to go back, look at it, and revise it. The code should be clear enough that it's self-explanatory. I'd also ask you to look at the test name as well. It just doesn't fit in with CodeCompanion standards.

For the LuaLS annotations, we're explaining things that should be self-explanatory. I'd also advise collapsing function parameters into a Lua table, it makes them more scalable in the future.

To make this easy, just advise your agent to read the agents.md file. That should be clear enough, and to also look at the existing work in the plugin.

**Comment by ImmanuelHaffner** [comment]
All right, will do.

> I'd also ask you to look at the test name as well. It just doesn't fit in with CodeCompanion standards.

Can you elaborate this? Do you expect the test naming convention to compose a proper sentence, like `T["Context"]["Can be added to the UI of the chat buffer"]`? This is not a strict rule, e.g. `T["Context"]["Prompt-library slash commands add a single tool"]`. Or is it about `["Parser resilience"]` not naming a component in code?
What is the standard?

#### New

`tests/interactions/chat/test_parser_resilience.lua`
```lua
103:T["Parser resilience"]["a balanced response leaves the prompt extractable"] = function()
108:T["Parser resilience"]["an unterminated fence does not eat the next prompt"] = function()
113:T["Parser resilience"]["an unterminated fence does not hide the last user header"] = function()
119:T["Parser resilience"]["context lines are stripped from the recovered prompt"] = function()
124:T["Parser resilience"]["a multi-line prompt is recovered whole"] = function()
129:T["Parser resilience"]["an empty user section still yields no message"] = function()
135:T["Parser resilience"]["an empty user section under a broken fence yields no message"] = function()
140:T["Parser resilience"]["recovery is reported once, not on every parse"] = function()
151:T["Parser resilience"]["a healthy buffer is parsed without warning"] = function()
```

**Existing**

`tests/interactions/chat/test_context.lua`
```lua
26:T["Context"]["Can be added to the UI of the chat buffer"] = function()
48:T["Context"]["Cannot be added twice with the same id"] = function()
65:T["Context"]["Prompt-library slash commands add a single tool"] = function()
83:T["Context"]["Prompt-library slash commands add a tool group"] = function()
102:T["Context"]["Can be deleted"] = function()
176:T["Context"]["Tools and their schema can be deleted"] = function()
211:T["Context"]["Can share all of a buffer"] = function()
301:T["Context"]["Tree-sitter test"] = function()
316:T["Context"]["Render"] = function()
341:T["Context"]["Render invisible"] = function()
360:T["Context"]["can be cleared from messages"] = function()
383:T["Context"]["file context_items use absolute paths"] = function()
407:T["Context"]["Correctly removes tool schema and usage flag on context deletion"] = function()
466:T["Context"]["Show icons immediately when added with default parameters"] = function()
521:T["Context"]["Extensions listed in the sync_diff config are watched when added"] = function()
551:T["Context"]["Files and buffers are formatted by the formatter for their extension"] = function()
581:T["Context"]["Tool group with collapse_tools shows single group context"] = function()
607:T["Context"]["Tool group without collapse_tools shows individual tools"] = function()
633:T["Context"]["Removing collapsed group removes all its tools and system message"] = function()
679:T["Context"]["does not fold single context item but applies extmark"] = function()
```

olimorris:
I meant that you've called it `test_parser_resilience.lua` rather than `test_parser`. I'd also argue that `TERMINATED` and `BALANCED`  in the context of testing the parser aren't as clear, when reading cold, as something like `INVALID` and `VALID`.

It's this general readability that I want to make sure this PR adheres to.

ImmanuelHaffner:
@olimorris all done

